PR #7, opened 2020-02-04
Update 92_downloads.rst

kapokasa:


MarkRiddoch:
The downloads page has had a revamp and this has been fixed as part of that.

PR #11, opened 2020-02-14
Fogl 3747: c++ PiServer North plugin does not properly manage the change of an OMF type

sstefano69:
initial stage


sstefano69:
bug not present.

PR #18, opened 2020-02-28
FOGL-3718 patch Remove plugin files with package vs manual install check

praveen-garg:


pintomax:
@praveen-garg 
tested removal of a manually installed plugin c++ Random. OK
tested return code of plugin removal by using original_name = lowercase name and got error != 0


LGTM!

PR #22, opened 2020-03-05
FOGL- 3718 FOGL-3775 Clear available packages cache on package removal via API

rajeshkumar426:
Signed-off-by: rajeshkumar426 <rajeshkumar426@outlook.com>

when plugin packages gets removed successfully via API, then clear cache to get latest available plugin list.

**note..** We have no control if purge / removal action taken directly on the system. then cache will refresh after configured time only i.e. be default 15 mins

### DIFF
--- a/python/fledge/services/core/api/plugins/remove.py
+++ b/python/fledge/services/core/api/plugins/remove.py
@@ -19,7 +19,6 @@ from fledge.common.storage_client.payload_builder import PayloadBuilder
 from fledge.common.configuration_manager import ConfigurationManager
 from fledge.common.common import _FLEDGE_ROOT
 from fledge.common.audit_logger import AuditLogger
-from fledge.services.core.server import Server
 
 __author__ = "Rajesh Kumar"
 __copyright__ = "Copyright (c) 2020, Dianomic Systems Inc."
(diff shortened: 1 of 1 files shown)


python/fledge/services/core/api/plugins/remove.py
```diff
@@ -208,6 +208,8 @@ def purge_plugin(plugin_type: str, name: str) -> tuple:
         code = os.system(cmd)
         if code:
             raise PackageError(link)
+        else:
+            common._get_available_packages.cache_clear()
```
ashish-jabble: Also update the _package_cache_manager list attributes (like last_accessed_time)

python/fledge/services/core/api/plugins/remove.py
```diff
@@ -19,6 +19,7 @@
 from fledge.common.configuration_manager import ConfigurationManager
 from fledge.common.common import _FLEDGE_ROOT
 from fledge.common.audit_logger import AuditLogger
+from fledge.services.core.server import Server
```
ashish-jabble: Move this import in function scope, As we have seen in past import Server statement giving us
circular import issue & to disallows many unit tests from running.

PR #30, opened 2020-03-16
FOGL-3427:  Implement the OCS connectivity in the PI_Server plugin allowing the user to select the proper Endpoint

sstefano69:

Work in progress

Tested on:
* Ubuntu 18
* Buster
* Centos - the OCS integration doesn't work

OCS:
image

image

image

PiWEB API:
image

CR:
image

### DIFF
--- a/C/plugins/common/include/omf.h
+++ b/C/plugins/common/include/omf.h
@@ -24,10 +24,12 @@
 #define OMF_TYPE_FLOAT		"number"
 #define OMF_TYPE_UNSUPPORTED	"unsupported"
 
-#define END_POINT_PIWEB_API  "p"
-#define END_POINT_CR         "c"     // Connector RELAY
-#define END_POINT_OCS        "o"
-#define END_POINT_EDS        "e"
+enum OMF_END_POINT {
+	END_POINT_PIWEB_API,
+	END_POINT_CR,
+	END_POINT_OCS,
+	END_POINT_EDS
+};
 
 using namespace std;
 using namespace rapidjson;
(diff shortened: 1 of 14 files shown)


C/plugins/common/include/omf.h
```diff
@@ -24,6 +24,11 @@
 #define OMF_TYPE_FLOAT		"number"
 #define OMF_TYPE_UNSUPPORTED	"unsupported"
 
+#define END_POINT_PIWEB_API  "p"
```
MarkRiddoch: It might be better to make this an enumerated type, that way you can avoid the expensive string comparisons.

C/plugins/common/omf.cpp
```diff
@@ -1427,12 +1426,16 @@ const std::string OMF::createTypeData(const Reading& reading)
 		tData.append("\": {\"type\": \"string\"},");
 	}
 
-	if (m_PIServerEndpoint.compare("c") == 0)
+	// Connector relay / ODS / EDS
+	if (m_PIServerEndpoint.compare(END_POINT_CR) == 0 ||
```
MarkRiddoch: If you use an enum then you can avoid the if else clauses and have a simple switch statement.

C/plugins/common/include/omf.h
sstefano69: done, thanks.

C/plugins/common/omf.cpp
sstefano69: done, thanks.

PR #39, opened 2020-04-01
FOGL-3743 Updated documentation images

MarkRiddoch:


ashish-jabble:
@MarkRiddoch Please fix DCO sign-off and have a look on below points may be some of them to get resolve.

dashboard.JPG - As our demo screenshots with sense hat south service but I don't see any info related here instead I see with sinusoid.
fledge_architecture.png - Splunk? we don't have the repo in fledge side
PI_connect.jpg - fyi, foglamp connector used in screenshot. Do we need to replace it?
pi_plugin_config.JPG - As we removed URL config item in FOGL-3869, I think we should update this as well may be after merge?
PI_token.jpg - fyi, foglamp connector used in screenshot. Do we need to replace it?
south_service_config.JPG - fyi, Display name is not matched with latest code base.
south_services.JPG - Still has the FogLAMP header which needs to be fixed for sure

MarkRiddoch:
> @MarkRiddoch Please fix DCO sign-off and have a look on below points may be some of them to get resolve.
> 
> dashboard.JPG - As our demo screenshots with sense hat south service but I don't see any info related here instead I see with sinusoid.

Updated

> fledge_architecture.png - Splunk? we don't have the repo in fledge side

Let's not worry about that. This is the picture we use on the fledge presentation deck

> PI_connect.jpg - fyi, foglamp connector used in screenshot. Do we need to replace it?
> pi_plugin_config.JPG - As we removed URL config item in FOGL-3869, I think we should update this as well may be after merge?
> PI_token.jpg - fyi, foglamp connector used in screenshot. Do we need to replace it?

Yes, all of these three need to be redone, in fact I aim to rewrite that entire section in the next few days.

> south_service_config.JPG - fyi, Display name is not matched with latest code base.
Not a big deal, I woudl say we leave that

> south_services.JPG - Still has the FogLAMP header which needs to be fixed for sure
Fixed - thanks.

PR #41, opened 2020-04-02
FOGL-3912 Add generation of plugin list for Fledge documentation

MarkRiddoch:


docs/Makefile
```diff
@@ -9,12 +9,16 @@ SOURCEDIR     = .
 BUILDDIR      = _build
 
 # Put it first so that "make" without argument is like "make help".
-help:
+help:	fledge_plugins.rst
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
 .PHONY: help Makefile
```
ashish-jabble, line 15: ADD `fledge_plugins.rst`

docs/Makefile
```diff
@@ -9,12 +9,16 @@ SOURCEDIR     = .
...
 
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
-%: Makefile
-	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
\ No newline at end of file
+%: Makefile fledge_plugins.rst
+	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+fledge_plugins.rst:
```
ashish-jabble, line 22: To avoid conflict Better to rename the target without extension, as we are creating same output file with same name

docs/Makefile
```diff
@@ -9,12 +9,16 @@ SOURCEDIR     = .
...
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
-%: Makefile
-	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
\ No newline at end of file
+%: Makefile fledge_plugins.rst
+	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+fledge_plugins.rst: 
+	@echo Building page with table of plugins
+	@bash scripts/fledge_plugin_list fledge_plugins.rst
```
ashish-jabble, line 24: Better to add `clean` target - remove fledge_plugins.rst file. As of now it removes only _build content i.e from BUILDDIR
Sphinx inbuilt variable

docs/Makefile
MarkRiddoch: Sorry, I don't understand your comment here?

ashish-jabble: As `rst` output is being saved & named as "fledge_plugins.rst".
So, If the file exists  `make` actually got confused. As by default Makefile targets are the file targets.
Hence either to change the target/output file name OR add this target in `.PHONY`

docs/scripts/fledge_plugin_list
```diff
@@ -0,0 +1,120 @@
...
+		if [ $product = "foglamp" ]; then
+			org=dianomic
+		else
+			org=fledge-iot
+		fi
+		type=`echo $repo | sed -e 's/fledge-//' -e 's/foglamp-//' -e 's/-.*//'`
+		name=`echo $repo | sed -e 's/fledge-//' -e 's/foglamp-//' -e "s/${type}-//"`
+		echo $repo, $name, $type >>/tmp/repos
+		if [ "$type" = "$tableType" ]; then
+			description=`curl -s -H "$header" https://api.github.com/repos/$org/$repo | jq '.description' | sed -e 's/"//g'`
```
ashish-jabble, line 25: In case of White-labelled description always returns null, because it has a check of organization and that repo does not exist on dianomic org.

docs/scripts/fledge_plugin_list
```diff
@@ -0,0 +1,120 @@
...
+if [ ! -f ~/.gittoken ]; then
+	echo "You must have a git access token in the file ~/.gittoken"
+	exit 1
+fi
+
+output=$1
+
+token=`cat ~/.gittoken`
+header="Authorization: token $token" 
+fledgeRepos=`curl -s -H "$header" https://api.github.com/orgs/fledge-iot/repos\?per_page=1000 | jq '.[].name' | sed -e 's/"//g' | sort`
```
ashish-jabble, line 42: 1) We can refactor the `jq` as we want only plugins to be fetched out.

```
jq '.[] | select(.name | contains("fledge-south") or contains("fledge-north-") or contains("fledge-filter-") or contains("fledge-rule-") or contains("fledge-notify-")) | .name' | tr -d '""'
```
2) For description info to be fetched out, explicit curl request is NOT needed if add `.description` into above jq but yes it needs some modification in `$REPOSITORIES list & function table` as well as below has description as well.

https://github.com/fledge-iot/fledge/blob/65b8e428f48bdd10990fa3b5adeffd0dfa92f849/docs/scripts/fledge_plugin_list#L25

docs/scripts/fledge_plugin_list
```diff
@@ -0,0 +1,121 @@
...
+foglampRepos=`curl -s -H "$header" https://api.github.com/orgs/dianomic/repos\?per_page=1000 | jq '.[].name' | sed -e 's/"//g' | sort`
```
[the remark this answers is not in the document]
MarkRiddoch: There is a separate script for the foglamp case, it makes up part of the FogLAMP documentation overlay that will be applied. Other things are images and extra sections etc.

docs/Makefile
MarkRiddoch: Make is not at all confused by this. However it does no harm to add to .PHONY, although of course this is a file that is being created so is a very normal way to use make.

docs/scripts/fledge_plugin_list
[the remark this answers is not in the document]
ashish-jabble: Okay, I aware only with white-labelled via dianomic/FogLAMP-release code base which I did the replace the images part itself in script.

docs/scripts/fledge_plugin_list
```diff
@@ -0,0 +1,121 @@
...
+			echo "      - $description" >> $output
+		fi
+	done
+	echo "" >> $output
+}
+
+if [ ! -f ~/.gittoken ]; then
+	echo "You must have a git access token in the file ~/.gittoken"
+	exit 1
+fi
```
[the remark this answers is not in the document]
ashish-jabble: FOGL-3942 created for this so we don't forget it.

docs/scripts/fledge_plugin_list
MarkRiddoch: This is never run over the white labelled code, only on fledge, so there is no null case.

docs/scripts/fledge_plugin_list
MarkRiddoch: We could refactor in lots of ways, but frankly we run this infrequently and it is not worth the effort.

PR #47, opened 2020-04-08
FOGL-3932 Add statistics history and service syslogs to the support

MarkRiddoch:
bundle

### DIFF
--- a/python/fledge/services/core/support.py
+++ b/python/fledge/services/core/support.py
@@ -22,6 +22,7 @@ from fledge.services.core.connect import *
 from fledge.common import logger
 from fledge.services.core.api.service import get_service_records
 from fledge.common.common import _FLEDGE_ROOT, _FLEDGE_DATA
+from fledge.common.configuration_manager import ConfigurationManager
 
 
 __author__ = "Amarendra K Sinha"
@@ -67,10 +68,26 @@ class SupportBuilder:
             try:
                 self.add_syslog_fledge(pyz, file_spec)
                 self.add_syslog_storage(pyz, file_spec)
+                cf_mgr = ConfigurationManager(self._storage)
+                try:
+                    south_cat = await cf_mgr.get_category_child("South")
+                    south_categories = [nc["key"] for nc in south_cat]
+                    for service in south_categories:
+                        self.add_syslog_service(pyz, file_spec, service)
+                except:
+                    pass
+                try:
+                    north_cat = await cf_mgr.get_category_child("North")
+                    north_categories = [nc["key"] for nc in north_cat]
+                    for service in north_categories:
+                        self.add_syslog_service(pyz, file_spec, service)
+                except:
+                    pass
                 await self.add_table_configuration(pyz, file_spec)
                 await self.add_table_audit_log(pyz, file_spec)
                 await self.add_table_schedules(pyz, file_spec)
                 await self.add_table_scheduled_processes(pyz, file_spec)
+                await self.add_table_statistics_history(pyz, file_spec)
                 self.add_service_registry(pyz, file_spec)
                 self.add_machine_resources(pyz, file_spec)
                 self.add_psinfo(pyz, file_spec)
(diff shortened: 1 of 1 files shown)


python/fledge/services/core/support.py
```diff
@@ -67,10 +68,26 @@ def __init__(self, support_dir):
             try:
                 self.add_syslog_fledge(pyz, file_spec)
                 self.add_syslog_storage(pyz, file_spec)
+                cf_mgr = ConfigurationManager(self._storage)
+                try:
+                    south_cat = await cf_mgr.get_category_child("South")
+                    south_categories = [nc["key"] for nc in south_cat]
```
ashish-jabble, line 74: Better to say `sc`

python/fledge/services/core/support.py
```diff
@@ -123,6 +140,15 @@ def add_syslog_storage(self, pyz, file_spec):
             raise RuntimeError("Error in creating {}. Error-{}".format(temp_file, str(ex)))
         pyz.add(temp_file, arcname=basename(temp_file))
 
+    def add_syslog_service(self, pyz, file_spec, service):
+        # The fledge entries from the syslog file
+        temp_file = self._interim_file_path + "/" + "syslog-{}-{}".format(service, file_spec)
+        try:
+            subprocess.call("grep '{}' {} > {}".format(service, _SYSLOG_FILE, temp_file), shell=True)
```
ashish-jabble, line 147: Filtered data is not upto the mark. It's free text search problem. See comments in FOGL-3826, syslog filtration asked in JIRA.

python/fledge/services/core/support.py
```diff
@@ -123,6 +140,15 @@ def add_syslog_storage(self, pyz, file_spec):
             raise RuntimeError("Error in creating {}. Error-{}".format(temp_file, str(ex)))
         pyz.add(temp_file, arcname=basename(temp_file))
 
+    def add_syslog_service(self, pyz, file_spec, service):
+        # The fledge entries from the syslog file
```
ashish-jabble, line 144: services & tasks entries

python/fledge/services/core/support.py
```diff
@@ -123,6 +141,15 @@ def add_syslog_storage(self, pyz, file_spec):
...
+        temp_file = self._interim_file_path + "/" + "syslog-{}-{}".format(service, file_spec)
+        try:
+            subprocess.call("grep '{}\\[' {} > {}".format(service, _SYSLOG_FILE, temp_file), shell=True)
```
ashish-jabble, line 148: This will only work with C Plugins. 
For example: I have 2 north plugins one with C version & another with Python version
```
PI-C[14549]: ERROR: Sending JSON dataType message 'Type' - generic error:
                                         Vs
FogLAMP[23255] INFO: sending_process: sending_process_PI-Python: Started
```

python/fledge/services/core/support.py
```diff
@@ -67,10 +69,26 @@ def __init__(self, support_dir):
...
+                try:
+                    south_cat = await cf_mgr.get_category_child("South")
+                    south_categories = [sc["key"] for sc in south_cat]
+                    for service in south_categories:
+                        self.add_syslog_service(pyz, file_spec, service)
+                except:
+                    pass
+                try:
+                    north_cat = await cf_mgr.get_category_child("North")
+                    north_categories = [nc["key"] for nc in north_cat]
```
ashish-jabble, line 82: `"key": "OMF_TYPES"` also have a children of North, I see related syslog file has been created with empty content. We may skip this key.

python/fledge/services/core/support.py
MarkRiddoch: The Python PI plugin is obsolete. All north plugin should be being loaded with the C task framework. If we find we have problems with specific ones then I suggest we add them as and when they are an issue. In the meantime we cover the 90% of cases that matter in deployments.

ashish-jabble: > This will only work with C Plugins.

Example shows only with north one. And same apply for Python South plugins as well.

MarkRiddoch: We no longer use python south

ashish-jabble: I meant for `south/north python plugins` logs info. 

You mean to collect only South services (which is loaded by C south server) & C-North tasks logs info?

PR #72, opened 2020-04-29
FOGL-4000 common bash logger script used for scripts.services.storage

ashish-jabble:

**Fixed items**
1) indentation & style fixes (square brackets [[ ]] )
2) common write_log bash logger used for logging in case of FLEDGE_ROOT is set

**O/P**
New log
```
Apr 29 17:22:35 aj Fledge [31919] INFO: scripts.services.storage: Fledge storage microservice found in FLEDGE_ROOT location: /home/Development/fledge
```

Previously it was without Fledge Prefix & PID. Below is the old log

```
Apr 28 08:18:55 aj pi: Fledge storage microservice found in FLEDGE_ROOT location: /home/pi/fledge
```

### DIFF
--- a/scripts/services/storage
+++ b/scripts/services/storage
@@ -1,30 +1,33 @@
-#!/bin/sh
-if [ "${FLEDGE_ROOT}" = "" ]; then
-	if [ ! -x /usr/local/fledge/services/fledge.services.storage ] && [ ! -x /usr/local/fledge/services/storage ]; then
-		logger "Unable to find Fledge storage microservice in the default location"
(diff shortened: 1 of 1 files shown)


ashish-jabble:
Because File path included via FLEDGE_ROOT `. $FLEDGE_ROOT/scripts/common/write_log.sh` so that's why kept under else part. Though I am NOT sure about the condition for case when  FLEDGE_ROOT is NOT set seems like a dead code for me.
And I see only log (else one) is happening with both package & non-package installation

PR #73, opened 2020-04-30
FOGL-3980 module name fixes for loading python plugin

ashish-jabble:

**Logs**

```
Apr 30 11:52:26 aj Fledge S I N E[17333] INFO: sinusoid: fledge.plugins.south.sinusoid: Sinusoid plugin_init: called

Apr 30 11:54:02 aj Fledge HTTP South #1[17337] INFO: http_south: fledge.plugins.south.http_south: South HTTP plugin shut down.

Apr 30 11:55:55 aj Fledge Ht #1[18938] ERROR: http_north: fledge.plugins.north.http_north: Server error code: 500, reason: Internal Server Error

Apr 30 12:16:25 aj Fledge Env[13750] INFO: envirophat: fledge.plugins.south.envirophat: Old config for Enviro pHAT plugin {'plugin': {'description': 'Enviro pHAT Poll Plugin', 'type': 'string', 'value': 'envirophat', 'default': 'envirophat'}, 'assetNamePrefix': {'description': ........
```

### DIFF
--- a/python/fledge/services/core/api/plugins/common.py
+++ b/python/fledge/services/core/api/plugins/common.py
@@ -32,7 +32,7 @@ _NO_OF_FILES_TO_RETAIN = 10
 
 def load_python_plugin(plugin_module_path: str, plugin: str, _type: str) -> Dict:
     _plugin = None
-    module_name = "fledge.plugins.{}.{}.{}".format(_type, plugin, plugin)
+    module_name = "fledge.plugins.{}.{}".format(_type, plugin)
     try:
         spec = importlib.util.spec_from_file_location(module_name, "{}/{}.py".format(plugin_module_path, plugin))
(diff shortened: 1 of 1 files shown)


python/fledge/services/core/api/plugins/common.py
```diff
@@ -32,8 +32,9 @@
 
 def load_python_plugin(plugin_module_path: str, plugin: str, _type: str) -> Dict:
     _plugin = None
+    module_name = "fledge.plugins.{}.{}.{}".format(_type, plugin, plugin)
```
praveen-garg, line 35: Please remove .{} plugin. Having `module_name = "fledge.plugins.{}.{}".format(_type, plugin) is enough, so that we can avoid too much of plugin name in logs.

ashish-jabble: sure

PR #100, opened 2020-05-27
Correct where the download link points.

MarkRiddoch:


praveen-garg:
@MarkRiddoch do we also need for docs/REDHAT.rst ?

PR #105, opened 2020-05-29
FOGL-4122 Update zeroconf library and allow south services on remote machines

MarkRiddoch:
The version of zeroconf used previously stops currently announcing services after a short period. This prevents services running outside the machine, e.g. ESP8266 service, to find services other than during a short window after fledge is start.

Also the fledge/south API gives an internal error if a service outside of the machine is registered with the core.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,103 @@
+# Contributing to Fledge
+
+The project welcomes contributions of all types; documentation, code
+changes, new plugins, scripts, just simply reports of the way you use Fledge
+or suggestions of features you would like to see within Fledge.
+
+The following is a set of guidelines for contributing to Fledge IoT
+project and its plugins, which are hosted in
+the [fledge-iot Organization](https://github.com/fledge-iot) on GitHub.
+
+To give us feedback or make suggestions use the [Fledge Slack Channel](https://slack/lfedge.org/).
+
+## Pull requests
+
+**Please ask first** before embarking on any significant work (e.g. implementing new features,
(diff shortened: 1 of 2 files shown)


python/fledge/services/common/service_announcer.py
```diff
@@ -48,14 +48,14 @@ def __init__(self, sname, stype, port, txt):
...
-            socket.inet_aton(host),
             port,
+            addresses=[socket.inet_aton(host)],
             properties=desc,
             server="{}.local.".format(host_name)
         )
         zeroconf = Zeroconf()
         # Refresh zeroconf cache
-        browser = ServiceBrowser(zeroconf, stype, handlers=[self.on_service_state_change])
+        # browser = ServiceBrowser(zeroconf, stype, handlers=[self.on_service_state_change])
```
ashish-jabble, line 58: Zeroconf cache refresh not needed?

MarkRiddoch: It doesn't seem to be. In the testing I did it made no difference, but since I believe it was you that added it I meant to ask you if there was a particular case that I may have missed.

ashish-jabble: It is used to browse for a service of a specific type.
The listener object will have its add_service() and remove_service() methods called when this browser discovers changes in the services availability.
Callback invoked by Zeroconf when new information arrives.Updates information required by browser in the Zeroconf cache.

However, this was even NOT correct before as `info` is NOT being updated when `ServiceStateChange.Added` case. See internal `on_service_state_change` def.

PR #107, opened 2020-06-01
FOGL-3949 Add repository documentation and fix title issue

MarkRiddoch:


docs/quick_start.rst
```diff
@@ -64,10 +65,137 @@ In general, Fledge installation will require the following packages:
...
+    .. code-block:: console
+
+       sudo add-apt-repository ‘deb http://archives.fledge-iot.org/latest/ubuntu1804/aarch64/ / ‘
+
+
+  - Users of the Mendel operating system on a Google Coral create the file /etc/apt/sources.list.d/fledge.list and insert the following content
+
+    .. code-block:: console
+
+       deb http://archives.fledge-iot.org/latest/mendel/aarch64/
```
ashish-jabble, line 116: Add at the end ` /` followed by space

docs/quick_start.rst
```diff
@@ -64,10 +65,137 @@ In general, Fledge installation will require the following packages:
...
+
+On CentOS run the commands
+
+.. code-block:: console
+
+   sudo yum install -y centos-release-scl-rh
+   sudo yum install -y epel-release
+
+
+On RehHat run the command
```
ashish-jabble, line 182: RedHat

docs/quick_start.rst
```diff
@@ -64,10 +65,137 @@ In general, Fledge installation will require the following packages:
...
+
+.. code-block:: console
+
+   sudo yum-config-manager --enable 'Red Hat Enterprise Linux Server 7 RHSCL (RPMs)'
+
+You can now install and upgrade fledge packages using the yum command. For example to install fledge and the fledge GUI you run the command
+
+.. code-block:: console
+
+   sudo yum install fledge fledge-gui
```
ashish-jabble, line 192: Better to add `-y` into it

docs/quick_start.rst
```diff
@@ -64,10 +65,137 @@ In general, Fledge installation will require the following packages:
...
+
+    .. code-block:: console
+
+       sudo vi /etc/apt/sources.list
+       
+    and add the line
+    
+    .. code-block:: console
+
+       deb  http://archives.fledge-iot.org/latest/buster/arm7l/ /
```
ashish-jabble, line 94: Also I see foglamp packages which is available under at `foglamp` directory NOT at the root level. Don't you think it should directly inside the `archives.dianomic.com/`? Otherwise special case needs to be added at white-labelled documentation side.

MarkRiddoch: Yes it should. We want to avoid the special cases.

ashish-jabble: cc/ @YashTatkondawar @praveen-garg

praveen-garg: https://github.com/dianomic/FogLAMP-release/pull/22/files will resolve this too. We should keep the foglamp directory. It was kept separate to allow other product packages too especially **/fogman** binaries

ashish-jabble: if that's the case then change required in `foglamp-documentation` repo

praveen-garg: Added https://github.com/dianomic/foglamp-documentation/pull/5 
I was assuming the pipeline has FogLAMP-release and then pass that to foglamp-documention. Probably we should do that way to avoid scripting duplication and put the processed source code into readthedocs server and archives repo from that final output only.  (instead of baking partially on both)

MarkRiddoch: > https://github.com/dianomic/FogLAMP-release/pull/22/files will resolve this too. We should keep the foglamp directory. It was kept separate to allow other product packages too especially **/fogman** binaries

Currently we are not planning to do this, but you are right we should leave scope for it. @Ashish, let's do that as part of the processing of the FogLAMP variant.

PR #112, opened 2020-06-02
FOGL-4134 Failed external services can not register when they restart

MarkRiddoch:
due to too strict checks in the service registry

### DIFF
--- a/python/fledge/services/core/service_registry/service_registry.py
+++ b/python/fledge/services/core/service_registry/service_registry.py
@@ -162,7 +162,7 @@ class ServiceRegistry:
     def check_address_and_port(cls, address, port):
         # AND based check
         # ugly hack! <Make filter to support AND | OR>
-        services = [s for s in cls._registry if getattr(s, "_address") == address and getattr(s, "_port") == port]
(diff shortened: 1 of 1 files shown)


MarkRiddoch:
> I ran custom CI JOB ub1804 against this branch and system tests are failing. Here is the link `Fledge%20Nightly%20Jobs/job/fledge_nightly_build_ub1804/189/`

That looks very odd as those tests have nothing to do with the change. Will try again with latest develop.

ashish-jabble:
fyi, I have re-triggered JOB against commit `a7c4c5d` and got same results. Also latest develop is broken (See FOGL-4145) you will NOT get any suitable results as fledge is not started.

MarkRiddoch:
> fyi, I have re-triggered JOB against commit `a7c4c5d` and got same results. Also latest develop is broken (See FOGL-4145) you will NOT get any suitable results as fledge is not started.

I've rerun the tests today after the fix for the configuration category changes and since you restarted the PI Server. Now all but one pass, the one that fails is the modes end to end test and it looks like the modbus simulator is not running correctly.

PR #120, opened 2020-06-16
FOGL-4133 POST repository API endpoint added for debian systems

ashish-jabble:

**Assumption**
1)There is no existing setup for currently.
2) Non Package installation (i.e sudo make install) - manual install of FL via sudo will work
3) WL code available [here](http://archives.dianomic.com/foglamp/fixes/FOGL-4133/ubuntu1804/x86_64/sources.tar.gz)

**How to test?**
```
$ curl -sX POST localhost:8081/foglamp/repository -d '{"url": "http://archives.fledge-iot.org", "version": "nightly"}'

$ curl -sX POST localhost:8081/foglamp/repository -d '{"url": "http://archives.dianomic.com/foglamp", "version": "fixes/FOGL-4133"}'
```

**Pending**
1) RPM Platform - Raise JIRA FOGL-4176

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ data/scripts
 data/plugins
 data/snapshots
 data/logs
+data/configure_repo_output.txt
 
 # SQLite3 default db location and after migration
 data/*.db
--- a/python/fledge/services/core/api/repos/configure.py
+++ b/python/fledge/services/core/api/repos/configure.py
@@ -77,6 +77,7 @@ async def add_package_repo(request: web.Request) -> web.Response:
         else:
             raise ValueError("{} is not supported".format(_platform))
 
+        # TODO: FOGL-4176 RPM support
         stdout_file_path = _FLEDGE_ROOT + "/data/configure_repo_output.txt"
         cmd = "wget -q -O - {}/KEY.gpg | sudo apt-key add - > {} 2>&1".format(url, stdout_file_path)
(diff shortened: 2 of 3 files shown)


ashish-jabble:
> Only thing is it is strict n the platform and structure (nightly, latest, etc) which we will need to evolve as we add more platforms and will have to be aware of when we bundle the package repository with FogMan.

There is no restriction as such `my_list = ['nightly', 'latest', 'fixes/']` This list only to check semantic version format when version value is different and valid for case like 1.8.0, 1.7.0 etc. So basically error handling of version field

And all platforms which we supported exists under the version.

For example:
http://archives.fledge-iot.org/nightly/{os_name}/
http://archives.fledge-iot.org/latest/{os_name}/
http://archives.fledge-iot.org/fixes/FOGL-XXXX/{os_name}/
http://archives.fledge-iot.org/1.8.0/{os_name}/

> We will need to update the sudo commands we can use also.

I have used commands from our google document for setting up package repository

.gitignore
```diff
@@ -18,6 +18,7 @@ data/scripts
 data/plugins
 data/snapshots
 data/logs
+data/configure_repo_output.txt
```
praveen-garg, line 21: oh, Yes!

praveen-garg:
We should allow to update existing when re-POSTed (or may be PUT) in case of any user error.

PR #124, opened 2020-06-23
improved error message for plugin version if supplied; and repo url s…

praveen-garg:
…lug check

### DIFF
--- a/python/fledge/services/core/api/plugins/install.py
+++ b/python/fledge/services/core/api/plugins/install.py
@@ -75,13 +75,12 @@ async def add_plugin(request: web.Request) -> web.Response:
                 raise ValueError('name param is required')
             version = data.get('version', None)
             if version:
-                delimiter = '.'
-                if str(version).count(delimiter) != 2:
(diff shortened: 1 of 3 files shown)


python/fledge/services/core/api/plugins/install.py
```diff
@@ -75,13 +75,12 @@ async def add_plugin(request: web.Request) -> web.Response:
                 raise ValueError('name param is required')
             version = data.get('version', None)
             if version:
-                delimiter = '.'
-                if str(version).count(delimiter) != 2:
-                    raise ValueError('Plugin semantic version is incorrect; it should be like X.Y.Z')
+                if str(version).count('.') != 2:
+                    raise ValueError('Invalid version; it should be empty to install as per the configured repository version or a valid semantic version X.Y.Z i.e. major.minor.patch')
```
ashish-jabble, line 79: Invalid package version; it should be empty to install if a valid semantic version X.Y.Z i.e. major.minor.patch' is not given

PR #125, opened 2020-06-24
FOGL-4199 Fix issue with mandatory JSON type configuration items

MarkRiddoch:

### DIFF
--- a/python/fledge/common/configuration_manager.py
+++ b/python/fledge/common/configuration_manager.py
@@ -516,8 +516,12 @@ class ConfigurationManager(ConfigurationManagerSingleton):
                         raise TypeError('Unrecognized value name for item_name {}'.format(item_name))
 
                 if 'mandatory' in cat_info[item_name]:
-                    if cat_info[item_name]['mandatory'] == 'true' and not len(new_val.strip()):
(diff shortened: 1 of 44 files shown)


python/fledge/common/configuration_manager.py
```diff
@@ -516,8 +516,12 @@ async def update_configuration_item_bulk(self, category_name, config_item_list):
                         raise TypeError('Unrecognized value name for item_name {}'.format(item_name))
 
                 if 'mandatory' in cat_info[item_name]:
-                    if cat_info[item_name]['mandatory'] == 'true' and not len(new_val.strip()):
-                        raise ValueError("A value must be given for {}".format(item_name))
+                    if cat_info[item_name]['mandatory'] == 'true':
+                        if cat_info[item_name]['type'] == 'JSON':
+                            if not len(new_val):
+                                raise ValueError("A value must be given for {}".format(item_name))
```
ashish-jabble, line 522: if new_val supplies with empty dict i.e {}
It will never be caught inside it as the type of new_val is always be in string. In that case length is 2.
You need to convert python str to dict object

MarkRiddoch: That may be valid for a JSON type of course. I'm tempted to leave this for now

PR #140, opened 2020-07-14
FOGL-4264 - extend service API for management type

ashish-jabble:

- [x] POST - /fledge/service
- [x] GET -  /fledge/service/installed
- [x] POST - /fledge/schedule/process - `We can avoid that work to insert direct entry in init.sql along with upgrade/downgrade scripts`

### DIFF
--- a/python/fledge/services/core/api/service.py
+++ b/python/fledge/services/core/api/service.py
@@ -214,7 +214,7 @@ async def add_service(request):
         if service_type == 'north':
             raise web.HTTPNotAcceptable(reason='north type is not supported for the time being.')
         if service_type not in ['south', 'notification', 'management']:
-            raise web.HTTPBadRequest(reason='Only south, notification, management types are supported.')
+            raise web.HTTPBadRequest(reason='Only south, notification and management types are supported.')
         if plugin is None and service_type == 'south':
             raise web.HTTPBadRequest(reason='Missing plugin property for type south in payload.')
         if plugin and utils.check_reserved(plugin) is False:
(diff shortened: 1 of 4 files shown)


python/fledge/services/core/api/service.py
```diff
@@ -213,8 +213,8 @@ async def add_service(request):
         service_type = str(service_type).lower()
         if service_type == 'north':
             raise web.HTTPNotAcceptable(reason='north type is not supported for the time being.')
-        if service_type not in ['south', 'notification']:
-            raise web.HTTPBadRequest(reason='Only south and notification type are supported.')
+        if service_type not in ['south', 'notification', 'management']:
+            raise web.HTTPBadRequest(reason='Only south, notification, management types are supported.')
```
praveen-garg, line 217: Only south, notification and management types

python/fledge/services/core/api/service.py
```diff
@@ -294,6 +297,13 @@ async def add_service(request):
             for ps in res['rows']:
                 if 'notification_c' in ps['process_name']:
                     raise web.HTTPBadRequest(reason='A Notification service schedule already exists.')
+        # check that management service is not already registered, right now management service LIMIT to 1
+        elif service_type == 'management':
+            # TODO: we may rename check_notification_schedule def to check_schedule_entry
+            res = await check_notification_schedule(storage)
```
praveen-garg, line 303: check_management_schedule

praveen-garg:
available, we can leave until we have deb and nightly repo setup for the same. installed we should be able to test.

python/fledge/services/core/api/service.py
praveen-garg: Yes, let's rename

MarkRiddoch:
I'm not altogether happy with this. Fledge should not have any knowledge of the management service as it is not a service that is offered under fledge. I think I would prefer we open the list of available services up in general rather than have service types in Fledge that don't exist for Fledge.

ashish-jabble:
@MarkRiddoch 
> Fledge should not have any knowledge of the management service as it is not a service that is offered under fledge

Management Service will be installed and listed via **white-labelled**/service endpoints.

>  I think I would prefer we open the list of available services up in general rather than have service types in Fledge that don't exist for Fledge.

Could you please elaborate, How the service will be registered with what type, discover it after installation?

PR #143, opened 2020-07-24
FOGL-4233 plugin available packages with 'north', 'south', 'filter', 'notify', 'rule' types only

ashish-jabble:

@MarkRiddoch fyi, We would also want to see all available packages. 
At the moment we have restricted with supported types
a) For plugins: `http://localhost:8081/fledge/plugins/available?type=` | `'north', 'south', 'filter', 'notify', 'rule'` 
b) For service package we have separate endpoint `http://localhost:8081/fledge/service/available`

We may add new endpoint and get it via `curl -sX GET http://localhost:8081/fledge/packages/available`

### DIFF
--- a/python/fledge/services/core/api/plugins/discovery.py
+++ b/python/fledge/services/core/api/plugins/discovery.py
@@ -74,9 +74,9 @@ async def get_plugins_available(request: web.Request) -> web.Response:
         if package_type and package_type not in ['north', 'south', 'filter', 'notify', 'rule']:
             raise ValueError("Invalid package type. Must be 'north' or 'south' or 'filter' or 'notify' or 'rule'.")
         plugins, log_path = await common.fetch_available_packages(package_type)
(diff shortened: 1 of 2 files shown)


MarkRiddoch:
> Signed-off-by: ashish-jabble [ashish@dianomic.com](mailto:ashish@dianomic.com)
> 
> @MarkRiddoch fyi, We would also want to see all available packages.
> At the moment we have restricted with supported types
> a) For plugins: `http://localhost:8081/fledge/plugins/available?type=` | `'north', 'south', 'filter', 'notify', 'rule'`
> b) For service package we have separate endpoint `http://localhost:8081/fledge/service/available`
> 
> We may add new endpoint and get it via `curl -sX GET http://localhost:8081/fledge/packages/available`

I think having the separate endpoint for services is the way to go. It's unlikely to be dealing with all packages together, much more common to deal with plugins by type or services. It means two calls to get all packages, but I don't see that as a big issue as it would be a rare request.

PR #147, opened 2020-07-30
enhance plugin doc to include:

doug-dianomic:
. required apis by plugin type
. skeleton plugin data message format
. required config options by plugin type
reorder default / loading information to follow developer-relevant data


doug-dianomic:
(In second push I fixed table formatting, removed a couple of fields Mark and I agreed should just disappear, and provisionally fixed the timestamp definitions -- those may still not be quite right (I think my main source of csvplayback isn't right), so I'll either figure that out and let you know I've changed it or you can point me in the right direction...)

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
-- **Filter Plugins** - Filter plugins are used to modify data as it flows through Fledge. Filter plugins may be combined into a set of ordered filters that are applied as a pipeline to either the south ingress service or the north egress task that sends data to external systems.
-- **Notification Rule Plugins** - These are used by the optional notification service in order to evaluate data that flows into the notification service to determine if a notification should be sent.
-- **Notification Delivery Plugins** - These plugins are used by the optional notification service to deliver a notification to a system when a notification rule has triggered. These plugins allow the mechanisms to deliver notifications to be extended.
+- **Filter Plugins** - Filter plugins are used to modify data as it flows through Fledge. One or more filter plugins may compose a pipeline which modifies, either, data flowing out from the South ingress tasks into the Fledge Storage system, or out from Fledge storage into the North egress tasks.
```
MarkRiddoch, line 45: "South ingress service" rather than "South ingress task".

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
+- **Notification Rule Plugins** - Notification plugins evaluate data after it has been entered into the Fledge Storage, before it enters the northbound Filter pipeline(s). If the notification evaluates to True, a rule associated with the condition is executed. (Notification rules plugins require installation of the optional notification service.)
```
MarkRiddoch, line 46: Notification are not evaluated after data has been stored, it happens in parallel. 

"If the notification evaluates to True, a rule associated with the condition is executed." - no. If the notification rule evaluates fires then a notification delivery es executed. Rules can fire for triggering or clearing of the rule condition. 

"(Notification rules plugins require installation of the optional notification service.)"
->
"Notification rule plugins are applied to the notification service, the notification service is an optional service and must be installed before the notification plugins can be used."

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
+- **Notification Delivery Plugins** - These plugins deliver a notification event; events, trigger "reason", and accompanying "message" are specified, and delivered to a given delivery channel (eg., email/sms/...). (Notification delivery plugins require installation of the optional notification service.)
```
MarkRiddoch, line 47: Add a link to the notification delivery plugin list rather than the eg.

As above the bracketed text could be better (I hate text in brackets!).

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
+- **North Plugins** - These plugins take data, originated by South service(s), optionally transformed by filters, optionally triggering notifications, and export those data to the outside world.
```
MarkRiddoch, line 48: Strictly the data does not always come from the south, it can be reading data, or statistics from Fledge itself. Plus of course filters can add data.

"export those data" -> "exports the data", although not sure about the connotations of using "export" to define what happens as the data remains in storage and can be sent to multiple destinations.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
+
+Fledge message contents
+=======================
```
MarkRiddoch, line 51: This will break when we white label the documentation for FogLAMP

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
+- **Filter Plugins** - Filter plugins are used to modify data as it flows through Fledge. One or more filter plugins may compose a pipeline which modifies, either, data flowing out from the South ingress tasks into the Fledge Storage system, or out from Fledge storage into the North egress tasks.
+- **Notification Rule Plugins** - Notification plugins evaluate data after it has been entered into the Fledge Storage, before it enters the northbound Filter pipeline(s). If the notification evaluates to True, a rule associated with the condition is executed. (Notification rules plugins require installation of the optional notification service.)
+- **Notification Delivery Plugins** - These plugins deliver a notification event; events, trigger "reason", and accompanying "message" are specified, and delivered to a given delivery channel (eg., email/sms/...). (Notification delivery plugins require installation of the optional notification service.)
+- **North Plugins** - These plugins take data, originated by South service(s), optionally transformed by filters, optionally triggering notifications, and export those data to the outside world.
+
+Fledge message contents
```
MarkRiddoch, line 50: I think it is better to not use the work "message" as it is not a message as such here. Also avoid using Fledge in titles for the reason given below. How about "Readings Format"

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
+=======================
+Information flows through fledge in "messages". Messages have a few "tags" which have well defined values. Messages may be extended to contain additional tags appropriate to particular plugins and data flows.
```
MarkRiddoch, line 52: It does, but these are not the messages, the messages have more than just a reading in them. What you are describing is the reading that get passed to plugins, not the messages between services.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
+
+Most messages include:
```
MarkRiddoch, line 54: All readings contain...

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
+
++------------+------------------+-------------------------------------------------------------------------------------------------------+
+| Tag        | Type             | Description                                                                                           |
++============+==================+=======================================================================================================+
+| asset      | string           | Name under which this stream of data are placeed in Storage                                           |
++------------+------------------+-------------------------------------------------------------------------------------------------------+
+| timestamp  | string           | Fledge timestamp when this asset was generated                                                        |
```
MarkRiddoch, line 61: Table will break when we white label to FogLAMP due to string length issues. Use non-ascii-art format if you have the word Fledge in the table content.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
++------------+------------------+-------------------------------------------------------------------------------------------------------+
+| key        | string           | unique ID                                                                                             |
```
MarkRiddoch, line 63: This is added by storage and is only visible in the north plugins.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
++------------+------------------+-------------------------------------------------------------------------------------------------------+
+| readings   | dict             | {"<name-1>": <value-1> [, "<name-2>": <value-2>,... ["user_ts": <device-timestamp>]]}                 |
```
MarkRiddoch, line 65: There is a lot more to this as it represents the actual data, we should probably tell the reader that vital bit of information.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
++------------+------------------+-------------------------------------------------------------------------------------------------------+
+
+Fledge plugin configuration
```
MarkRiddoch, line 68: Again tis breaks with white labelling. We can drop the Fledge word here.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
+===========================
+
+Different plugin types (eg., north/south/...) have required configuration entries
```
MarkRiddoch, line 71: They all have required configuration entries. They differ in what is required. The statement is a bit misleading I would suggest.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
+|                        |                        |                      | readonly: "true"                                      |
++------------------------+------------------------+----------------------+-------------------------------------------------------+
+| enable                 | Filter |br|            | Enable/disable       | description: "<Describe plugin being enabled>" |br|   |
+|                        | Notification |br|      | plugin function      | type: "boolean"                                |br|   |
+|                        | Notification delivery  |                      | default: "false"                               |br|   |
+|                        |                        |                      | displayName: "<name for UI display>"           |br|   |
+|                        |                        |                      | order: "<order of display in UI>"                     |
++------------------------+------------------------+----------------------+-------------------------------------------------------+
+| asset                  | South |br|             | Asset where data     | description: "<name of asset being monitored>" |br|   |
```
MarkRiddoch, line 89: This is not mandatory for south plugins.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
+|                        | Notification           | fields are deposited | type: "string"                                 |br|   |
```
MarkRiddoch, line 90: Again notification plugins do not have to have this. Also do you mean notification rules to notification delivery plugins?

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
+|                        |                        |                      | default: ""                                    |br|   |
+|                        |                        |                      | displayName: "<name for UI display>"           |br|   |
+|                        |                        |                      | mandatory: "true"                              |br|   |
+|                        |                        |                      | order: "<order of display in UI>"                     |
++------------------------+------------------------+----------------------+-------------------------------------------------------+
+| source                 | North                  | Choice of data to    | description: "<Resource being forwarded>" |br|        |
+|                        |                        | forward from Foglamp | type: "enumeration"                  |br|             |
```
MarkRiddoch, line 97: Use of Foglamp name in Fledge documentation. Note also the table formatting issue here with white labelling.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
+|                        |                        | (readings or         | options: ["readings", "statistics"]  |br|             |
+|                        |                        | statistics)          | default: "readings"                  |br|             |
+|                        |                        |                      | displayName: "<name for UI display>" |br|             |
+|                        |                        |                      | order: "<order of display in UI>"                     |
++------------------------+------------------------+----------------------+-------------------------------------------------------+
+
+Fledge plugin methods
```
MarkRiddoch, line 104: Drop "Fledge" from title.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
+---------------------
+Different plugin types (eg., north/south/...) have common and distinct APIs they must export.
```
MarkRiddoch, line 106: There are a common set for all plugins and then there are specifics for each type of plugin, I would suggest we are better documenting the specifics with the documentation on those plugin type rather than lumping them into one big table.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
++========================+=================+=======================================================================================================+
+| plugin_info            | all             | Returns the info needed to load the plugin (interface spec, type, etc.)                               |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_init            | all             | Takes the config values; one time initialization; returns opaque handle for this instance             |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_shutdown        | all             | Destroys plugin and related state                                                                     |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_reconfigure     | all             | Replaces existing configuration with new values; may need to call shutdown/init                       |
```
MarkRiddoch, line 119: Not all plugins have a plugin_reconfigure method.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_ingest          | Filter          | Provides data which is modified, then sent on to ingest callback                                      |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_eval            | Notification    | Takes JSON asset document to eval; Returns True if should "notify"                                    |
```
MarkRiddoch, line 123: Notification what? Rule or Delivery?

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_triggers        | Notification    | Returns JSON asset document describing what notification triggers have fired                          |
```
MarkRiddoch, line 125: As above

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_reason          | Notification    | Takes JSON asset document describing why notifications have fired                                     |
```
MarkRiddoch, line 127: As above

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_deliver         | Notification    | Takes name/notification/trigger/message strings to be sent to notification target                     |
```
MarkRiddoch, line 129: As above

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
+|                        | |br| delivery   |                                                                                                       |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_send            | North           | Provides data,input_ref to be sent to North plugin target                                             |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_poll            | South           | Initiates pull (return) of next set of data from south data source                                    |
```
MarkRiddoch, line 134: Only poll type south plugins

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_start           | South           | Initiates async "pumping" of data (typically threaded)                                                |
```
MarkRiddoch, line 136: Hm, not sure about that description.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_register_ingest | South           | Registers callback and ingest "ref" which receive new data as available                               |
```
MarkRiddoch, line 138: Only async south plugins.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,109 @@ Plugins
...
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+
+
+
+
+Existing plugins and plugin extensions
+======================================
+Fledge comes pre-loaded with a number of plugins. Additional plugins may be loaded from the standard Fledge collection, from third pary collections, or from code developed by users.
```
MarkRiddoch, line 146: It comes with a single plugin, the OMF plugin, everything else is optional.

docs/plugin_developers_guide/01_Fledge_plugins.rst
doug-dianomic: OK, mostly your text but how does this look: 

 **Notification plugins evaluate data in parallel with their entry into the Fledge Storage, before it enters the northbound Filter pipeline(s). If the notification "fires", notification delivery is executed. Rules can fire for triggering or clearing of the rule condition.
  Notification rule plugins are applied to the notification service, the notification service is an optional service and must be installed before the notification plugins can be used.**

What I've seen with the bad bearing is that it effects action by returning True from plugin eval. So, are you saying that the plugin_eval returns "deliver" as its true/false result, and that what's actually done isn't a "true" notification per se, just a signal to "do delivery"?  (And that it could signal that either after clearing its condition, when something needs to happen, etc., but that's a up to the delivery logic?)

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -35,42 +35,46 @@ This chapter presents the plugins that are bundled with Fledge, how to write and
...
-- **Notification Delivery Plugins** - These plugins deliver a notification event; events, trigger "reason", and accompanying "message" are specified, and delivered to a given delivery channel (eg., email/sms/...). (Notification delivery plugins require installation of the optional notification service.)
-- **North Plugins** - These plugins take data, originated by South service(s), optionally transformed by filters, optionally triggering notifications, and export those data to the outside world.
+- **Filter Plugins** - Filter plugins are used to modify data as it flows through Fledge. One or more filter plugins may compose a pipeline which modifies, either, data flowing out from the South ingress service into the Fledge Storage system, or out from Fledge storage into the North egress tasks.
+- **Rule Plugins** - Rule plugins evaluate data in parallel with their entry into the Fledge Storage. If a rule "fires", it may trigger or clear the rule condition. If notification delivery is configured, the notification delivery plugin will be called to deliver an event.
```
MarkRiddoch, line 46: Why take the Notification out of the title for rules and leave it for delivery plugins?

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -35,42 +35,46 @@ This chapter presents the plugins that are bundled with Fledge, how to write and
...
-Most messages include:
+All readings containe:
 
 +------------+------------------+-------------------------------------------------------------------------------------------------------+
 | Tag        | Type             | Description                                                                                           |
 +============+==================+=======================================================================================================+
 | asset      | string           | Name under which this stream of data are placeed in Storage                                           |
 +------------+------------------+-------------------------------------------------------------------------------------------------------+
-| timestamp  | string           | Fledge timestamp when this asset was generated                                                        |
+| timestamp  | string           | Timestamp when this asset was generated                                                               |
```
MarkRiddoch, line 61: Strictly I is not the asset that was generated, it is the reading.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -86,26 +90,21 @@ Required configuration entries include:
...
-|                        | Notification           | fields are deposited | type: "string"                                 |br|   |
-|                        |                        |                      | default: ""                                    |br|   |
-|                        |                        |                      | displayName: "<name for UI display>"           |br|   |
-|                        |                        |                      | mandatory: "true"                              |br|   |
-|                        |                        |                      | order: "<order of display in UI>"                     |
-+------------------------+------------------------+----------------------+-------------------------------------------------------+
 | source                 | North                  | Choice of data to    | description: "<Resource being forwarded>" |br|        |
-|                        |                        | forward from Foglamp | type: "enumeration"                  |br|             |
+|                        |                        | forward from system  | type: "enumeration"                  |br|             |
```
MarkRiddoch, line 94: from the system

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -116,38 +115,97 @@ Required APIs include:
...
+--------------------------
+
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_deliver         | Notification    | Takes name/notification/trigger/message strings to be sent to the                                     |
+|                        |                 | `notification delivery types`_ target.                                                                |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_reconfigure     | Notification    | Replaces existing configuration with new values; may need to call shutdown/init                       |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+
+Notification delivery types
```
MarkRiddoch, line 154: No, these are not types of anything they are plugins. The set of plugins is already automatically generated and documented. This just duplicates that in a way that will need manual maintenance.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -116,38 +115,97 @@ Required APIs include:
...
-|                        | |br| delivery   |                                                                                                       |
+| plugin_reconfigure     | Rule            | Replaces existing configuration with new values; may need to call shutdown/init                       |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+
+
+Notification delivery APIs
+--------------------------
+
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_deliver         | Notification    | Takes name/notification/trigger/message strings to be sent to the                                     |
```
MarkRiddoch, line 148: The type is not Notification.

MarkRiddoch (changes requested):

doug-dianomic:
This poor benighted PR was a well intended attempt to improve docs that got mixed up with a well intended attempt to improve logging. 

Unless anyone objects, I propose we close this, revisit docs separately, and rest confident that the logging code has already been checked in.

PR #149, opened 2020-08-04
FOGL-3596: multi db

sstefano69:

Attach, improved error handling:
 
image

*C units tests*
Ubuntu 01:
image

 R4, the 3 failing are related to the averages calculation:
image

up to 8 dbs tested, 2 tables each:

image

attach at the foglamp restart:

image

post / fetch data:
image

image

### DIFF
--- a/C/plugins/storage/sqlite/common/connection_manager.cpp
+++ b/C/plugins/storage/sqlite/common/connection_manager.cpp
@@ -130,7 +130,10 @@ bool ConnectionManager::attachNewDb(std::string &path, std::string &alias)
 	int rc;
 	std::string sqlCmd;
 	sqlite3 *dbHandle;
+	bool result;
+	char *zErrMsg = NULL;
 
+	result = true;
 
 	sqlCmd = "ATTACH DATABASE '" + path + "' AS " + alias + ";";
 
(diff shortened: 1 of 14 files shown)


C/plugins/storage/sqlite/common/connection_manager.cpp
```diff
@@ -125,6 +125,59 @@ Connection *conn = 0;
 	return conn;
 }
 
+bool ConnectionManager::attachNewDb(std::string &path, std::string &alias)
```
MarkRiddoch, line 128: Could you add a function header comment here please.

C/plugins/storage/sqlite/common/readings.cpp
```diff
@@ -408,7 +441,8 @@ int Connection::readingStream(ReadingStream **readings, bool commit)
 	struct timeval start, t1, t2, t3, t4, t5;
 #endif
 
-	const char *sql_cmd = "INSERT INTO  " DB_READINGS ".readings ( asset_code, reading, user_ts ) VALUES  (?,?,?)";
+	// * TODO: the current code should be adapted to use the multi databases/tables implementation
```
MarkRiddoch, line 444: Let's raise a Jira for this so we don't forget. I can imagine the streaming will come back into importance if this uncovers that as the throttle point now.

C/plugins/storage/sqlite/common/readings.cpp
```diff
@@ -741,36 +761,39 @@ int localNReadingsTotal;
...
 				{
-					//# FIXME_I:
-					localNReadingsTotal = readingsId + 1;
-					readingsStmt.resize(localNReadingsTotal, nullptr);
-
-					//# FIXME_I
-					Logger::getLogger()->setMinLevel("debug");
-					Logger::getLogger()->debug("xxx4 readingsStmt resize size :%d: idx :%d: ", localNReadingsTotal, readingsId);
-					Logger::getLogger()->setMinLevel("warning");
+					Logger::getLogger()->warn("Impossible to insert the row for the asset_code :%s: into the readings, row ignored.", asset_code);
```
MarkRiddoch, line 766: Rather than say "Impossible to..." can we make it say "It was not possible to ...". Also if we could say why it was not possible that would be useful.

MarkRiddoch (commented):
A few minor things. I would also suggest putting the readings catalogue in a separate source file as readings.cpp is now way too big.

C/plugins/storage/sqlite/common/connection_manager.cpp
sstefano69: done

C/plugins/storage/sqlite/common/readings.cpp
sstefano69: done, FOGL-4457

C/plugins/storage/sqlite/common/readings.cpp
sstefano69: Text adapted.
that piece of code state that was not possible to insert the row (as a WARNING) the function called immediately before will log the ERROR for the specific issue, log sample: 

``` 
ERROR: attachNewDb - It was not possible to attach the db :/home/foglamp/Development/fledge/data/readings_10.db: to an idle connection, error :too many attached databases - max 10:
WARNING: appendReadings - It was not possible to insert the row for the asset_code :msec_011_OK: into the readings, row ignored.
```

PR #151, opened 2020-08-09
Fogl 4387

doug-dianomic:
stdout -> stderr for CONSOLE logging
add the ability to set a default destination for all loggers (for debugging)
allow levels to be inherited if not explicitly set.

(support for using modules both in server and in interactive context...)

### DIFF
--- a/docs/plugin_developers_guide/01_Fledge_plugins.rst
+++ b/docs/plugin_developers_guide/01_Fledge_plugins.rst
@@ -41,11 +41,117 @@ Plugins
 In this version of Fledge you have six types of plugins:
 
 - **South Plugins** - They are responsible for communication between Fledge and the sensors and actuators they support. Each instance of a Fledge South microservice will use a plugin for the actual communication to the sensors or actuators that that instance of the South microservice supports.
-- **North Plugins** - They are responsible for taking reading data passed to them from the South bound service and doing any necessary conversion to the data and providing the protocol to send that converted data to a north-side task.
 - **Storage Plugins** - They sit between the Storage microservice and the physical data storage mechanism that stores the Fledge configuration and readings data. Storage plugins differ from other plugins in that they are written exclusively in C/C++, however they share the same common attributes and entry points that the other filter must support.
-- **Filter Plugins** - Filter plugins are used to modify data as it flows through Fledge. Filter plugins may be combined into a set of ordered filters that are applied as a pipeline to either the south ingress service or the north egress task that sends data to external systems.
-- **Notification Rule Plugins** - These are used by the optional notification service in order to evaluate data that flows into the notification service to determine if a notification should be sent.
-- **Notification Delivery Plugins** - These plugins are used by the optional notification service to deliver a notification to a system when a notification rule has triggered. These plugins allow the mechanisms to deliver notifications to be extended.
+- **Filter Plugins** - Filter plugins are used to modify data as it flows through Fledge. One or more filter plugins may compose a pipeline which modifies, either, data flowing out from the South ingress tasks into the Fledge Storage system, or out from Fledge storage into the North egress tasks.
+- **Notification Rule Plugins** - Notification plugins evaluate data after it has been entered into the Fledge Storage, before it enters the northbound Filter pipeline(s). If the notification evaluates to True, a rule associated with the condition is executed. (Notification rules require installation of the optional notification service.)
+- **Notification Delivery Plugins** - These plugins deliver a notification event; events, trigger "reason", and accompanying "message" are specified, and delivered to a given delivery channel (eg., email/sms/...). (Notification rules require installation of the optional notification service.)
+- **North Plugins** - These plugins take data, originated by South service(s), optionally transformed by filters, optionally triggering notifications, and export those data to the outside world.
+
+Fledge message contents
+=======================
+Information flows through fledge in "messages". Messages have a few "tags" which have well defined values. Messages may be extended to contain additional tags appropriate to particular plugins and data flows.
+
+Most messages include:
++------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| Tag        | Type            | Description                                                                                           |
++============+=================+============+==========================================================================================+
+| asset      | string          | Name under which this stream of data are placeed in Storage                                           |
++------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| ts         | string          | Timestamp when this asset was generated                                                               |
++------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| id         | string          | unique ID                                                                                             |
++------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| readings   | single dict |br|| {"user_ts": timestamp, <name>: <data-value>}  |br|                                                    |
+|            |  -or-    |br|   |                                               |br|                                                    |
+|            | json            | "data": "[{"user_ts": <timestamp1>; <name>: <data-value1>}, {"user_ts": <timestamp2>, <name>...]"     |
++------------+-----------------+-------------------------------------------------------------------------------------------------------+
+
+Fledge plugin configuration
+=======================
+Different plugin types (eg., north/south/...) have required configuration entries
+
+Required configuration entries include:
++------------------------+--------------------------+----------------------------------------------------------------------------------------------+
+| Entry                  | Required by              | Description                                                                                  |
++========================+==========================+==============================================================================================+
+| plugin                 | all                      | description: "<Describe plugin>" |br|                                                        |
+|                        |                          | type: "string" |br|                                                                          |
+|                        |                          | name: "<plugin name>" |br|                                                                   |
+|                        |                          | readonly: "true"      |br|                                                                   |
++------------------------+--------------------------+----------------------------------------------------------------------------------------------+
+| enable                 | filter                   | description: "<Describe plugin being enabled>" |br|                                          |
+|                        | |br| notification        | type: "boolean"                                |br|                                          |
+|                        | |br|notification delivery| default: "false"                               |br|                                          |
+|                        |                          | displayName: "<name for UI display>"           |br|                                          |
+|                        |                          | order: "<order of display in UI>"                                                            |
++------------------------+--------------------------+----------------------------------------------------------------------------------------------+
+| asset                  | south                    | description: "<name of asset being monitored>" |br|                                          |
+|                        | |br| notification        | type: "string"                                 |br|                                          |
+|                        |                          | default: ""                                    |br|                                          |
+|                        |                          | displayName: "<name for UI display>"           |br|                                          |
+|                        |                          | mandatory: "true"                              |br|                                          |
+|                        |                          | order: "<order of display in UI>"                                                            |
++------------------------+--------------------------+----------------------------------------------------------------------------------------------+
+| source                 | north                    | description: "<Describe resource being forwarded>" |br|                                      |
+|                        |                          | type: "enumeration"                                |br|                                      |
+|                        |                          | options: ["readings", "statistics"]                |br|                                      |
+|                        |                          | default: "readings"                                |br|                                      |
+|                        |                          | displayName: "<name for UI display>"               |br|                                      |
+|                        |                          | order: "<order of display in UI>"                                                            |
++------------------------+--------------------------+----------------------------------------------------------------------------------------------+
+| applyFilter            | north                    | description: "<Describe plugin being enabled>"     |br|                                      |
+|                        |                          | type: "boolean"                                    |br|                                      |
+|                        |                          | default: "false"                                   |br|                                      |
+|                        |                          | displayName: "<name for UI display>"               |br|                                      |
+|                        |                          | order: "<order of display in UI>"                                                            |
++------------------------+--------------------------+----------------------------------------------------------------------------------------------+
+| filterRule             | north                    | description: "<describe JQ north filter rule>"     |br|                                      |
+|                        |                          | type: "string"                                     |br|                                      |
+|                        |                          | default: "[]"                                      |br|                                      |
+|                        |                          | displayName: "<name for UI display>"               |br|                                      |
+|                        |                          | order: "<order of display in UI>"                                                            |
++------------------------+--------------------------+----------------------------------------------------------------------------------------------+
+
+Fledge plugin methods
+=======================
+Different plugin types (eg., north/south/...) have common and distinct APIs they must export.
+
+Required APIs include:
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| Entry                  | Required by     | Description                                                                                           |
++========================+=================+============+==========================================================================================+
+| plugin_info            | all             | Returns the info needed to load the plugin (interface spec, type, etc.)                               |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_init            | all             | Takes the config values; one time initialization; returns opaque handle for this instance             |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_shutdown        | all             | Destroys plugin and related state                                                                     |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_reconfigure     | all             | Replaces existing configuration with new values; may need to call shutdown/init                       |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_ingest          | filter          | Provides data which is modified, then sent on to ingest callback                                      |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_eval            | notification    | Takes JSON asset document to eval; Returns True if should "notify"                                    |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_triggers        | notification    | Returns JSON asset document describing what notification triggers have fired                          |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_reason          | notification    | Takes JSON asset document describing why notifications have fired                                     |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_deliver         | notification    | Takes name/notification/trigger/message strings to be sent to notification target                     |
+|                        | |br| delivery   |                                                                                                       |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_send            | north           | Provides data,input_ref to be sent to North plugin target                                             |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_start           | south           | Initiates async "pumping" of data (typically threaded)                                                |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_register_ingest | south           | Registers callback and ingest "ref" which receive new data as available                               |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+
+
+
+
+Existing plugins and plugin extensions
+======================================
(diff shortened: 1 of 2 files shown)


python/fledge/common/logger.py
```diff
@@ -23,12 +23,18 @@
 
 """
 CONSOLE = 1
-"""Send log entries to STDOUT"""
+"""Send log entries to STDERR"""
 
+FLEDGE_DESTINATION_ENV='FLEDGE_LOG_DEVICE'
```
praveen-garg, line 28: unused?

python/fledge/common/logger.py
```diff
@@ -76,19 +82,27 @@ def _get_process_name():
...
 
+    # if no destination is set, use the fledge default
+    if destination == None:
+        destination = default_destination
+
     if destination == SYSLOG:
         handler = SysLogHandler(address='/dev/log')
     elif destination == CONSOLE:
-        handler = logging.StreamHandler(sys.stdout)
+        handler = logging.StreamHandler(sys.stderr)
```
praveen-garg, line 92: probably we can remove this elif, default StreamHandler is stderr

python/fledge/common/logger.py
```diff
@@ -76,19 +82,27 @@ def _get_process_name():
...
-    # TODO: Consider using %r with message when using syslog .. \n looks better than #
     process_name = _get_process_name()
+    # TODO: Consider using %r with message when using syslog .. \n looks better than #
     fmt = '{}[%(process)d] %(levelname)s: %(module)s: %(name)s: %(message)s'.format(process_name)
     formatter = logging.Formatter(fmt=fmt)
+
     handler.setFormatter(formatter)
-    logger.setLevel(level)
-    logger.propagate = propagate
+    if level != None:
```
praveen-garg, line 102: `if level is not None:`  or just `if level`

python/fledge/common/logger.py
```diff
@@ -76,19 +82,27 @@ def _get_process_name():
 
     logger = logging.getLogger(logger_name)
 
+    # if no destination is set, use the fledge default
+    if destination == None:
```
praveen-garg, line 86: if destination is None:

python/fledge/common/logger.py
doug-dianomic: explicitly checking against None to differentiate between "no level provided" and "level set to zero" cases.

python/fledge/common/logger.py
doug-dianomic: done

python/fledge/common/logger.py
doug-dianomic: the ENV is used -- it's an env variable that's queried to see if there's a default set when starting a new process, this allows the process to load the module, determine which env variable to check, then reset the default if set. We could also have the logging module check the ENV and reset the default itself when it loads? (I chose the option of letting clients do it as a little less invasive, but, probably it makes more sense for the module to do it, itself?)

python/fledge/common/logger.py
doug-dianomic: Changed the == to "is"

doug-dianomic:
changes related to comments above. PTAL.

python/fledge/common/logger.py
praveen-garg: To set as an env variable, I think os.environ['FLEDGE_LOGS_DESTINATION'] = Should be used. And better to add some docstring comment to explain the need and usage.


> FLEDGE_DESTINATION_ENV better named as FLEDGE_LOGS_DESTINATION_ENV or just FLEDGE_LOGS_DESTINATION.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,136 @@ Plugins
...
+| plugin_poll            | South           | Initiates pull (return) of next set of data from south data source.                                   |
+|                        |                 | Only applicable for "poll" mode South plugins.                                                        |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_start           | South           | Starts components needed to service async mode South operation, such as service threads.              |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_register_ingest | South           | Registers callback and ingest "ref" which receive new data as available                               |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+
+Notification delivery types
```
praveen-garg, line 143: These changes should have been outside the scope of this pull request. Probably you checked out branch from another topic branch instead of develop.

However,

- docs/scripts generate docs for plugins, you can see generated docs at https://fledge-iot.readthedocs.io/en/latest/fledge_plugins.html

-  There also the explanation of each plugin type is given, probably we want here to differentiate developer vs user guide.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,136 @@ Plugins
...
+- **Filter Plugins** - Filter plugins are used to modify data as it flows through Fledge. One or more filter plugins may compose a pipeline which modifies, either, data flowing out from the South ingress service into the Fledge Storage system, or out from Fledge storage into the North egress tasks.
+- **Notification Rule Plugins** - Notification plugins evaluate data in parallel with their entry into the Fledge Storage, before it enters the northbound Filter pipeline(s). If the notification "fires", notification delivery is executed. Rules can fire for triggering or clearing of the rule condition.
+  Notification rule plugins are applied to the notification service, the notification service is an optional service and must be installed before the notification plugins can be used.
+- **Notification Delivery Plugins** - These plugins deliver a notification event; events, trigger "reason", and accompanying "message" are specified, and delivered to a given [notification delivery type](#notification-delivery-types). Notification delivery plugins require installation of the optional notification service.
```
praveen-garg, line 48: the common info can be added as an RST note directive. I mean for, require installation of the optional notification ... text

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,136 @@ Plugins
...
++------------------+-------------------------------------------------------------------------------------------------------+
+| slack            | Sends a message to be delivered to a Slack channel.                                                   |
++------------------+-------------------------------------------------------------------------------------------------------+
+| telegram         | Sends a message to be delivered to the telegram secure messaging platform.                            |
++------------------+-------------------------------------------------------------------------------------------------------+
+
+
+Existing plugins and plugin extensions
+======================================
+Fledge comes with a number of plugins in its main repository. Additional plugins may be loaded from the standard Fledge collection, from third pary collections, or from code developed by users.
```
praveen-garg, line 173: pary -> party*

docs/plugin_developers_guide/01_Fledge_plugins.rst
doug-dianomic: Oh dang. Yes, they should have been submitted in a different branch. I'll reset them on this one. My git-fu is not strong.

docs/plugin_developers_guide/01_Fledge_plugins.rst
doug-dianomic: Good catch, thanks.

python/fledge/common/logger.py
doug-dianomic: OK, sounds good, will change.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,136 @@ Plugins
...
-- **Notification Rule Plugins** - These are used by the optional notification service in order to evaluate data that flows into the notification service to determine if a notification should be sent.
-- **Notification Delivery Plugins** - These plugins are used by the optional notification service to deliver a notification to a system when a notification rule has triggered. These plugins allow the mechanisms to deliver notifications to be extended.
+- **Filter Plugins** - Filter plugins are used to modify data as it flows through Fledge. One or more filter plugins may compose a pipeline which modifies, either, data flowing out from the South ingress service into the Fledge Storage system, or out from Fledge storage into the North egress tasks.
+- **Notification Rule Plugins** - Notification plugins evaluate data in parallel with their entry into the Fledge Storage, before it enters the northbound Filter pipeline(s). If the notification "fires", notification delivery is executed. Rules can fire for triggering or clearing of the rule condition.
```
MarkRiddoch, line 46: There is no guarantee of timing between notifications and north egress, the two things are asynchronous. I think what you are trying to say is the notifications are not able to see any modification made by filters in the egress, since egress filters are specific to the output channel.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,136 @@ Plugins
...
-- **Filter Plugins** - Filter plugins are used to modify data as it flows through Fledge. Filter plugins may be combined into a set of ordered filters that are applied as a pipeline to either the south ingress service or the north egress task that sends data to external systems.
-- **Notification Rule Plugins** - These are used by the optional notification service in order to evaluate data that flows into the notification service to determine if a notification should be sent.
-- **Notification Delivery Plugins** - These plugins are used by the optional notification service to deliver a notification to a system when a notification rule has triggered. These plugins allow the mechanisms to deliver notifications to be extended.
+- **Filter Plugins** - Filter plugins are used to modify data as it flows through Fledge. One or more filter plugins may compose a pipeline which modifies, either, data flowing out from the South ingress service into the Fledge Storage system, or out from Fledge storage into the North egress tasks.
```
MarkRiddoch, line 45: The north stuff is a little misleading, it makes it sound like there is a filtering process that happens before it goes to all north's, where in fact it is as it goes to each north. Every north has a unique filter pipeline, there is no sharing.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,136 @@ Plugins
...
+- **Notification Rule Plugins** - Notification plugins evaluate data in parallel with their entry into the Fledge Storage, before it enters the northbound Filter pipeline(s). If the notification "fires", notification delivery is executed. Rules can fire for triggering or clearing of the rule condition.
+  Notification rule plugins are applied to the notification service, the notification service is an optional service and must be installed before the notification plugins can be used.
+- **Notification Delivery Plugins** - These plugins deliver a notification event; events, trigger "reason", and accompanying "message" are specified, and delivered to a given [notification delivery type](#notification-delivery-types). Notification delivery plugins require installation of the optional notification service.
+- **North Plugins** - These plugins take data moving through Fledge and exports the data to the outside world. Data may originate from South plugins, from the Fledge system, or be added by filters.
```
MarkRiddoch, line 49: The plugins don't take anything, they are run by the north tasks to send data to external systems. The data originates from south services or internal Fledge data.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,136 @@ Plugins
...
+
+Message contents
```
MarkRiddoch, line 51: "Message" is the wrong noun to use here, the text is describing a reading not a message of any type.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,136 @@ Plugins
...
+Most messages include:
+
++------------+------------------+-------------------------------------------------------------------------------------------------------+
+| Tag        | Type             | Description                                                                                           |
++============+==================+=======================================================================================================+
+| asset      | string           | Name under which this stream of data are placeed in Storage                                           |
++------------+------------------+-------------------------------------------------------------------------------------------------------+
+| timestamp  | string           | Fledge timestamp when this asset was generated                                                        |
++------------+------------------+-------------------------------------------------------------------------------------------------------+
+| key        | string           | unique ID                                                                                             |
```
MarkRiddoch, line 64: This is added by the storage when a reading is stored. Readings only have an ID when you view them via the REST API or in north tasks.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,136 @@ Plugins
...
++------------+------------------+-------------------------------------------------------------------------------------------------------+
+| readings   | dict             | {"<name-1>": <value-1> [, "<name-2>": <value-2>,... ["user_ts": <device-timestamp>]]}                 |
```
MarkRiddoch, line 66: User_ts is not at this level it is at the root of a reading.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,136 @@ Plugins
...
+|                        |                        |                      | readonly: "true"                                      |
++------------------------+------------------------+----------------------+-------------------------------------------------------+
+| enable                 | Filter |br|            | Enable/disable       | description: "<Describe plugin being enabled>" |br|   |
+|                        | Notification |br|      | plugin function      | type: "boolean"                                |br|   |
+|                        | Notification delivery  |                      | default: "false"                               |br|   |
+|                        |                        |                      | displayName: "<name for UI display>"           |br|   |
+|                        |                        |                      | order: "<order of display in UI>"                     |
++------------------------+------------------------+----------------------+-------------------------------------------------------+
+| asset                  | South |br|             | Asset where data     | description: "<name of asset being monitored>" |br|   |
```
MarkRiddoch, line 90: There is nothing mandatory about having an asset configuration item in the south.

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,136 @@ Plugins
...
++========================+=================+=======================================================================================================+
+| plugin_info            | all             | Returns the info needed to load the plugin (interface spec, type, etc.)                               |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_init            | all             | Takes the config values; one time initialization; returns opaque handle for this instance             |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_shutdown        | all             | Destroys plugin and related state                                                                     |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_reconfigure     | all             | Replaces existing configuration with new values; may need to call shutdown/init                       |
```
MarkRiddoch, line 120: No, not used in the north or notification delivery

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,136 @@ Plugins
...
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_ingest          | Filter          | Provides data which is modified, then sent on to ingest callback                                      |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_eval            | Notification    | Takes JSON asset document to eval; Returns True if should "notify"                                    |
```
MarkRiddoch, line 124: Notification rules

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,136 @@ Plugins
...
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_triggers        | Notification    | Returns JSON asset document describing what notification triggers have fired                          |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_reason          | Notification    | Takes JSON asset document describing why notifications have fired                                     |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_deliver         | Notification    | Takes name/notification/trigger/message strings to be sent to                                         |
```
MarkRiddoch, line 130: notification delivery

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,136 @@ Plugins
...
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_reconfigure     | all             | Replaces existing configuration with new values; may need to call shutdown/init                       |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_ingest          | Filter          | Provides data which is modified, then sent on to ingest callback                                      |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_eval            | Notification    | Takes JSON asset document to eval; Returns True if should "notify"                                    |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_triggers        | Notification    | Returns JSON asset document describing what notification triggers have fired                          |
```
MarkRiddoch, line 126: notification rule

docs/plugin_developers_guide/01_Fledge_plugins.rst
```diff
@@ -41,11 +41,136 @@ Plugins
...
+| plugin_start           | South           | Starts components needed to service async mode South operation, such as service threads.              |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+| plugin_register_ingest | South           | Registers callback and ingest "ref" which receive new data as available                               |
++------------------------+-----------------+-------------------------------------------------------------------------------------------------------+
+
+Notification delivery types
+---------------------------
+Notifications can be delivered through a variety of media.
+
+Notification delivery types include:
```
MarkRiddoch, line 147: There is a generated list that's is always up to date, better to reference his than hard code a list that will become out of date very quickly.

praveen-garg (approved):
logger part looks okay to me!

python/fledge/common/logger.py
```diff
@@ -23,12 +23,31 @@
...
+        processes. (makes logging consistent for interactive stderr vs server syslog applications """
+    global default_destination
+    default_destination = destination
+    os.environ[FLEDGE_LOGS_DESTINATION] = str(destination)
+
+
+if (FOGLAMP_LOGS_DESTINATION in os.environ) and \
+   os.environ[FOGLAMP_LOGS_DESTINATION] in [str(CONSOLE), str(SYSLOG)]:
+    # inherit (valid) default from the environment
+    set_default_destination(int(os.environ[FOGLAMP_LOGS_DESTINATION]))
```
praveen-garg, line 44: FLEDGE?

MarkRiddoch (changes requested):
Please remove the documentation changes by replacing this file with the one from develop. It's a different PR.

I question why we are doing this change, we really, really do not want the possibility that in production a fledge/FogLAMP could direct the logs anywhere other than the place we expect them. It breaks the support bundle implementation and risks logs being lost completely.

Inheriting the log level I agree with, it should be set globally for the service in the way the C++ services handle it. Individual Python files should not have a setting of the minimum log level.

doug-dianomic:
Yes, I'll definitely get the docs changes out of the way -- I think Praveen
has given me the recipe. My git competence is growing more slowly than I'd
prefer.

The other part of the logging change is (proposed) so that I can use our
standard logging code when I run plugins from a user application, where I'd
want to use stderr. The *intent* is only to make it possible to send all
logging to stderr or all logging to syslog, not to pick and choose or
modify the target later. The console/syslog option was there before, it
just used stdout not stderr.

I can do something different with the logging if it is too worrisome the
way it is. But, the standard python logging code lets you set the level in
each handler as well as on the logger itself, so there's some basic
messed-up-edness that's built in, thanks to Python.

  Doug

MarkRiddoch (commented):
This still appears to have the documentation changes mixed with the logger changes

doug-dianomic:
Yes, sorry. I'll delete that branch if you want to delete the JIRA, and
start over with the doc changes. I believe the logger changes got folded
in, in a different branch cleaning that out.

praveen-garg:
https://github.com/fledge-iot/fledge/pull/165/files 

https://github.com/fledge-iot/fledge/pull/147/files

PR #155, opened 2020-08-25
GCP system test fixes

YashTatkondawar:


praveen-garg (approved):
lgtm! Probably we would need a long running test next, verifying latest 5 on regular interval

PR #158, opened 2020-08-26
FOGL-4444 Out of index range fixes in south API when schedule entry is missing for matching South child category

ashish-jabble:

```
// http://localhost:8081/foglamp/south?cached=false

{
  "services": [
    {
      "name": "sinusoid",
      "address": "",
      "management_port": "",
      "service_port": "",
      "protocol": "",
      "status": "",
      "assets": [
        
      ],
      "plugin": {
        "name": "",
        "version": ""
      },
      "schedule_enabled": "unknown"
    }
  ]
}
```

### DIFF
--- a/python/fledge/services/core/api/south.py
+++ b/python/fledge/services/core/api/south.py
@@ -91,7 +91,12 @@ async def _services_with_assets(storage_client, south_services):
                     if p["name"] == plugin:
                         plugin_version = p["version"]
                         break
-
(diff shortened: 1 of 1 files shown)


praveen-garg (approved):

MarkRiddoch (approved):

praveen-garg:
This is an interesting case, We should check how GUI respond to this. When there is only category but no schedule (hence no service info at all). What does /service endpoint return and should south return `"services": []` in such case?!

PR #164, opened 2020-09-10
Fogl 4524

doug-dianomic:
logger changes to allow common defaults, use stderr as a default
add iprpc as a common service
change cmake rules to make python3 libs more explicit


doug-dianomic:
Re-applied logger changes to a new, clean branch, plus iprpc. I'll sort out the docs changes that got mixed in, before, separately.

PR #185, opened 2020-10-14
FOGL-4598 update plugin unit tests coverage added

ashish-jabble:

### DIFF
--- a/python/fledge/services/core/api/plugins/update.py
+++ b/python/fledge/services/core/api/plugins/update.py
@@ -65,7 +65,7 @@ async def update_plugin(request: web.Request) -> web.Response:
         # if status is -1 (Already in progress) then return as rejected request
         result_payload = {}
         action = 'update'
-        package_name = "fledge-{}-{}".format(_type, name.lower())
+        package_name = "fledge-{}-{}".format(_type, name.lower().replace('_', '-'))
         storage_client = connect.get_storage_async()
         select_payload = PayloadBuilder().SELECT("status").WHERE(['action', '=', action]).AND_WHERE(
             ['name', '=', package_name]).payload()
@@ -133,7 +133,6 @@ async def update_plugin(request: web.Request) -> web.Response:
                             _logger.warning("Disabling {} {} instance, as {} plugin is updating...".format(
                                 p['service'], _type, name))
                             sch_list.append(sch_info[0]['id'])
-
         # Insert record into Packages table
         insert_payload = PayloadBuilder().INSERT(id=str(uuid.uuid4()), name=package_name, action=action, status=-1,
                                                  log_file_uri="").payload()
@@ -169,7 +168,7 @@ async def update_plugin(request: web.Request) -> web.Response:
     except Exception as ex:
         raise web.HTTPInternalServerError(reason=str(ex))
     else:
-        return web.json_response({'message': result_payload})
+        return web.json_response(result_payload)
 
 
 async def _get_plugin_and_sch_name_from_asset_tracker(_type: str) -> list:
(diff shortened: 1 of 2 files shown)


python/fledge/services/core/api/plugins/update.py
```diff
@@ -65,7 +65,7 @@ async def update_plugin(request: web.Request) -> web.Response:
         # if status is -1 (Already in progress) then return as rejected request
         result_payload = {}
         action = 'update'
-        package_name = "fledge-{}-{}".format(_type, name.lower())
+        package_name = "fledge-{}-{}".format(_type, name.lower().replace('_', '-'))
```
praveen-garg, line 68: not sure, if we should replace _ with - where we expect package name only (not the plugin name); may be better to reject if such plugin is not installed and let client fix it.

tests/unit/python/fledge/services/core/api/plugins/test_update.py
```diff
@@ -0,0 +1,381 @@
...
+    @pytest.mark.parametrize("param", [
+        "blah",
+        1,
+        "notificationDelivery"
+        "notificationRule"
+    ])
+    async def test_bad_type_plugin(self, client, param):
+        resp = await client.put('/fledge/plugins/{}/name/update'.format(param), data=None)
+        assert 400 == resp.status
+        assert "Invalid plugin type. Must be 'north' or 'south' or 'filter' or 'notify' or 'rule'" == resp.reason
```
praveen-garg, line 50: May be better message is `Must be one of 'south' , north', 'filter', 'notify' or 'rule'" `

tests/unit/python/fledge/services/core/api/plugins/test_update.py
```diff
@@ -0,0 +1,381 @@
...
+                "status": -1,
+                "log_file_uri": ""
+            }]}
+        svc_name = 'R1'
+        tracked_plugins = [{'plugin': 'sinusoid', 'service': 'S1'}, {'plugin': 'Random', 'service': svc_name}]
+        sch_info = [{'id': '6637c9ff-7090-4774-abca-07dee59a0610', 'enabled': 't'}]
+        server.Server.scheduler = Scheduler(None, None)
+        storage_client_mock = MagicMock(StorageClientAsync)
+        with patch.object(connect, 'get_storage_async', return_value=storage_client_mock):
+            with patch.object(storage_client_mock, 'query_tbl_with_payload',
```
praveen-garg, line 199: are we patching insert? should not insert come from actual code and we patch output of storage?

tests/unit/python/fledge/services/core/api/plugins/test_update.py
```diff
@@ -0,0 +1,381 @@
...
+                                                      return_value=async_mock(insert)) as insert_tbl_patch:
+                                        with patch('multiprocessing.Process'):
+                                            resp = await client.put('/fledge/plugins/{}/{}/update'.format(
+                                                plugin_type, plugin_installed_dirname), data=None)
+                                            server.Server.scheduler = None
+                                            assert 200 == resp.status
+                                            result = await resp.text()
+                                            response = json.loads(result)
+                                            assert 'id' in response
+                                            assert 'Plugin update started.' == response['message']
```
praveen-garg, line 221: May be we should return `<pkg name>` instead of `Plugin `

tests/unit/python/fledge/services/core/api/plugins/test_update.py
```diff
@@ -0,0 +1,381 @@
...
+                                    assert 'packages' == args[0]
+                                    actual = json.loads(args[1])
+                                    assert 'id' in actual
+                                    assert pkg_name == actual['name']
+                                    assert 'update' == actual['action']
+                                    assert -1 == actual['status']
+                                    assert '' == actual['log_file_uri']
+                                assert 1 == log_warn_patch.call_count
+                                log_warn_patch.assert_called_once_with(
+                                    'Disabling {} {} instance, as {} plugin is updating...'.format(
```
praveen-garg, line 233: `updating` -> `is being updated`

praveen-garg (approved):
looks okay to me. Few more things we can do: 
1) parametrize for south, north
2) add special for filter 
3) parametrize for rule and delivery
4) add 2 more test for failed, and success verify status link
I would suggest we should add test modules with stubs only in a flow and lots of comments there we want to verify/ put e.g. sch disable/enable, config category disable/enable and then when manually tested then fill those tests per action

praveen-garg (approved):

PR #187, opened 2020-10-14
FOGL-4599 Plugin remove unit tests coverage added

ashish-jabble:

### DIFF
--- a/tests/unit/python/fledge/services/core/api/plugins/test_remove.py
+++ b/tests/unit/python/fledge/services/core/api/plugins/test_remove.py
@@ -1,0 +1,261 @@
+# -*- coding: utf-8 -*-
+
+# FLEDGE_BEGIN
+# See: http://fledge.readthedocs.io/
+# FLEDGE_END
+
+import json
+from unittest.mock import patch, MagicMock
+import pytest
+
+from aiohttp import web
+
+from fledge.services.core import routes
+from fledge.services.core import connect
+from fledge.services.core.api.plugins import remove as plugins_remove
+from fledge.services.core.api.plugins.exceptions import *
+from fledge.common.storage_client.storage_client import StorageClientAsync
+from fledge.common.plugin_discovery import PluginDiscovery
+
+
+__author__ = "Ashish Jabble"
+__copyright__ = "Copyright (c) 2020 Dianomic Systems Inc."
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+@pytest.allure.feature("unit")
+@pytest.allure.story("api", "plugins", "remove")
+class TestPluginRemove:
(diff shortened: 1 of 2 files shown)


tests/unit/python/fledge/services/core/api/plugins/test_remove.py
```diff
@@ -0,0 +1,261 @@
...
+    async def test_package_when_not_in_use(self, client):
+        async def async_mock(return_value):
+            return return_value
+        _type = "south"
+        name = 'http_south'
+        pkg_name = "fledge-south-http-south"
+        payload = {"return": ["status"], "where": {"column": "action", "condition": "=", "value": "purge",
+                                                   "and": {"column": "name", "condition": "=", "value": pkg_name}}}
+        select_row_resp = {'count': 1, 'rows': [{
+            "id": "c5648940-31ec-4f78-a7a5-b1707e8fe578",
```
praveen-garg, line 196: this for last failure call check?  ID does not matter as patched here, but should be unique on new request insert

tests/unit/python/fledge/services/core/api/plugins/test_remove.py
```diff
@@ -0,0 +1,261 @@
...
+        }]}
+        insert = {"response": "inserted", "rows_affected": 1}
+        insert_row = {'count': 1, 'rows': [{
+                "id": "c5648940-31ec-4f78-a7a5-b1707e8fe578",
+                "name": pkg_name,
+                "action": "purge",
+                "status": -1,
+                "log_file_uri": ""
+            }]}
+        delete = {"response": "deleted", "rows_affected": 1}
```
praveen-garg, line 210: this is to verify that select_row_resp that last failed entry gets removed on new request?

praveen-garg (approved):
We need to ensure, we update `available` cache when process completes with exit code 0.

tests/unit/python/fledge/services/core/api/plugins/test_remove.py
ashish-jabble: It is for the deletion of entry from database only when status code is Not in progress mode (i.e -1)

tests/unit/python/fledge/services/core/api/plugins/test_remove.py
ashish-jabble: This is the return value (And expected one). For return values as close as with real storage result.

tests/unit/python/fledge/services/core/api/plugins/test_remove.py
praveen-garg: So We expect  delete to be called with what we have in `select_row_resp` (failed or success) and insert with new ID for new request if plugin is discoverable. right?

PR #190, opened 2020-10-20
FOGL-4614 - Async implementation for service install & update endpoints along with their unit tests

ashish-jabble:


python/fledge/services/core/api/service.py
```diff
@@ -183,29 +185,72 @@ async def add_service(request):
...
+                    raise ValueError('name should start with "fledge-service-" prefix')
                 version = data.get('version', None)
                 if version:
                     delimiter = '.'
                     if str(version).count(delimiter) != 2:
                         raise ValueError('Service semantic version is incorrect; it should be like X.Y.Z')
 
-                services, log_path = await common.fetch_available_packages("service")
-                if name not in services:
-                    raise KeyError('{} service is not available for the given repository or already installed'.format(name))
```
praveen-garg, line 194: We should check this,

python/fledge/services/core/api/service.py
```diff
@@ -183,29 +185,72 @@ async def add_service(request):
...
+                    if exit_code == -1:
+                        msg = "{} package installation already in progress".format(name)
+                        return web.HTTPTooManyRequests(reason=msg, body=json.dumps({"message": msg}))
+                    # Remove old entry from table for other cases
+                    delete_payload = PayloadBuilder().WHERE(['action', '=', action]).AND_WHERE(
+                        ['name', '=', name]).payload()
+                    await storage.delete_from_tbl("packages", delete_payload)
+
+                # Check If requested service is already installed and then return immediately
+                services = get_service_installed()
```
praveen-garg, line 217: should not we move this up? exit early if nothing to do

python/fledge/services/core/api/service.py
```diff
@@ -183,29 +185,72 @@ async def add_service(request):
...
+                insert_payload = PayloadBuilder().INSERT(id=str(uuid.uuid4()), name=name, action=action, status=-1,
+                                                         log_file_uri="").payload()
+                result = await storage.insert_into_tbl("packages", insert_payload)
+                response = result['response']
+                if response:
+                    # GET id from Packages table to track the installation response
+                    select_payload = PayloadBuilder().SELECT("id").WHERE(['action', '=', action]).AND_WHERE(
+                        ['name', '=', name]).payload()
+                    result = await storage.query_tbl_with_payload('packages', select_payload)
+                    response = result['rows']
```
praveen-garg, line 238: why select after insert, don't we already know the inserted items?

python/fledge/services/core/api/service.py
```diff
@@ -183,29 +185,72 @@ async def add_service(request):
...
+                    if response:
+                        pn = "{}-{}".format(action, name)
+                        uid = response[0]['id']
+                        p = multiprocessing.Process(name=pn, target=install.install_package_from_repo,
+                                                    args=(name, pkg_mgt, version, uid, storage))
+                        p.daemon = True
+                        p.start()
+                        _logger.info("{} service started...".format(name))
```
praveen-garg, line 246: what started :)

python/fledge/services/core/api/service.py
```diff
@@ -454,8 +499,26 @@ async def update_service(request: web.Request) -> web.Response:
         if name not in installed_services:
             raise KeyError("{} service is not installed yet. Hence update is not possible.".format(name))
 
+        # Check Pre-conditions from Packages table
```
praveen-garg, line 502: do we check if already up to date to newest version or try update anyway?

python/fledge/services/core/api/service.py
```diff
@@ -466,54 +529,93 @@ async def update_service(request: web.Request) -> web.Response:
...
+
+def do_update(http_enabled: bool, host: str, port: int, storage: connect, pkg_name: str, uid: str,
+              schedules: list) -> None:
+    _logger.info("{} service update started...".format(pkg_name))
     _platform = platform.platform()
-    stdout_file_path = common.create_log_file("update", name)
+    stdout_file_path = common.create_log_file("update", pkg_name)
     pkg_mgt = 'apt'
     cmd = "sudo {} -y update > {} 2>&1".format(pkg_mgt, stdout_file_path)
+    protocol = "HTTP" if http_enabled else "HTTPS"
```
praveen-garg, line 592: I believe just for future; core server runs on http always

tests/unit/python/fledge/services/core/api/test_service.py
```diff
@@ -608,43 +605,121 @@ async def mock_bad_result():
...
+        svc_list = ["storage", "south", "notification"]
+        msg = '{} package is already installed'.format(pkg_name)
+        storage_client_mock = MagicMock(StorageClientAsync)
+        with patch.object(connect, 'get_storage_async', return_value=storage_client_mock):
+            with patch.object(storage_client_mock, 'query_tbl_with_payload',
+                              return_value=async_mock({'count': 0, 'rows': []})) as query_tbl_patch:
+                with patch.object(service, 'get_service_installed', return_value=svc_list
+                                  ) as svc_list_patch:
+                    resp = await client.post('/fledge/service?action=install', data=json.dumps(param))
+                    assert 304 == resp.status
```
praveen-garg, line 655: if already installed return 400 bad request or 422 not 3xx (user agent specific status codes); same for plugins

python/fledge/services/core/api/service.py
```diff
@@ -183,29 +185,72 @@ async def add_service(request):
...
+                if response:
+                    # GET id from Packages table to track the installation response
+                    select_payload = PayloadBuilder().SELECT("id").WHERE(['action', '=', action]).AND_WHERE(
+                        ['name', '=', name]).payload()
+                    result = await storage.query_tbl_with_payload('packages', select_payload)
+                    response = result['rows']
+                    if response:
+                        pn = "{}-{}".format(action, name)
+                        uid = response[0]['id']
+                        p = multiprocessing.Process(name=pn, target=install.install_package_from_repo,
```
praveen-garg, line 242: Do we upgrade before install if configured cached limit allows? and then update on each call

praveen-garg (changes requested):

python/fledge/services/core/api/service.py
ashish-jabble: This piece of code line moves to next position
https://github.com/fledge-iot/fledge/blob/1131dbab6d86e38dc04ed95be958d0adcc643a79/python/fledge/services/core/api/service.py#L227

python/fledge/services/core/api/service.py
ashish-jabble: INSERT returns only affected rows from storage. And SELECT needs to get id which we have used further...

python/fledge/services/core/api/service.py
ashish-jabble: installation text needs to be placed before

python/fledge/services/core/api/service.py
ashish-jabble: As we do not have the version info details when installed initially. Therefore due to unavailability of previous version we always try to update and if update with same version it returns "Nothing to update"

And in FL world there is NO such update  scenario exists; as we are releasing all plugins/service with same bumped versions.

python/fledge/services/core/api/service.py
ashish-jabble: yupe

python/fledge/services/core/api/service.py
ashish-jabble: upgrade is missing since beginning

python/fledge/services/core/api/service.py
praveen-garg: On success of INSERT, instead of select we can assign/use same id `.INSERT(id=str(uuid.uuid4()),` instead of SELECT?

python/fledge/services/core/api/service.py
praveen-garg: OK

python/fledge/services/core/api/service.py
praveen-garg: I would suggest to keep it  here itself before going to check status from package management table; and return early in this case.

python/fledge/services/core/api/service.py
praveen-garg: Yeah, let's log the `msg` itself (the next line)

praveen-garg (approved):

PR #206, opened 2020-11-12
FOGL-4678: temporary workaround

sstefano69:
**this is a temporary workaround the issue FOGL-4678 not the final solution**


sstefano69:
#obsolte,  the functionality is already developed and merged

PR #216, opened 2020-11-26
FOGL-4689 scheduler process_script corruption fixes which causes scheduler gets stopped

ashish-jabble:

Root cause was found by @praveen-garg 

On POST scheduled process entry scheduler gets stucked (corrupted scheduler process script dict)
Also We don't need to add south to see scheduler gets stopped, it gets corrupted as soon as service agent make an entry for FogMan Agent Poll task.


praveen-garg (approved):

MarkRiddoch (approved):

ashish-jabble:
Also system/unit tests health are okay with custom CI Job for ub1804

PR #235, opened 2020-12-18
FOGL-4821 Move restart of plugin outside of the change callback in or…

MarkRiddoch:
…der to prevent the request timing out


ashish-jabble (approved):
With this change JIRA problem is gone!

PR #236, opened 2020-12-21
FOGL-4814 syslog API enhancements

ashish-jabble:

Notes:
```
Info & Above => info | warning | error | fatal
Warning & Above => warning | error | fatal
Error & Above => error | fatal
debug => all
```

### DIFF
--- a/python/fledge/services/core/api/support.py
+++ b/python/fledge/services/core/api/support.py
@@ -35,13 +35,11 @@ __GET_SYSLOG_CMD_TEMPLATE = "grep -a -E '({})\[' {} | head -n {} | tail -n {}"
 __GET_SYSLOG_CMD_WITH_INFO_TEMPLATE = "grep -a -E '({})\[' {} | grep -a -E -i '(info|warning|error|fatal)' | head -n {} | tail -n {}"
 __GET_SYSLOG_CMD_WITH_ERROR_TEMPLATE = "grep -a -E '({})\[' {} | grep -a -E -i '(error|fatal)' | head -n {} | tail -n {}"
 __GET_SYSLOG_CMD_WITH_WARNING_TEMPLATE = "grep -a -E '({})\[' {} | grep -a -E -i '(warning|error|fatal)' | head -n {} | tail -n {}"
-__GET_SYSLOG_CMD_WITH_DEBUG_TEMPLATE = "grep -a -E '({})\[' {} | grep -a -E -i 'debug' | head -n {} | tail -n {}"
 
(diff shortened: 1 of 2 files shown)


python/fledge/services/core/api/support.py
```diff
@@ -31,18 +32,22 @@
 __DEFAULT_OFFSET = 0
 __DEFAULT_LOG_SOURCE = 'Fledge'
 __GET_SYSLOG_CMD_TEMPLATE = "grep -a -E '({})\[' {} | head -n {} | tail -n {}"
-__GET_SYSLOG_CMD_WITH_ERROR_TEMPLATE = "grep -a -E '({})\[' {} | grep -a -E -i 'error' | head -n {} | tail -n {}"
-__GET_SYSLOG_CMD_WITH_WARNING_TEMPLATE = "grep -a -E '({})\[' {} | grep -a -E -i '(error|warning)' | head -n {} | tail -n {}"
+__GET_SYSLOG_CMD_WITH_INFO_TEMPLATE = "grep -a -E '({})\[' {} | grep -a -E -i '(info|warning|error|fatal)' | head -n {} | tail -n {}"
+__GET_SYSLOG_CMD_WITH_ERROR_TEMPLATE = "grep -a -E '({})\[' {} | grep -a -E -i '(error|fatal)' | head -n {} | tail -n {}"
+__GET_SYSLOG_CMD_WITH_WARNING_TEMPLATE = "grep -a -E '({})\[' {} | grep -a -E -i '(warning|error|fatal)' | head -n {} | tail -n {}"
+__GET_SYSLOG_CMD_WITH_DEBUG_TEMPLATE = "grep -a -E '({})\[' {} | grep -a -E -i 'debug' | head -n {} | tail -n {}"
```
praveen-garg, line 38: No, this is simply __GET_SYSLOG_CMD_TEMPLATE (Debug and above),  as it needs to get all **Not debug only** (debug|info|warning|error|fatal)

praveen-garg (changes requested):
Also better to handle EXCEPTION along with FATAL

python/fledge/services/core/api/support.py
ashish-jabble: There is no option Debug & Above we need to show on GUI as per Mark. This is only stays at API level for development purpose And only get debug statements.

praveen-garg: GUI does not matter here.  it does not make any sense to see DEBUG only grep! What value those logs do have if not with other logs?

ashish-jabble: Fixed as per your suggestions

praveen-garg (approved):

mshariq-nerd (approved):

PR #239, opened 2020-12-22
Fogl 4524.patch

deepanshu-yadav:
Some clean up and a fix of small issue occured when importing eprint.

### DIFF
--- a/python/fledge/common/iprpc.py
+++ b/python/fledge/common/iprpc.py
@@ -18,28 +18,20 @@ import sys
 import io
 import subprocess
 import threading
-
-import time
-
 import traceback
-
-import json
 import pickle
-import re
 import tempfile
-
 import time
 import logging
-
 import fcntl
-from utils import eprint
+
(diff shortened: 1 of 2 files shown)


python/fledge/common/iprpc.py
```diff
@@ -18,28 +18,20 @@
...
-import re
 import tempfile
-
 import time
 import logging
-
 import fcntl
-from utils import eprint
+
+from fledge.common.utils import eprint
```
doug-dianomic, line 28: let's make this a relative path. the more complex the path is, the less flexible the module is.

python/fledge/common/iprpc.py
```diff
@@ -276,9 +251,9 @@ def serve(self):
                 # return the result of the call
                 self.rpc_write(_ret)
 
-
         os._exit(0)
```
doug-dianomic, line 255: I changed this to sys.exit() -- this may not be the latest version of the code?

PR #242, opened 2021-01-04
FOGL-4852: Not getting modified "retrigger_time" in POST

pintomax:
FOGL-4852: : Not getting correct "Retrigger time" in notification creation


praveen-garg (approved):
Interesting. 

retrigger_time config item was added with https://github.com/foglamp/foglamp-service-notification/pull/77/files merge! GUI does not care for it while adding and while updating/ get use category route.
We need to have it in GET (by name and all), and PUT API endpoints. Also need GUI to allow set retrigger time while adding an event engine instance). These should be via separate Jiras.

monika-Dianomic (approved):

PR #243, opened 2021-01-05
FOGL-3921 version info added to ping JSON response along with their unit/system tests

ashish-jabble:

```
// http://192.168.1.9:8081/fledge/ping

{
  "uptime": 59,
  "dataRead": 29,
  "dataSent": 0,
  "dataPurged": 0,
  "authenticationOptional": true,
  "serviceName": "Fledge",
  "hostName": "aj",
  "ipAddresses": [
    "192.168.1.9"
  ],
  "health": "green",
  "safeMode": false,
  "version": "1.8.2"
}
```

**NOTE** -  ~Onwards with bump/next released version we need to update these tests as well (as per released version number) to make sure version info verification is fine with bumped version in /ping~

### DIFF
--- a/python/fledge/services/core/api/common.py
+++ b/python/fledge/services/core/api/common.py
@@ -20,6 +20,7 @@ from fledge.services.core import connect
 from fledge.common.configuration_manager import ConfigurationManager
 from fledge.services.core.service_registry.service_registry import ServiceRegistry
 from fledge.common.service_record import ServiceRecord
+from fledge.common.common import _FLEDGE_ROOT
 
 __author__ = "Amarendra K. Sinha, Ashish Jabble"
 __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
@@ -85,6 +86,12 @@ async def ping(request):
 
     status_color = services_health_litmus_test()
     safe_mode = True if server.Server.running_in_safe_mode else False
+    version = ""
+    with open(_FLEDGE_ROOT + '/VERSION') as f:
+        for line in f:
+            if 'fledge_version=' in line:
+                key, value = line.strip().split('=')
+                version = value
 
     return web.json_response({'uptime': int(since_started),
                               'dataRead': data_read,
@@ -95,7 +102,8 @@ async def ping(request):
                               'hostName': host_name,
                               'ipAddresses': ip_addresses,
                               'health': status_color,
-                              'safeMode': safe_mode
+                              'safeMode': safe_mode,
+                              'version': version
                               })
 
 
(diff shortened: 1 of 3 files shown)


python/fledge/services/core/api/common.py
```diff
@@ -85,6 +86,12 @@ def services_health_litmus_test():
 
     status_color = services_health_litmus_test()
     safe_mode = True if server.Server.running_in_safe_mode else False
+    version = ""
+    with open(_FLEDGE_ROOT + '/VERSION') as f:
+        for line in f:
+            if 'fledge_version=' in line:
+                key, value = line.strip().split('=')
+                version = value
```
praveen-garg, line 94: let's move this to a **cached** def; so that we read only once on each start; 

Ping is already an expensive call; this loop + read file will make it more expensive. I can understand, we want to be sure to read the line having `fledge_version=` tough as we keep it always first line, may be we better use `.readline().split('=')[1].strip()`

tests/system/python/api/test_common.py
```diff
@@ -61,6 +61,7 @@ def test_ping_default(self, reset_and_start_fledge, fledge_url):
         assert 'green' == jdoc['health']
         assert jdoc['authenticationOptional'] is True
         assert jdoc['safeMode'] is False
+        assert '1.8.2' == jdoc['version']
```
praveen-garg, line 64: instead of changing with every release, I would suggest unit tests to have [regex match check for semVer ](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string). We don't use prerelease/buildmetadata, hence we can skip those group matches even if optional.  A simple check that version exists is also okay.

python/fledge/services/core/api/common.py
MarkRiddoch: Given ping is used frequently by the GUI it is important we make it as low impact as possible. Let's read the version from the file once at startup, we canteen add the version in the startup log message.

tests/system/python/api/test_common.py
MarkRiddoch: Agreed, a simple [0-9]+\.[0-9+]\.[0-9]+ regex check would be good enough.

praveen-garg (approved):

python/fledge/services/core/api/common.py
praveen-garg: Yes, This is memoized now with https://github.com/fledge-iot/fledge/pull/243/commits/a77564be28c1a1f3219b3deade418596ee5a64b3

tests/system/python/api/test_common.py
praveen-garg: Unit tests are updated in https://github.com/fledge-iot/fledge/pull/243/commits/0733b9588ee62d3f626f1b1bc25f7281edab46c3, to have a check contains as those are ultimately mocked. For API integration test, we can add a patch PR for this regex check.

ashish-jabble: Added under https://github.com/fledge-iot/fledge/pull/244

PR #247, opened 2021-01-08
FOGL-4709 on Creating user either certificate or password is required

ashish-jabble:
**Note** - ~Downgrade testing is blocked due to FOGL-4871~

**items**
- [x] SQLite support
- [x] PostgreSQL support
- [x] Unit tests (on going task...)
- [x] allow user to create without password
- [ ] Required changes in login endpoint as well
- [x] Allow to add and update Real name (first name + last name)
- [x] Allow to update Real name (first name + last name)
- [x] Allow to add description for the user (can be empty)
- [x] Allow to update description for the user (can be empty)
- [x]  Allow  admin only to enable/disable the user via PUT to user_id or username with payload {'enable': true/false}
- [x] Ensure username cannot have spaces and not case sensitive. Only allow 0-9a-zA-Z, - & _ (may be . too)
- [x] Ensure consistency either use user_id or username in URL slugs
- [x] Verify health of API system tests
- [ ] Package system tests (which may take later upon)


MarkRiddoch (approved):

python/fledge/services/core/api/auth.py
```diff
@@ -253,42 +256,64 @@ async def create_user(request):
...
+        valid_extensions = ('.cert', '.cer', '.crt', '.pem')
+        if not certificate.endswith(valid_extensions):
+            msg = "Accepted cert extensions are {}".format(valid_extensions)
+            _logger.warning(msg)
+            raise web.HTTPBadRequest(reason=msg, body=json.dumps({"message": msg}))
+        is_cert_found = await User.Objects.find_cert(certificate)
+        if not is_cert_found:
+            msg = "{} is not found in certificate store".format(certificate)
+            _logger.warning(msg)
+            raise web.HTTPBadRequest(reason=msg, body=json.dumps({"message": msg}))
```
praveen-garg, line 291: Why a certificate required to reside in fledge. User should get the cert from issuing authority and issuing authority CA cert is the only thing we need in fledge cert store to verify the issue certificate.

praveen-garg (changes requested):
Another issue is role verification when user issued an admin role based certificate but downgraded to role `user`.

python/fledge/services/core/api/auth.py
ashish-jabble: As per specs I have added this check, @MarkRiddoch Could you please confirm?

python/fledge/services/core/user_model.py
```diff
@@ -172,13 +180,14 @@ async def update(cls, user_id, user_data):
                 result = await storage_client.update_tbl("users", payload)
                 if result['rows_affected']:
                     # FIXME: FOGL-1226 active session delete only in case of role_id and password updation
-
-                    # delete all active sessions
-                    await cls.delete_user_tokens(user_id)
+                    if 'password' in user_data or 'role_id' in user_data:
```
praveen-garg, line 183: also if access method is chnaged from any/cert/pwd to pwd/cert itself (not to any)

praveen-garg (approved):
Need to ensure, disabled user can not login, pwd can not be changed. 

+ system test https://github.com/fledge-iot/fledge/blob/develop/tests/system/python/packages/test_authentication.py

MarkRiddoch (dismissed):

MarkRiddoch (approved):

PR #248, opened 2021-01-11
FOGL-4843 Make sure the description is added to the configuration

MarkRiddoch:
category of a north service

### DIFF
--- a/C/services/north/north.cpp
+++ b/C/services/north/north.cpp
@@ -366,7 +366,6 @@ void NorthService::createConfigCategories(DefaultConfigCategory configCategory, 
 
 	// Deal with registering and fetching the configuration
 	DefaultConfigCategory defConfig(configCategory);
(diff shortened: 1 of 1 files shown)


C/services/north/north.cpp
```diff
@@ -376,6 +375,7 @@ void NorthService::createConfigCategories(DefaultConfigCategory configCategory,
                                                defaultServiceConfig);
 	defConfig += serviceConfig;
 
+	defConfig.setDescription(current_name);	// TODO We do not have access to the description
```
ashish-jabble, line 378: Do we have any JIRA for TODO comment "We do not have access to the description"? It would be nice to have so that later/future we will come to know to fix this comment

ashish-jabble (approved):

C/services/north/north.cpp
MarkRiddoch: No, that has been around for a very long time and is very low priority as there is no meaningful comment usually.

PR #254, opened 2021-01-22
FOGL-4937 - fix the hardcoded file name

sstefano69:
- fix the hardcoded fledge file name use by the storage


sstefano69:
invalid

PR #256, opened 2021-01-22
FOGL-4937: fix the hardcoded fledge file name use by the storage

sstefano69:


MarkRiddoch (commented):
This is presumably going to have upgrade issue?

sstefano69:
> This is presumably going to have upgrade issue?

I don't so, 
it was an hardcoded error since a while, 
visible if you start in stand alone without environments configured, 
but let talk to clarify.

sstefano69:
within  #259

PR #267, opened 2021-02-01
FOGL-4927  return 401: Unauthorized when auth creds required; not 403: forbidden which is meant for RBAC

praveen-garg:


praveen-garg:
/cc @mshariq-nerd  for fledge-gui that @monika-Dianomic is handling.

@saurabhagrdev This will need  http code check fix in FogMan too.

monika-Dianomic (approved):

MarkRiddoch (approved):

PR #274, opened 2021-02-10
file with common plugin boilerplate classes which aid in writing simp…

doug-dianomic:
…ler plugins.

Helper classes to aid writing simpler plugins:
. HandleMap -- allows passing a json dict handle back to clients, but managing (per-client unique) internal handles within plugin
. PluginHandle -- base class for internal handles which manages converting and updating config state, among other things so that we don't (necessarily) have to shutdown/init and create a new user handle when we get reconfigure requests.

Apart from my normal spacing problems, this shouldn't cause trouble for anyone. It's mostly pointed at foglamp plugins but we have no mechanism for saving foglamp common python utilities. (If it's generally useful for fledge, all the better...)


praveen-garg (approved):

python/fledge/common/plugin_helpers.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
+
+# FLEDGE_BEGIN
+# See: http://fledge.readthedocs.io/
```
praveen-garg, line 4: https://fledge-iot.readthedocs.io/

praveen-garg (approved):

PR #277, opened 2021-02-12
Increase size of asset_code in Postgres database

MarkRiddoch:

@sstefano69 I still need to do SQLite, and I have doubts about that. But in order to let Sebastian move forward can you see a problem with this?

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,2 +1,2 @@
 fledge_version=1.8.2
-fledge_schema=39
+fledge_schema=40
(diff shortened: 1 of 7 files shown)


sstefano69 (approved):
it worked fine, upgrade also

Detected 'postgres' Fledge DB schema change from version [35] to [40], applying Upgrade/Downgrade ...
Fledge DB schema has been upgraded to version [40]
Starting Fledge v1.8.2........
Fledge started.

PR #290, opened 2021-02-18
FOGL-5035: allow some time for filter pipeline setup

pintomax:
FOGL-5035: allow some time for filter pipeline setup


MarkRiddoch (approved):
I wonder if we should include this in 1.9?

PR #305, opened 2021-02-23
FOGL-5053: Added the schema while creating the packages table

YashTatkondawar:
Note: For sqlite init.sql added the schema details for the sake of consistency although it is not required.


praveen-garg (approved):
thanks! sqlite thing is good for better readability/consistency; but not required.

PR #312, opened 2021-02-26
update PluginHandle

doug-dianomic:
[Hey Praveen & Deepanshu, plesae take a look. My sigfns should be the only client of these changes...]

. work with filters or plugins that don't need the same callbacks
. set up a logger by default for servers -- move away from those global loggers
. add rpc_setup, rpc_config, which will create an rpc server that the plugin is a client for
. add shutdown which will proxy to rpc.shutdown if rpc exists (assume it may be called as super())
add PluginRPCServer to reduce boilerplate for iprpc servers
. add logger by default
. add config_update, which moves all config items into server's handle member functions

### DIFF
--- a/C/common/include/filter_pipeline.h
+++ b/C/common/include/filter_pipeline.h
@@ -54,7 +54,6 @@ public:
 
 private:
 	PLUGIN_HANDLE	loadFilterPlugin(const std::string& filterName);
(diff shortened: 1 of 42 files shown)


praveen-garg (approved):

praveen-garg:
lgtm!

PR #317, opened 2021-03-03
FOGL-5079 Fix system test

MarkRiddoch:


YashTatkondawar (approved):
For CoAP and http south like services, will the service port be as configured or something else?

MarkRiddoch:
> This specific thing is resolved but we are seeing issues on Rpi for south services which I have mentioned in FOGL-5079.
> Also for CoAP and http south like services, will the service port be as configured or something else?

Service port is a randomly allocated port. I see no errors on Raspberry Pi

PR #320, opened 2021-03-04
FOGL-5062: Deep copy in copy constructor

pintomax:
FOGL-5062: Deep copy in copy constructor

### DIFF
--- a/C/common/datapoint.cpp
+++ b/C/common/datapoint.cpp
@@ -89,29 +89,42 @@ std::string DatapointValue::toString() const
 }
 
 /**
- * Delete the DatapointValue alongwith possibly nested Datapoint objects
+ * Delete the DatapointValue along with possibly nested Datapoint objects
  */
 void DatapointValue::deleteNestedDPV()
 {
 	if (m_type == T_STRING)
 	{
-		delete m_value.str;
-		m_value.str = NULL;
+		if (m_value.str) {
+			delete m_value.str;
+			m_value.str = NULL;
+		}
 	}
 	else if (m_type == T_FLOAT_ARRAY)
 	{
-		delete m_value.a;
-		m_value.a = NULL;
-	}
-	else if (m_type == T_DP_DICT || m_type == T_DP_LIST)
-	{
-		for (auto it = m_value.dpa->begin(); // std::vector<Datapoint *>*	dpa;
-			 it != m_value.dpa->end();
-			 ++it)
-		{
-			delete (*it);
+		if (m_value.a) {
+			delete m_value.a;
+			m_value.a = NULL;
 		}
-		delete m_value.dpa;
+	}
+	else if (m_type == T_DP_DICT ||
+		 m_type == T_DP_LIST)
+	{
+		if (m_value.dpa) {
+			for (auto it = m_value.dpa->begin();
+				 it != m_value.dpa->end();
+				 ++it)
+			{
+				if (*it) {
+					// Call DatapointValue destructor
+					delete(*it);
+				}
+			}
+
+			// Remove vector pointer
+			delete m_value.dpa;
+			m_value.dpa = NULL;
+		}
 	}
 }
 
@@ -122,17 +135,40 @@ DatapointValue::~DatapointValue()
 {
 	if (m_type == T_STRING)
 	{
-		delete m_value.str;
-		m_value.str = NULL;
+		if (m_value.str) {
+			delete m_value.str;
+			m_value.str = NULL;
+		}
 	}
 	if (m_type == T_FLOAT_ARRAY)
 	{
-		delete m_value.a;
-		m_value.a = NULL;
+		if (m_value.a) {
+			delete m_value.a;
+			m_value.a = NULL;
+		}
 	}
-	// For nested DPV, d'tor is always called from holding Datapoint object's destructor
-}
+	// Check for T_DP_DICT or T_DP_LIST
+	if (m_type == T_DP_DICT ||
+	    m_type == T_DP_LIST)
+	{
+		if (m_value.dpa)
+		{
+			for (auto it = m_value.dpa->begin();
+				 it != m_value.dpa->end();
+				 ++it)
+			{
+				if (*it) {
+					// Call DatapointValue destructor
+					delete(*it);
+				}
+			}
 
+			// Remove vector pointer
+			delete m_value.dpa;
+			m_value.dpa = NULL;
+		}
+	}
+}
 
 /**
  * Copy constructor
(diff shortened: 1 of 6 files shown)


pintomax:
Unit tests will come with next commits

pintomax:
@AmandeepArora has been added as reviewer

C/common/datapoint.cpp
```diff
@@ -150,7 +150,32 @@ DatapointValue::DatapointValue(const DatapointValue& obj)
...
+				else if (d->getData().getType() == T_FLOAT_ARRAY)
+				{
+					std::vector<double> *currA = d->getData().getDpArr();
+					DatapointValue v(*currA);
+					m_value.dpa->push_back(new Datapoint(d->getName(), v));
+				}
+				else
+				{
+					DatapointValue v(d->getData());
+					m_value.dpa->push_back(new Datapoint(d->getName(), v));
```
AmandeepArora, line 175: Can you please clarify how this will take care of the nested case? e.g. list inside dict inside list. Each level would required new allocation on heap.

AmandeepArora (commented):
Added comment.

C/common/datapoint.cpp
```diff
@@ -150,7 +150,32 @@ DatapointValue::DatapointValue(const DatapointValue& obj)
...
 		case T_DP_DICT:
 		case T_DP_LIST:
-			m_value.dpa = obj.m_value.dpa; // TODO: need to fix this, need to do nested copying in newly allocated memory
+			m_value.dpa = new std::vector<Datapoint*>();
+			for (auto it = obj.m_value.dpa->begin();
+				it != obj.m_value.dpa->end();
+				++it)
+			{
+				Datapoint *d = *it;
+				if (d->getData().getType() == T_STRING)
```
MarkRiddoch, line 159: Why the different cases? You could simply invoke the copy constructor of each datapoint in the vector to create copies. That woudl also then recurse down the structure and be simpler I believe.

C/common/datapoint.cpp
```diff
@@ -89,29 +89,42 @@ std::string DatapointValue::toString() const
...
- * Delete the DatapointValue alongwith possibly nested Datapoint objects
+ * Delete the DatapointValue along with possibly nested Datapoint objects
  */
 void DatapointValue::deleteNestedDPV()
 {
 	if (m_type == T_STRING)
 	{
-		delete m_value.str;
-		m_value.str = NULL;
+		if (m_value.str) {
```
AmandeepArora, line 98: delete is safe on NULL pointers. "The C++ language guarantees that delete p will do nothing if p is equal to NULL."

C/common/datapoint.cpp
```diff
@@ -122,17 +135,40 @@ DatapointValue::~DatapointValue()
...
-		delete m_value.a;
-		m_value.a = NULL;
+		if (m_value.a) {
+			delete m_value.a;
+			m_value.a = NULL;
+		}
 	}
-	// For nested DPV, d'tor is always called from holding Datapoint object's destructor
-}
+	// Check for T_DP_DICT or T_DP_LIST
```
AmandeepArora, line 150: Is this destructor and DatapointValue::deleteNestedDPV() exactly same now?

C/common/datapoint.cpp
pintomax: Added simplification where possible

C/common/datapoint.cpp
pintomax: Yes, they look similar right now

C/common/datapoint.cpp
```diff
@@ -103,15 +103,24 @@ void DatapointValue::deleteNestedDPV()
...
-			 it != m_value.dpa->end();
-			 ++it)
-		{
-			delete (*it);
+		if (m_value.dpa) {
+			for (auto it = m_value.dpa->begin();
+				 it != m_value.dpa->end();
+				 ++it)
+			{
+				if (*it) {
```
AmandeepArora, line 114: There is no need to check for NULL before delete.

AmandeepArora (commented):
May be just as a verification check, print address of all DP and DPV objects when you call the copy constructor to make sure that the new object is not sharing any sub-objects with original object. Try this for a complex nesting like dict inside list inside dict.

pintomax:
> May be just as a verification check, print address of all DP and DPV objects when you call the copy constructor to make sure that the new object is not sharing any sub-objects with original object. Try this for a complex nesting like dict inside list inside dict.

@AmandeepArora There will be a new unit test that will deal with that

AmandeepArora (approved):
Looks good.

C/common/datapoint.cpp
MarkRiddoch: I was kind of expecting you to drop all the if, else stuff and just have something that created a copy of the datapoint itself. 

```
m_value.dpa = new std::vector<Datapoint*>();
for (auto it = obj.m_value.dpa->begin();
                it != obj.m_value.dpa->end(); ++it)
{       
        Datapoint *dp = *it;
        m_value.dpa->push_back(new Datapoint(*dp));
}
```

You need a copy constructor for the datapoint anyway, the default one will cause problems.  Then you don't care what this datapoint contains. It is also more future proof.

MarkRiddoch (approved):
Looks good to me. Just one thing to check, does the Datapoint copy constructor behave correctly?

pintomax:
> Looks good to me. Just one thing to check, does the Datapoint copy constructor behave correctly?

Datapoint constructor is not implemented, so default one is being used with shallow copy.

Unit tests and south server ingest with this structure is ok, Valgrind tests are ok as well:
```
{
 "reading": {
  "features": {
   "Value": {
    "M1": -89,
    "M2": -8,
    "M3": 0
   },
   "QualityDescriptor": 101,
   "TimeStamp": "2021-03-05",
   "IsTimestampValid": 0
  }
 }
}
```

A new unit tests which will track all involved object memory addresses is being implemented.

pintomax:
> Looks good to me. Just one thing to check, does the Datapoint copy constructor behave correctly?

Tested this object:

```
{
  "asset_code": "test55",
  "user_ts": "2021-03-08 15:16:24.520443+00:00",
  "ts": "2021-03-08 15:16:24.520443+00:00",
  "reading": {
    "aaa": {
      "final": {
        "unnamed_list_elem#1": {
          "first": 1,
          "third": {
            "M1": "x"
          }
        },
        "unnamed_list_elem#2": {
          "second": 1.1
        }
      }
    }
  }
}
```

MarkRiddoch (approved):

pintomax:
> Looks good to me. Just one thing to check, does the Datapoint copy constructor behave correctly?

Yes

PR #322, opened 2021-03-09
FOGL-3125, FOGL-5121: Update  audit entries tests

pintomax:
FOGL-3125: fix for adding audit entries in Postgres

### DIFF
--- a/python/fledge/services/core/api/audit.py
+++ b/python/fledge/services/core/api/audit.py
@@ -114,14 +114,15 @@ async def create_audit_entry(request):
                    'severity': severity,
                    'details': details
                    }
-        return web.json_response(message)
     except AttributeError as e:
         # Return error for wrong severity method
         err_msg = "severity type {} is not supported".format(severity)
         _logger.error("Error in create_audit_entry(): %s | %s", err_msg, str(e))
         raise web.HTTPNotFound(reason=err_msg)
-    except Exception as ex:
+    except (StorageServerError, Exception) as ex:
         raise web.HTTPInternalServerError(reason=str(ex))
+    else:
+        return web.json_response(message)
 
 
 async def get_audit_entries(request):
--- a/python/fledge/services/core/api/audit.py
+++ b/python/fledge/services/core/api/audit.py
@@ -124,12 +124,12 @@ async def create_audit_entry(request):
     except StorageServerError as ex:
         if int(ex.code) in range(400, 500):
             err_msg = 'Audit entry cannot be logged. {}'.format(ex.error['message'])
-            raise web.HTTPBadRequest(body=json.dumps({"error": err_msg}))
+            raise web.HTTPBadRequest(body=json.dumps({"message": err_msg}))
         else:
             err_msg = 'Failed to log audit entry. {}'.format(ex.error['message'])
-            raise web.HTTPInternalServerError(body=json.dumps({"error": err_msg}))
+            raise web.HTTPInternalServerError(body=json.dumps({"message": err_msg}))
     except Exception as ex:
-        raise web.HTTPInternalServerError(reason=str(ex))
+        raise web.HTTPInternalServerError(reason=str(ex), body=json.dumps({"message": str(ex)}))
     else:
         return web.json_response(message)
 
--- a/python/fledge/services/core/api/audit.py
+++ b/python/fledge/services/core/api/audit.py
@@ -119,8 +119,15 @@ async def create_audit_entry(request):
         # Return error for wrong severity method
         err_msg = "severity type {} is not supported".format(severity)
         _logger.error("Error in create_audit_entry(): %s | %s", err_msg, str(e))
-        raise web.HTTPNotFound(reason=err_msg)
-    except (StorageServerError, Exception) as ex:
+        raise web.HTTPNotFound(reason=err_msg, body=json.dumps({"error": err_msg})
+    except StorageServerError as ex:
+        if int(ex.code) in range(400, 500):
+            err_msg = 'Audit entry cannot not be logged. {}'.format(json.loads(ex.error)['message'])
+            raise web.HTTPBadRequest(body=json.dumps({"error": err_msg}))
+        else:
+            err_msg = 'Failed to log audit entry. {}'.format(json.loads(ex.error)['message'])
+            raise web.HTTPInternalServerError({"error": err_msg}
+    except Exception as ex:
         raise web.HTTPInternalServerError(reason=str(ex))
     else:
         return web.json_response(message)
--- a/python/fledge/services/core/api/audit.py
+++ b/python/fledge/services/core/api/audit.py
@@ -10,6 +10,7 @@ from enum import IntEnum
 from aiohttp import web
 
 from fledge.common.storage_client.payload_builder import PayloadBuilder
+from fledge.common.storage_client.exceptions import StorageServerError
 from fledge.services.core import connect
 from fledge.common.audit_logger import AuditLogger
 from fledge.common import logger
--- a/python/fledge/services/core/api/audit.py
+++ b/python/fledge/services/core/api/audit.py
@@ -119,14 +119,14 @@ async def create_audit_entry(request):
         # Return error for wrong severity method
         err_msg = "severity type {} is not supported".format(severity)
         _logger.error("Error in create_audit_entry(): %s | %s", err_msg, str(e))
-        raise web.HTTPNotFound(reason=err_msg, body=json.dumps({"error": err_msg})
+        raise web.HTTPNotFound(reason=err_msg, body=json.dumps({"error": err_msg}))
     except StorageServerError as ex:
         if int(ex.code) in range(400, 500):
             err_msg = 'Audit entry cannot not be logged. {}'.format(json.loads(ex.error)['message'])
             raise web.HTTPBadRequest(body=json.dumps({"error": err_msg}))
         else:
             err_msg = 'Failed to log audit entry. {}'.format(json.loads(ex.error)['message'])
-            raise web.HTTPInternalServerError({"error": err_msg}
+            raise web.HTTPInternalServerError({"error": err_msg})
     except Exception as ex:
         raise web.HTTPInternalServerError(reason=str(ex))
     else:
--- a/python/fledge/services/core/api/audit.py
+++ b/python/fledge/services/core/api/audit.py
@@ -8,6 +8,7 @@ import copy
 from datetime import datetime
 from enum import IntEnum
 from aiohttp import web
+import json
 
 from fledge.common.storage_client.payload_builder import PayloadBuilder
 from fledge.common.storage_client.exceptions import StorageServerError
--- a/python/fledge/services/core/api/audit.py
+++ b/python/fledge/services/core/api/audit.py
@@ -120,7 +120,7 @@ async def create_audit_entry(request):
         # Return error for wrong severity method
         err_msg = "severity type {} is not supported".format(severity)
         _logger.error("Error in create_audit_entry(): %s | %s", err_msg, str(e))
-        raise web.HTTPNotFound(reason=err_msg, body=json.dumps({"error": err_msg}))
+        raise web.HTTPNotFound(reason=err_msg, body=json.dumps({"message": err_msg}))
     except StorageServerError as ex:
         if int(ex.code) in range(400, 500):
             err_msg = 'Audit entry cannot be logged. {}'.format(ex.error['message'])
--- a/python/fledge/services/core/api/audit.py
+++ b/python/fledge/services/core/api/audit.py
@@ -123,7 +123,7 @@ async def create_audit_entry(request):
         raise web.HTTPNotFound(reason=err_msg, body=json.dumps({"error": err_msg}))
     except StorageServerError as ex:
         if int(ex.code) in range(400, 500):
-            err_msg = 'Audit entry cannot not be logged. {}'.format(ex.error['message'])
+            err_msg = 'Audit entry cannot be logged. {}'.format(ex.error['message'])
             raise web.HTTPBadRequest(body=json.dumps({"error": err_msg}))
         else:
             err_msg = 'Failed to log audit entry. {}'.format(ex.error['message'])
--- a/python/fledge/services/core/api/audit.py
+++ b/python/fledge/services/core/api/audit.py
@@ -123,11 +123,11 @@ async def create_audit_entry(request):
         raise web.HTTPNotFound(reason=err_msg, body=json.dumps({"error": err_msg}))
     except StorageServerError as ex:
         if int(ex.code) in range(400, 500):
-            err_msg = 'Audit entry cannot not be logged. {}'.format(json.loads(ex.error)['message'])
+            err_msg = 'Audit entry cannot not be logged. {}'.format(ex.error['message'])
             raise web.HTTPBadRequest(body=json.dumps({"error": err_msg}))
         else:
-            err_msg = 'Failed to log audit entry. {}'.format(json.loads(ex.error)['message'])
-            raise web.HTTPInternalServerError({"error": err_msg})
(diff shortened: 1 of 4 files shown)


pintomax:
I see tests for both engines are failing:
Postgres:

FAILED tests/system/python/TestAudit test_default_get_audit - AssertionError: assert 'Installation' == 'service'
FAILED tests/system/python/TestAudit test_default_get_audit - AssertionError: assert 'SCHEDULER' == 'Advanced'
FAILED tests/system/python/TestAudit test_default_get_audit - AssertionError: assert 'SMNTR' == 'SCHEDULER'

Sqlite
FAILED tests/system/python/TestAudit test_default_get_audit - AssertionError: assert '' == 'sqlite'
FAILED tests/system/python/TestAudit test_default_get_audit - AssertionError: assert '' == 'sqlite'
FAILED tests/system/python/TestAudit test_default_get_audit - AssertionError: assert 'Utilities' == 'sqlite'
FAILED tests/system/python/TestAudit test_default_get_audit - AssertionError: assert 'Utilities' == 'sqlite'
FAILED tests/system/python/TestAudit test_default_get_audit - AssertionError: assert 'General' == 'Utilities'
FAILED tests/system/python/TestAudit test_default_get_audit - AssertionError: assert '' == 'sqlite'


Sometimes 3 errors, sometimes 5 etc

It seems it's related to db content

tests/system/python/api/test_audit.py
```diff
@@ -79,20 +79,40 @@ def test_default_get_audit(self, fledge_url, wait_time, request_params, total_co
         assert 200 == r.status
         r = r.read().decode()
         jdoc = json.loads(r)
+
+        elems = jdoc['audit']
+        if storage_plugin == 'postgres':
+            for i, o in enumerate(elems):
+                if o['details'] and o['details']['name'] and o['details']['name'] == 'sqlite':
+                    #print("Skipping item " + str(i) + " = " + str(o))
+                    elems.remove(o)
```
praveen-garg, line 88: why shall we see sqlite entry when on postgres?!

tests/system/python/api/test_audit.py
```diff
@@ -79,20 +79,40 @@ def test_default_get_audit(self, fledge_url, wait_time, request_params, total_co
...
+
+            if audit_count > 1:
+                audit_count = audit_count -1
+
         assert len(jdoc), "No data found"
         assert total_count == jdoc['totalCount']
-        assert audit_count == len(jdoc['audit'])
+        assert audit_count == len(elems)
+
+        foundCategory = True
```
praveen-garg, line 104: what is this for?

tests/system/python/api/test_audit.py
```diff
@@ -110,5 +130,11 @@ def test_create_audit_entry(self, fledge_url, payload, total_count):
         assert 200 == r.status
         r = r.read().decode()
         jdoc = json.loads(r)
+
+        # sqlite has an additional entry: CONAD, INFORMATION, '{{"name":"sqlite", ...}'
+        # so total_count is being decresed in case of 'postgres' storage engine
+        if storage_plugin == 'postgres':
+           total_count = total_count -1
```
praveen-garg, line 137: this is okay!

tests/system/python/api/test_audit.py
```diff
@@ -79,20 +79,40 @@ def test_default_get_audit(self, fledge_url, wait_time, request_params, total_co
...
+
+        if storage_plugin == 'postgres':
+            # sqlite has an additional entry: CONAD, INFORMATION, '{{"name":"sqlite", ...}'
+            # so total_count is being decresed in case of 'postgres' storage engine
+            # same for audit_count
+            if total_count > 1:
+                total_count = total_count -1
+
+            if audit_count > 1:
+                audit_count = audit_count -1
```
praveen-garg, line 98: I believe this logic will fail given the query params in API. better option is to skip this and have separate test /test data for test_default_get_audit_postgres  etc. based on storage_plugin fixture

tests/system/python/api/test_audit.py
praveen-garg: we can put `test_create_audit_entry` fix in specific pull request and can merge this only fix (and a separate pull request for api.test_audit.TestAudit.test_default_get_audit)

tests/system/python/api/test_audit.py
```diff
@@ -79,20 +79,40 @@ def test_default_get_audit(self, fledge_url, wait_time, request_params, total_co
...
+
+        foundCategory = True
         if audit_count:
             if jdoc['audit'][0]['details']:
                 assert cat_name == jdoc['audit'][0]['details']['name']
 
     @pytest.mark.parametrize("payload, total_count", [
-        ({"source": "LMTR", "severity": "warning", "details": {"message": "Engine oil pressure low"}}, 12),
-        ({"source": "LMTR", "severity": "success", "details": {}}, 13),
+        ({"source": "SRVFL", "severity": "warning", "details": {"message": "Engine oil pressure low"}}, 12),
```
praveen-garg, line 110: does not matter, but message should justify the SRVFL code :]

tests/system/python/api/test_audit.py
praveen-garg: @pintomax limit and skip logic will fail. Even this will start failing for sqlite etc too. see my previous comment.

tests/system/python/api/test_audit.py
```diff
@@ -79,20 +79,34 @@ def test_default_get_audit(self, fledge_url, wait_time, request_params, total_co
         assert 200 == r.status
         r = r.read().decode()
         jdoc = json.loads(r)
+
+        elems = jdoc['audit']
+
+        if storage_plugin == 'postgres':
+            # sqlite has an additional entry: CONAD, INFORMATION, '{{"name":"sqlite", ...}'
```
MarkRiddoch, line 86: Why do the different plugins have different audit entries?

praveen-garg: @MarkRiddoch Storage category when with SQLite plugin, creates a child category `sqlite'; hence an additional CONAD entry. I'm not sure if when storage plugin is postgres there should be a child category for the same?

praveen-garg:
We should be good to go now. test_create_nonexistent_log_code_audit_entry will fail on sqlite until #339 is merged. 

SQLite:

```{"error": "Audit entry cannot be logged. FOREIGN KEY constraint failed"}```

PostgresSQL:

```{"error": "Audit entry cannot be logged. ERROR:  insert or update on table \"log\" violates foreign key constraint \"log_fk1\"DETAIL:  Key (code)=(LMR  ) is not present in table \"log_codes\"."}```

File: python/fledge/services/core/api/audit.py [hunk 7]
```diff
@@ -114,14 +116,22 @@ async def create_audit_entry(request):
                    'severity': severity,
                    'details': details
                    }
-        return web.json_response(message)
     except AttributeError as e:
         # Return error for wrong severity method
         err_msg = "severity type {} is not supported".format(severity)
         _logger.error("Error in create_audit_entry(): %s | %s", err_msg, str(e))
-        raise web.HTTPNotFound(reason=err_msg)
+        raise web.HTTPNotFound(reason=err_msg, body=json.dumps({"error": err_msg}))
```
ashish-jabble, line 123: we have used "message" KV pair in our API for such scenario. So this needs to be changed.

python/fledge/services/core/api/audit.py
```diff
@@ -114,14 +116,22 @@ async def create_audit_entry(request):
...
+    except StorageServerError as ex:
+        if int(ex.code) in range(400, 500):
+            err_msg = 'Audit entry cannot be logged. {}'.format(ex.error['message'])
+            raise web.HTTPBadRequest(body=json.dumps({"error": err_msg}))
```
ashish-jabble, line 127: Add 'reason' consistency wise across API & for the key KV change (same as above)

python/fledge/services/core/api/audit.py
```diff
@@ -114,14 +116,22 @@ async def create_audit_entry(request):
...
+        else:
+            err_msg = 'Failed to log audit entry. {}'.format(ex.error['message'])
+            raise web.HTTPInternalServerError(body=json.dumps({"error": err_msg}))
```
ashish-jabble, line 130: Add 'reason' consistency wise across API & for the key KV change (same as above)

tests/system/python/api/test_audit.py
```diff
@@ -53,24 +53,63 @@ def test_get_severity(self, fledge_url):
...
+        ('?source=START&severity=information&limit=1', 1, 1),
+        ('?source=START&severity=information&limit=1&skip=1', 1, 0)
     ])
-    def test_default_get_audit(self, fledge_url, wait_time, request_params, total_count, audit_count, cat_name):
+    def test_default_get_audit_sqlite(self, fledge_url, wait_time, request_params, total_count, audit_count, storage_plugin):
+        if storage_plugin == 'postgres':
+            pytest.skip('Test is for SQLite only. Storage category when with SQLite plugin, creates a child category `sqlite` hence one additional CONAD entry.')
+
+        if request_params == '':
+            time.sleep(wait_time)
```
ashish-jabble, line 78: Could you please add why extra wait time is required?

tests/system/python/api/test_audit.py
```diff
@@ -110,5 +148,26 @@ def test_create_audit_entry(self, fledge_url, payload, total_count):
...
+
+    @pytest.mark.parametrize("payload", [
+        ({"source": "LOGGN_X", "severity": "warning", "details": {"message": "Engine oil pressure low"}}),
+        ({"source": "LOG_X", "severity": "information", "details": {"message": "Engine oil pressure is okay."}})
+    ])
+    def test_create_nonexistent_log_code_audit_entry(self, fledge_url, payload, storage_plugin):
+        conn = http.client.HTTPConnection(fledge_url)
+        conn.request('POST', '/fledge/audit', body=json.dumps(payload))
+        r = conn.getresponse()
+        assert 400 == r.status
```
ashish-jabble, line 169: Add 1 more assertion with `r.reason`

tests/system/python/api/test_audit.py
```diff
@@ -110,5 +148,26 @@ def test_create_audit_entry(self, fledge_url, payload, total_count):
...
+        r = r.read().decode()
+        jdoc = json.loads(r)
+        print(jdoc)
```
ashish-jabble, line 172: remove print statement

tests/system/python/api/test_audit.py
```diff
@@ -110,5 +148,26 @@ def test_create_audit_entry(self, fledge_url, payload, total_count):
...
+           total_count = total_count -1
+
         assert len(jdoc), "No data found"
         assert total_count == jdoc['totalCount']
+
+    @pytest.mark.parametrize("payload", [
+        ({"source": "LOGGN_X", "severity": "warning", "details": {"message": "Engine oil pressure low"}}),
+        ({"source": "LOG_X", "severity": "information", "details": {"message": "Engine oil pressure is okay."}})
+    ])
+    def test_create_nonexistent_log_code_audit_entry(self, fledge_url, payload, storage_plugin):
```
ashish-jabble, line 165: I see storage_plugin parser option is not used for the test. Does it require?

tests/system/python/api/test_audit.py
```diff
@@ -110,5 +148,26 @@ def test_create_audit_entry(self, fledge_url, payload, total_count):
         assert 200 == r.status
         r = r.read().decode()
         jdoc = json.loads(r)
+
+        # Storage category when with SQLite plugin, creates a child category `sqlite` hence one additional CONAD entry.
+        # CONAD, INFORMATION, '{{"name":"sqlite", ...}'
+        # so total_count is being decresed in case of 'postgres' storage engine
```
ashish-jabble, line 154: Interesting one! It's by design?

ashish-jabble (changes requested):
Left some minor comments. Otherwise looks good!

tests/system/python/api/test_audit.py
praveen-garg: It should not be required, there is nothing to wait after start for defaults check. It existed in develop so considered trivial to check and kept. I think should I go ahead and remove this?

python/fledge/services/core/api/audit.py
praveen-garg: default reason text will be there? or we should add explicitly? As we are adding response body so I though we can skip custom reason text.

tests/system/python/api/test_audit.py
praveen-garg: I was thinking, we can keep such; for diagnosis using -vv such help. May be we can --verbose fixture in conftest and for system tests we can keep print for that if block?

tests/system/python/api/test_audit.py
praveen-garg: No, I had a skip with storage_plugin sqlite (TODO: merge foreign key support first) then removed that to let it fail but forgot to remove. I will remove it.

tests/system/python/api/test_audit.py
praveen-garg: Let @MarkRiddoch confirm on this. I think yes. When storage is configured with Postgres there are no children (child category) for Storage but SQLite only.

tests/system/python/api/test_audit.py
praveen-garg: Done

tests/system/python/api/test_audit.py
praveen-garg: Removed, but please confirm (I'm not sure what was the reason why a specific request_params == '' conditional check was added).

python/fledge/services/core/api/audit.py
praveen-garg: OK

tests/system/python/api/test_audit.py
```diff
@@ -110,5 +143,27 @@ def test_create_audit_entry(self, fledge_url, payload, total_count):
...
+    def test_create_nonexistent_log_code_audit_entry(self, fledge_url, payload):
+        conn = http.client.HTTPConnection(fledge_url)
+        conn.request('POST', '/fledge/audit', body=json.dumps(payload))
+        r = conn.getresponse()
+        assert 400 == r.status
+        assert 'Bad Request' in r.reason
+        r = r.read().decode()
+        jdoc = json.loads(r)
+        print(jdoc)
+        assert "Audit entry cannot be logged" in jdoc['error']
```
ashish-jabble, line 169: As you have changed the key then use jdoc['message']

tests/system/python/api/test_audit.py
MarkRiddoch: Yes, only SQLite has a child category

MarkRiddoch (approved):

tests/system/python/api/test_audit.py
ashish-jabble: Without wait time tests are failing. Hence this needs to be reverted.
Same apply to **test_default_get_audit_postgres** def

> why a specific request_params == '' conditional check was added).

Because we need wait time only once after fledge start; Not on all parameterized test data.

praveen-garg: What a pain, We should have wait in fledge_start fixture itself, and poll ping, if failed for x max tries * w wait time assert Failed to start/reset & start/restart

praveen-garg: reverted.

ashish-jabble (approved):
LGTM! Please squash & merge!

python/fledge/services/core/api/audit.py
ashish-jabble: reason text is only appears if we return explicitly (i.e reason="") - We had this style previously and GUI client mostly rely on this text if HTTP status code range lies between (3XX to 5XX).

We already had a JIRA for API improvements to return JSON response in body. https://scaledb.atlassian.net/browse/FOGL-4547

PR #325, opened 2021-03-11
FOGL-5052: create database if missing while restarting the core service

pintomax:
FOGL-5052: create database if missing while restarting the core service


pintomax:
Need to add a new action in storage plugin script in order to removed duplicated log messages

MarkRiddoch (approved):

PR #351, opened 2021-04-01
Added wait time to avoid asset tracker failure in e2e/test_e2e_pi_sca…

YashTatkondawar:
…leset.py

### DIFF
--- a/tests/system/python/e2e/test_e2e_pi_scaleset.py
+++ b/tests/system/python/e2e/test_e2e_pi_scaleset.py
@@ -125,6 +125,8 @@ class TestE2ePiEgressWithScalesetFilter:
         assert "http-e1" == tracked_item["asset"]
         assert "http_south" == tracked_item["plugin"]
 
+        # Wait until asset tracker entry is created especially on Rpi board
+        time.sleep(wait_time * 3)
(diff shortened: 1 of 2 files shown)


praveen-garg (dismissed):

tests/system/python/e2e/test_e2e_pi_scaleset.py
```diff
@@ -125,6 +125,8 @@ def test_end_to_end(self, start_south_north_with_filter, read_data_from_pi, fled
         assert "http-e1" == tracked_item["asset"]
         assert "http_south" == tracked_item["plugin"]
 
+        # Wait until asset tracker entry is created especially on Rpi board
+        time.sleep(wait_time * 3)
```
praveen-garg, line 129: _verify_egress is done already. so it should be only before egress check 

and move

if not skip_verify_north_interface:
            self._verify_egress(read_data_from_pi, pi_host, pi_admin, pi_passwd, pi_db, wait_time, retries)

after utils.get_asset_tracking_details(fledge_url, "Ingest")

praveen-garg: add comment wait for egress scheduled task to run and let filter make asset tracker entry

praveen-garg (approved):

PR #363, opened 2021-04-15
Fogl 4871: downgrade fix 

sstefano69:


ashish-jabble (approved):
Potential fix works well!

PR #378, opened 2021-05-03
FOGL 5336 bearer token verification added in auth token API

ashish-jabble:

```
 $ curl -H "bearer_token: S1_29615de1cd2e4fe3ab2eaba10672b10" -sX GET http://localhost:39087/fledge/service/authtoken | jq
{
  "message": "Bearer token is invalid"
}

$ curl -H "bearer_token: S1_29615de1cd2e4fe3ab2eaba10672b102" -sX GET http://localhost:39087/fledge/service/authtoken
{"token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1aWQiOjEsImV4cCI6MTYyMDA0NzM5M30.ehyOPljhJbxfvm66TxTCz4t9Pmp6YxcjCVH5ZFWw0gU"}

```


python/fledge/services/core/server.py
```diff
@@ -1233,11 +1244,23 @@ async def get_service(cls, request):
...
 
             :Example:
                 curl -sX GET http://localhost:<core mgt port>/fledge/service/authtoken
         """
+        bearer_token = request.headers.get('bearer_token', "")
+        if bearer_token:
+            services_list = ServiceRegistry.all()
+            is_bearer_token_valid = False
+            for svc in services_list:
+                if svc._token == bearer_token:
```
ashish-jabble, line 1257: IMO, we should pass service name in HEADER as we don't have any service info in API route. So that we make sure that token is for that service only. Otherwise for the current scenario if passing bearer_token exists in service registry then TRUE irrespective of the service it is requesting with.

praveen-garg: request.headers.get('authorization', None) 
the bearer token better be a jwt token with refresh token and expiry; that bearer token will itself be used to identify service. if a service requires token it should return 401 if that does not match or empty.

In other words, 
service registration with name + single use token is similar to login (if arg parser has token=true in process script)
and then for next request bring that (similar to user' auth token) bearer token to identify the requester and allow to access resources.

python/fledge/common/service_record.py
```diff
@@ -45,9 +45,9 @@ class InvalidServiceStatus(Exception):
         # TODO: tell allowed service status?
         pass
 
-    __slots__ = ['_id', '_name', '_type', '_protocol', '_address', '_port', '_management_port', '_status', '_token']
+    __slots__ = ['_id', '_name', '_type', '_protocol', '_address', '_port', '_management_port', '_status', '_bearer_token']
```
praveen-garg, line 48: I would suggest not to put _bearer_token in slots. keep a separate data structure (say svcRegistryTokens) to hold serviceName/ID: issuedBearerToken (private to service registry). if single use token match then add entry to svcRegistryTokens and use that to verify at each request for the services requires auth. (via a middleware) pass/no_pass

praveen-garg: this shall also mitigate the risk/chances of exposure/snooping of bearer tokens in service API and to other service.

python/fledge/services/core/server.py
MarkRiddoch: Sure, if we have an expiry we will need to make sure there is a mechanism to extend it or to issue new tokens as these services could run for a very long time.

Passing the service name in the header is not a bad idea, even if it is also encrypted int e token, that way we can do an extra level of checking. We can hide all of this inside the management client code.

praveen-garg: Yeah.
re: no explicit header required
JWT aud(ience) claim will take care of that. i.e. exactly, encrypted in the token

MarkRiddoch:
FOGL-5336 appears to be closed yet we have this on-hold pull request which @pintomax says he is waiting for. Is this really part of 5336 or part of another change? The Jira that Misssimiliano has is FOGL5144, should this be part of that Jira?

We need to up the priority if this fix to get the functionality created.

PR #418, opened 2021-06-08
FOGL-5530 plugin_start is optional for south plugins that do not pers…

MarkRiddoch:
…ist data

### DIFF
--- a/python/fledge/plugins/common/shim/south_shim.py
+++ b/python/fledge/plugins/common/shim/south_shim.py
@@ -68,7 +68,8 @@ def plugin_shutdown(handle):
 
 def plugin_start(handle):
     _LOGGER.info("plugin_start")
-    return _plugin.plugin_start(handle)
+    if hasattr(_plugin, "plugin_start"):
+        return _plugin.plugin_start(handle)
 
 
 def plugin_register_ingest(handle, callback, ingest_ref):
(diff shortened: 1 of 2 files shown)


pintomax (approved):

ashish-jabble (approved):
Looks good! Also ran custom CI job for regression check `fledge_nightly_build_ub1804/720` - all good

python/fledge/plugins/common/shim/south_shim.py
```diff
@@ -68,7 +68,8 @@ def plugin_shutdown(handle):
 
 def plugin_start(handle):
     _LOGGER.info("plugin_start")
-    return _plugin.plugin_start(handle)
+    if hasattr(_plugin, "plugin_start"):
```
deepanshu-yadav, line 71: Hi @MarkRiddoch ,

What if a plugin has both plugin_poll and plugin_start API  methods implemented ? 

or 

what if plugin_start does not exists/misspelled  in async mode plugin;

or 

what if mistakenly wrote start handle/hook in poll mode?

I think the better way is to
1.  See the message from plugin_info.
2.  Decide whether that it is async or poll from plugin info itself.
3. Expect plugin_start if async is present , expect plugin_poll if poll is present.

cc 
@ashish-jabble 
@praveen-garg

PR #422, opened 2021-06-21
FOGL-5474: System test for OMF as north service

YashTatkondawar:

Tests with different scenarios related to north service are complete.

Pending Items related to filters with North service:
- [x] Add filter to the North service.
- [x] Disable/Enable the filter.
- [x] Reconfiguration of the attached filter North service.
- [x] Delete/re-add the North service (with filter now). - Skipped because of FOGL-5215
- [x] Change the order of the filters.
- [x] Delete and re-add the filter.

### DIFF
--- a/tests/system/python/conftest.py
+++ b/tests/system/python/conftest.py
@@ -213,8 +213,8 @@ def start_north_pi_v2_web_api():
 
 
 @pytest.fixture
-def start_north_pi_v2_web_api_service():
-    def _start_north_pi_server_c_web_api_service(fledge_url, pi_host, pi_port, pi_db="Dianomic", auth_method='basic',
+def start_north_omf_as_a_service():    
+    def _start_north_omf_as_a_service(fledge_url, pi_host, pi_port, pi_db="Dianomic", auth_method='basic',
                                          pi_user=None, pi_pwd=None, north_plugin="OMF",
                                          service_name="NorthReadingsToPI_WebAPI", start=True):
         """Start north service"""
@@ -241,12 +241,11 @@ def start_north_pi_v2_web_api_service():
         assert 200 == r.status
         retval = json.loads(r.read().decode())
         return retval
-    return _start_north_pi_server_c_web_api_service
+    return _start_north_omf_as_a_service
 
 
 start_north_pi_server_c = start_north_pi_v2
 start_north_pi_server_c_web_api = start_north_pi_v2_web_api
-start_north_pi_server_c_web_api_service = start_north_pi_v2_web_api_service
 
 @pytest.fixture
 def read_data_from_pi():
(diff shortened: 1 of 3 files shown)


tests/system/python/conftest.py
```diff
@@ -30,18 +31,21 @@
 
 @pytest.fixture
 def clean_setup_fledge_packages(package_build_version):
-    assert os.environ.get('FLEDGE_ROOT') is not None
```
praveen-garg, line 33: why this has been removed?

tests/system/python/conftest.py
```diff
@@ -30,18 +31,21 @@
...
+    # This  gives the path of directory where fledge is cloned. conftest_file < python < system < tests < ROOT
+    PROJECT_ROOT = Path(__file__).parent.parent.parent.parent
+    SCRIPTS_DIR_ROOT = "{}/tests/system/python/scripts/package/".format(PROJECT_ROOT)
```
praveen-garg, line 36: is this not for non-packages based test? here we have `clean_setup_fledge_packages`

tests/system/python/conftest.py
```diff
@@ -208,8 +212,41 @@ def _start_north_pi_server_c_web_api(fledge_url, pi_host, pi_port, pi_db="Dianom
...
+        r = conn.getresponse()
+        assert 200 == r.status
+        retval = json.loads(r.read().decode())
+        return retval
+    return _start_north_pi_server_c_web_api_service
+
+
 start_north_pi_server_c = start_north_pi_v2
 start_north_pi_server_c_web_api = start_north_pi_v2_web_api
+start_north_pi_server_c_web_api_service = start_north_pi_v2_web_api_service
```
praveen-garg, line 249: start_north_omf_service (no alias needed)

tests/system/python/conftest.py
YashTatkondawar: This is removed because this def expects `FLEDGE_ROOT` to be the path of the cloned fledge directory because it tries to run the script from that path. But in the case of packages `FLEDGE_ROOT` should be `/usr/local/fledge`. So added `PROJECT_ROOT` to run the scripts with reference to the `PROJECT_ROOT` path.

tests/system/python/conftest.py
YashTatkondawar: No, this is for packages based tests.

tests/system/python/conftest.py
```diff
@@ -208,8 +212,41 @@ def _start_north_pi_server_c_web_api(fledge_url, pi_host, pi_port, pi_db="Dianom
     return _start_north_pi_server_c_web_api
 
 
+@pytest.fixture
+def start_north_pi_v2_web_api_service():
```
praveen-garg, line 216: start_north_omf_as_a_service

praveen-garg (approved):

PR #453, opened 2021-08-12
FOGL-5827 upgrade and downgrade SQL fixes for purge_system schedule and process…

praveen-garg:
… script


YashTatkondawar (approved):

YashTatkondawar:
Upgrade and downgrade are good with these changes!

PR #457, opened 2021-08-20
FOGL-5809 : Exclude asset from purge process

sstefano69:


sstefano69:
developed by mark

PR #467, opened 2021-09-02
FOGL-5704-Add escaping to readings fetch using datapoint toString

MarkRiddoch:
This change has a potential to cause issues if upstream escaping is added so needs to be system tested against a set of plugins.


pintomax (approved):

YashTatkondawar (approved):
System tests look good with this PR.

PR #469, opened 2021-09-06
FOGL-5858 Restore API fixes with specific backup id

ashish-jabble:

_Tested with Sqlite & PostgreSQL engine both_

Custom Packages are available at http://archives.fledge-iot.org/fixes/FOGL-5858/

Also with the given change No regression is introduced in systems/unit tests. Related CI JOB results here `Fledge%20Nightly%20Jobs/job/fledge_nightly_build_ub1804/837/testReport/`


MarkRiddoch (approved):

python/fledge/services/core/scheduler/scheduler.py
```diff
@@ -179,6 +179,8 @@ def __init__(self, core_management_host=None, core_management_port=None, is_safe
         """Delete finished task rows when they become this old"""
         self._purge_tasks_task = None  # type: asyncio.Task
         """asynico task for :meth:`purge_tasks`, if scheduled to run"""
+        self._restore_backup_id = None # type: int
+        """Restore backup id and it will be used when SCHEDULE_RESTORE_ON_DEMAND runs"""
```
praveen-garg, line 183: We must avoid such hack! 
I'm failing to understand, why this `_restore_backup_id` is a member of schedule object. We should use a better approach for this and not violate the object oriented design. There can be n number of such type of processes, we should not involve all those in it.

sstefano69 (approved):

PR #475, opened 2021-09-16
FOGL-5934: fetch next readings block in case of all filtered out readings

pintomax:
FOGL-5934: fetch next readings block in case of all filtered out readings

### DIFF
--- a/C/services/north/data_send.cpp
+++ b/C/services/north/data_send.cpp
@@ -71,6 +71,7 @@ void DataSender::sendThread()
 				"Sending thread closing down after failing to fetch readings");
 			return;
 		}
+		bool removeReadings = false;
 		if (readings->getCount() > 0)
 		{
 			unsigned long lastSent = send(readings);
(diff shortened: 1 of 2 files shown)


pintomax:
test scenario:

Add a South service with sinusoid plugin
Add a North service (ie httpc plugin sending data to another instance)
Add an enabled asset filter to North service:
```
{
  "rules": [
      {
            "asset_name": "___sinusoid",
           "action": "exclude"
      }
   ]
}
```

Readings data are being sent.

Now change the rule:
"asset_name": "sinusoid"


All the readings are filtered out an no data sent to North end

This fix allows reading next blocks from readings database.

After some time, change the rule
 "asset_name": "___sinusoid",
 
 Readings data are being sent to North starting from last saved position in streams table.

YashTatkondawar (approved):
Tried the reconfig of asset filter (attached to north service) by changing asset name, and this is working as expected.

MarkRiddoch (approved):

PR #488, opened 2021-09-29
FOGL-5973: fix for retry parameter in North service

pintomax:
FOGL-5973: fix for retry parameter in North service

tested with

sleepTime = 1
MaximumRetry = 3
httpTimeout = 10

Setup:

South service with sinusoid plugin
North service with httpc plugin configured to send data to another Fledge instance

Step 1: change the IP address in URL address to something behind firewall or not existent in the LAN

Step 2: after a bit change back the IP address to good one:

reconfiguration will happen after retries and timeout and readings data are going to north destination

### DIFF
--- a/C/services/north/data_send.cpp
+++ b/C/services/north/data_send.cpp
@@ -173,10 +173,8 @@ unsigned long DataSender::send(ReadingSet *readings)
 void DataSender::pause()
 {
 	unique_lock<mutex> lck(m_pauseMutex);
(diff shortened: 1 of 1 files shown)


pintomax:
@YashTatkondawar Tested in Ubuntu 18 and Ubuntu 20 (where the issue is more evident)

Need to test this change on Pi4 as well

MarkRiddoch (approved):

PR #505, opened 2021-10-13
FOGL-6012: purge by row doesn't purge if purge exclusion is used

sstefano69:


pintomax (approved):

MarkRiddoch (commented):
I don't understand how this works? It appears to simply turn off the exclusion, which woudl remove the things that are excluded.

PR #509, opened 2021-10-15
FOGL-5665 Addition of new datapoint types and generic classes to handle transforming readings to Python

MarkRiddoch:
Eventually the idea is to replace all the implementation s for filters, south and north plugins to use the PythonReading and PythonReadingSet classes rather than have the duplication that we currently have. We also should look at the likes of the Python35 filter plugin, rule plugin, etc.

We have two new datapoint types, a data buffer, which is transformed to a numpy array in Python and a simple image format, again gets passed to Python as a two dimensional bumpy array. In JSON they are stored as base64 encoded data, so are not suitable for very large objects. The aim of them is to allow image and radiometric data to be passed into ML models written in Python.


docs/OMF.rst
```diff
@@ -164,7 +172,11 @@ The second screen will request the following information:
    - **Number Format:** Used to match Fledge data types to the data type configured in PI. The defaults is float64 but may be set to any OMF datatype that supports floating point values.
    - **Compression:** Compress the readings data before sending it to the PI System.
 
+<<<<<<< HEAD
```
AmandeepArora, line 175: Conflicts needs to be resolved in this file.

MarkRiddoch: Thanks had not noticed that.

pintomax (dismissed):

AmandeepArora (dismissed):

pintomax (approved):

pintomax (approved):

pintomax (approved):

pintomax (approved):

pintomax (approved):

pintomax (approved):

C/services/south-plugin-interfaces/python/python_plugin_interface.cpp
```diff
@@ -79,6 +79,13 @@ void *PluginInterfaceInit(const char *pluginName, const char * pluginPathName)
     PyList_Append(sysPath, PyUnicode_FromString((char *) southRootPath.c_str()));
     PyList_Append(sysPath, PyUnicode_FromString((char *) fledgePythonDir.c_str()));
 
+    // Set sys.argv for embedded Python 3.5
```
pintomax, line 82: Don't we need same fix for

C/services/notification-plugin-interfaces/python/python_plugin_interface.cpp ?

if yes, please add it

pintomax (changes requested):

C/services/south-plugin-interfaces/python/python_plugin_interface.cpp
MarkRiddoch: @AmandeepArora I am assuming the answer to this is yes we should add this for the notifications, correct?

AmandeepArora: Yes, adding.

AmandeepArora: Done, please verify using some notification python plugin by importing tensorflow.

pintomax: @AmandeepArora please suggest something easy to test. Some startup example code.
I.e. plugin skeleton, instructions for installing tensorflow etc

I just have the notification plugins with Numpy and nothing else.

AmandeepArora: You can install tensorflow using "pip install tensorflow"
And in python plugin, at top you can add "import tensorflow as tf" and in plugin_init, you can print "tf.__version__" using _LOGGER.info. That should suffice, earlier we were getting crash dump just at import time.

AmandeepArora: It is "`tf.__version__`"

praveen-garg:
@AmandeepArora  could you please fix the name for `tests/system/plugins/filter/fledge-filter-imageblock` to `tests/system/plugins/filter/imageblock`? 
May be better to put this under tests/system/python/fledge/plugins/filter/imageblock

praveen-garg:
@AmandeepArora  could you please fix the name for `tests/system/plugins/filter/fledge-filter-imageblock` to `tests/system/plugins/filter/imageblock`? 
May be better to put this under tests/system/python/fledge/plugins/filter/imageblock

AmandeepArora:
> @AmandeepArora could you please fix the name for `tests/system/plugins/filter/fledge-filter-imageblock` to `tests/system/plugins/filter/imageblock`? May be better to put this under tests/system/python/fledge/plugins/filter/imageblock

Done

pintomax (dismissed):

pintomax (approved):

AmandeepArora (approved):

pintomax (approved):

AmandeepArora (approved):

YashTatkondawar:
As we have removed a couple of default north plugins(OCS and pi_server), we need to modify the API test https://github.com/fledge-iot/fledge/blob/develop/tests/system/python/api/test_plugin_discovery.py accordingly.

pintomax (dismissed):

AmandeepArora (approved):

pintomax (dismissed):

pintomax (dismissed):

AmandeepArora (approved):

pintomax (approved):

PR #523, opened 2021-11-09
Improve statistics api system test

YashTatkondawar:


tests/system/python/api/test_statistics.py
```diff
@@ -83,22 +83,25 @@ def test_default_statistics_history(self, fledge_url):
...
+            r = r.read().decode()
+            jdoc = json.loads(r)
+            assert len(jdoc), "No data found"
+            assert 15 == jdoc['interval']
+            assert 1 == len(jdoc['statistics'])
+            statistics = jdoc['statistics'][0]
+            if i <= retries and Counter(STATS_HISTORY_KEYS) != Counter(statistics.keys()): continue
+            assert Counter(STATS_HISTORY_KEYS) == Counter(statistics.keys())
+            del statistics['history_ts']
+            assert Counter([0, 0, 0, 0, 0, 0]) == Counter(statistics.values())
```
praveen-garg, line 103: will not assert make tests fail if not as expected in initial iterations? may be we can check/wait for stats history task?

YashTatkondawar: In the first iteration, we have a sleep at the start and then also checked with `if i <= retries and Counter(STATS_HISTORY_KEYS) != Counter(statistics.keys()): continue`; if we do not get anything it will continue the loop

praveen-garg (approved):

PR #528, opened 2021-11-15
FOGL-5144: create, validate, verify JWT service token with needed claims; pass startup token to all services (except storage and API)

pintomax:
FOGL-5144: create jwt service token with needed claims


pintomax:
Start service SIN1 with --token="some_data" and log level debug

Simple test:
Check logs for service registration

Nov 15 14:23:21 ubuntu18-server-virt Fledge[30606] DEBUG: server: fledge.services.core.server: SIN1 SERVER RESPONSE: {'id': '4a30f1bd-fa13-4ab5-b21d-806becfc695a', 'message': 'Service registered successfully', 'bearer_token': 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJhdWQiOiJGbGVkZ2UiLCJzdWIiOiJTSU4xIiwiaXNzIjoiU291dGhib3VuZCIsImV4cCI6MTYzNjk4ODAwMX0.fGClzERZZiD-AGi-7SjGUoV2uIeDXGft7y9xBbpBKqI'}

Take 2 part of jwt token: eyJhdWQiOiJGbGVkZ2UiLCJzdWIiOiJTSU4xIiwiaXNzIjoiU291dGhib3VuZCIsImV4cCI6MTYzNjk4ODAwMX0

decode with base64 and see the token claims

echo 'eyJhdWQiOiJGbGVkZ2UiLCJzdWIiOiJTSU4xIiwiaXNzIjoiU291dGhib3VuZCIsImV4cCI6MTYzNjk4ODAwMX0' | base64 -d | jq
```
{
  "aud": "Fledge",
  "sub": "SIN1",
  "iss": "Southbound",
  "exp": 1636988001
}
```

python/fledge/services/core/server.py
```diff
@@ -67,6 +68,10 @@
 
 SSL_PROTOCOLS = (asyncio.sslproto.SSLProtocol,)
 
+SERVICE_JWT_SECRET = 'f0gl@mp+Fl3dG3'
```
MarkRiddoch, line 71: The secret is not very secret if it is in a public repository. I wonder if we could generate a random secret at build time in much the same way as we generate the version.h.

pintomax: We have a similar one in python/fledge/services/core/user_model.py and that is for user authentication.

A possible new solution will address both tokens generation

pintomax:
Example:

Fledge management port 40623

curl -v -X POST localhost:40623/fledge/service/verify_token -d '{"bearer_token" : "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJhdWQiOiJGbGVkZ2UiLCJzdWIiOiJOT1RTIiwiaXNzIjoiTm90aWZpY2F0aW9uIiwiZXhwIjoxNjM3NjYzOTc2fQ.64rIfkkKrIb7oh963cM29P60-az6ROmrb-bvY7fFiGo"}' | jq

Response
```
{
  "aud": "Fledge",
  "sub": "NOTS",
  "iss": "Notification",
  "exp": 1637663976
}
```

pintomax:
Token refresh example

Fledge management port 43015

ZGdrdmV4dWFsY2dsaHVyZmFxdmdybXB5dXduaXJvc3g= is base64 of startup token of SIN1 service (dgkvexualcglhurfaqvgrmpyuwnirosx)

curl -v -X POST -H 'Authorization: ZGdrdmV4dWFsY2dsaHVyZmFxdmdybXB5dXduaXJvc3g=' localhost:43015/fledge/service/refresh_token -d '{"bearer_token" : "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJhdWQiOiJGbGVkZ2UiLCJzdWIiOiJTSU4xIiwiaXNzIjoiU291dGhib3VuZCIsImV4cCI6MTYzNzg1NTU3MX0.a03PCYB6DIrqbdIkDvT42m_MJbniF-DHUB18h8m6RU4"}' | jq

response

```
{
     "bearer_token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJhdWQiOiJGbGVkZ2UiLCJzdWIiOiJTSU4xIiwiaXNzIjoiU291dGhib3VuZCIsImV4cCI6MTYzNzg1NTYxOH0.tJYMIlxbOOYU5rQkjsCnBWXTGLj2S1gJ8vp_B1irUxE"
}
```

Note: service XYZ can refresh only its own token

python/fledge/services/common/microservice_management/routes.py
```diff
@@ -58,6 +58,12 @@ def setup(app, obj, is_core=False):
         # Internal refresh cache
         app.router.add_route('PUT', '/fledge/cache', obj.refresh_cache)
 
+        # Service token verification
+        app.router.add_route('POST', '/fledge/service/verify_token', obj.verify_token)
+
+        # Service token refresh
+        app.router.add_route('POST', '/fledge/service/refresh_token', obj.refresh_token)
```
praveen-garg, line 65: better if we keep as,

PUT   /fledge/service/token/refresh
POST /fledge/service/token/verify

python/fledge/services/core/scheduler/scheduler.py
```diff
@@ -128,6 +130,9 @@ def __init__(self):
     _storage = None
     _storage_async = None
 
+    # Startup tokens to pass to service or tasks being started
+    _startupTokens = dict()
```
praveen-garg, line 134: I believe this should be with service registry or core server.py and only for services (schedule type STARTUP)

python/fledge/services/core/scheduler/scheduler.py
```diff
@@ -303,6 +311,16 @@ async def _start_task(self, schedule: _ScheduleRow) -> None:
         args_to_exec.append("--address=127.0.0.1")
         args_to_exec.append("--name={}".format(schedule.name))
 
+        # Check a service / task token exists or create a new one
+        startToken = self._startupTokens.get(schedule.name, None)
+        if startToken is None:
+            letters = string.ascii_lowercase
```
praveen-garg, line 317: much better if string.ascii_letters

python/fledge/services/core/scheduler/scheduler.py
```diff
@@ -303,6 +311,16 @@ async def _start_task(self, schedule: _ScheduleRow) -> None:
...
+            startToken = ''.join((random.choice(string.ascii_lowercase) for x in range(32)))
+            self._startupTokens[schedule.name] = startToken
```
praveen-garg, line 319: We should ask core service registry to assign a single use token and then use that to check and allow registration?

python/fledge/services/core/scheduler/scheduler.py
```diff
@@ -303,6 +311,16 @@ async def _start_task(self, schedule: _ScheduleRow) -> None:
...
+
+            # Add startup token to args
+            args_to_exec.append("--token={}".format(startToken))
```
praveen-garg, line 322: only if schedule type is STARTUP i.e. services?

python/fledge/services/core/scheduler/scheduler.py
```diff
@@ -1287,6 +1305,9 @@ async def disable_schedule(self, schedule_id: uuid.UUID, bypass_check=None):
                 if task_future.cancel() is True:
                     await self._wait_for_task_completion(task_process)
 
+        # Delete startup token
+        self._startupTokens.pop(schedule.name, None)
```
praveen-garg, line 1309: on succesful registration, the set/dict holding that single use token should pop that.

python/fledge/services/core/server.py
praveen-garg: please add a TODO

python/fledge/services/core/server.py
```diff
@@ -1115,7 +1120,14 @@ async def register(cls, request):
             if token is not None:
                 if not isinstance(token, str):
                     msg = 'Token can be a string only'
-                    raise web.HTTPBadRequest(reason=msg, body=json.dumps(msg))
+                    raise web.HTTPBadRequest(reason=msg)
```
praveen-garg, line 1123: we should add a body as well with json object

python/fledge/services/core/server.py
```diff
@@ -1115,7 +1120,14 @@ async def register(cls, request):
...
+
+                # Check startup token exists
+                foundToken = cls.scheduler._startupTokens.get(service_name)
+                if foundToken is None or foundToken != token:
+                    msg = 'Token for the service was not found' 
+                    raise web.HTTPBadRequest(reason=msg)
```
praveen-garg, line 1129: We better log and separate the check, not found or invalid supplied

python/fledge/services/core/server.py
```diff
@@ -1138,7 +1150,22 @@ async def register(cls, request):
...
 
-            bearer_token = "{}_{}".format(service_name.strip(), uuid.uuid4().hex) if token is not None else ""
+            # Set JWT bearewr token
+            # Set expiration
+            exp = datetime.now() + timedelta(seconds=SERVICE_JWT_EXP_DELTA_SECONDS)
+            # Add public token claims
+            claims = {
+                         'aud': SERVICE_JWT_AUDIENCE,
+                         'sub' : service_name,
+                         'iss' : service_type,
```
praveen-garg, line 1160: should not the issuer be Fledge and audience the service_type?

python/fledge/services/core/server.py
```diff
@@ -1615,3 +1642,107 @@ async def add_audit(cls, request):
...
+             -H 'Authorization: ZGdrdmV4dWFsY2dsaHVyZmFxdmdybXB5dXduaXJvc3g='
+             http://localhost:<core mngmt port>/fledge/service/refresh_token
+
+        Authorization header must contain the startup token in base64 format
+
+        Note: token will be refresh for the service it belongs
+        """
+
+        data = await request.json()
+        authData = request.headers.get('Authorization', "")
```
praveen-garg, line 1709: default value from get empty?

python/fledge/services/core/server.py
```diff
@@ -1615,3 +1642,107 @@ async def add_audit(cls, request):
...
+        auth = base64.b64decode(authData).decode('utf-8')
+        bearer_token = data.get('bearer_token',"")
```
praveen-garg, line 1711: default value from get empty?

python/fledge/services/core/server.py
```diff
@@ -1615,3 +1642,107 @@ async def add_audit(cls, request):
...
+
+        try:
+            for k, v in cls.scheduler._startupTokens.items():
```
praveen-garg, line 1714: why do we need to walk through startup/single use tokens for refreshing the token?
I believe we need to decode the refresh token and then use the subject to find ServiceRegistry.get(name={}) and then match the token, refresh and assign new.

python/fledge/services/core/server.py
```diff
@@ -1615,3 +1642,107 @@ async def add_audit(cls, request):
...
+
+                                # Find service name in registry and update the bearer token for that service
+                                obj = ServiceRecord(s_id=service._id,
+                                                    s_name=service._name,
+                                                    s_type=service._type,
+                                                    s_port=service._port,
+                                                    m_port=service._management_port,
+                                                    s_address=service._address,
+                                                    s_protocol=service._protocol,
+                                                    s_bearer_token=bearer_token)
```
praveen-garg, line 1735: as pointed out in previous pull requests, we should be careful having this in slots and printing tokens in logs/repr things.

python/fledge/services/core/server.py
```diff
@@ -1615,3 +1642,107 @@ async def add_audit(cls, request):
...
+                                        ServiceRegistry._registry[idx] = obj
+
+                                return web.json_response(ret)
+
+            msg = 'service authentication failed for service'
+            raise web.HTTPBadRequest(reason=msg, body=json.dumps({"error": msg}))
+
+        except Exception as e:
+            msg = str(e)
+            raise web.HTTPBadRequest(reason=msg, body=json.dumps({"error": msg}))
```
praveen-garg, line 1747: error key is correct, though as we are using `message` for error response body at other places, better we use message in these changes.

python/fledge/services/common/microservice_management/routes.py
MarkRiddoch: POST should be used to create objects, neither of these are creating objects.

python/fledge/services/core/scheduler/scheduler.py
MarkRiddoch: It would make more sense if there are created and maintained by the service registry since that will be doing the checking. It also gives us a single place to tighten up on the way they are used.

MarkRiddoch (commented):
This still seems to be failing the PR tester, we need to fix that.

MarkRiddoch (dismissed):

python/fledge/services/core/server.py
```diff
@@ -1237,6 +1280,18 @@ async def get_auth_token(cls, request: web.Request) -> web.Response:
             :Example:
                 curl -sX GET http://localhost:<core mgt port>/fledge/service/authtoken?_vid=<>
         """
+        bearer_token = request.headers.get('bearer_token', "")
```
praveen-garg, line 1283: we should handle the Auth token via request header here the same way, we did in last commits. Eventually this endpoint will go away as we had this interim solution for (foglamp-management) poll agent service. and that will be updated to use the originally planned mechanism with (single use) token arg and return JWT token on register.

python/fledge/services/core/server.py
```diff
@@ -1623,3 +1678,120 @@ async def add_audit(cls, request):
...
+            services_list = ServiceRegistry.all()
+            for service in services_list:
+                t = service._bearer_token if service._bearer_token is not None else ""
+                if t == bearer_token:
+                    foundToken = True
+                    break
+            # Raise error if token does not exists
+            if foundToken == False:
+                msg = 'token does not exist in system'
+                raise web.HTTPBadRequest(reason=msg, body=json.dumps({"error": msg}))
```
praveen-garg, line 1714: We should avoid looping, and should have a private dict for service name and jwt tokens (same way we have for single use tokens) and first,
decode the token, if decoded okay, find sub/aud and allow to proceed if matching that pair.

python/fledge/services/core/server.py
```diff
@@ -1623,3 +1678,120 @@ async def add_audit(cls, request):
...
+             http://localhost:<core mngmt port>/fledge/service/refresh_token
+
+        Authorization header must contain the Bearer token
+        No post data
+
+        Note: token will be refresh for the service it belongs
+        """
+
+        authData = request.headers.get('Authorization', "")
+        parts = authData.split("Bearer ")
```
praveen-garg, line 1756: if authData is not ""? or may be auth_header = request.headers.get('Authorization', None) and then if auth_header

python/fledge/services/core/server.py
pintomax: This is related to "Poll service"

python/fledge/services/core/server.py
pintomax: I think this internal change could be postponed

praveen-garg: Having token attribute in slots as ServiceRecord member is not a good idea. As I stated earlier (in original commits PR), we are at risk of exposing tokens, having a better control is important. 
We must avoid from `def __repr__` and can see possible risks which will need explicit needs to delattr before sharing service entries among app components.

python/fledge/services/common/microservice_management/routes.py
praveen-garg: PUT is better to refresh. GET is better to verify as we don't need that explicit payload anymore but auth request header, Otherwise I see no harm in POSTing a request to get verified. 
also see URL constructs

python/fledge/services/core/server.py
praveen-garg: Add a Jira for this. We should remove this `request.headers.get('bearer_token', "")` thing anyway and use Authorization header. Until then poll agent can continue to work on old mechanism using `_vid`/ explicit verification_id

MarkRiddoch (approved):

python/fledge/services/core/server.py
pintomax: @praveen-garg needed changes have been applied

python/fledge/common/service_record.py
```diff
@@ -45,9 +45,9 @@ class InvalidServiceStatus(Exception):
         # TODO: tell allowed service status?
         pass
 
-    __slots__ = ['_id', '_name', '_type', '_protocol', '_address', '_port', '_management_port', '_status', '_token']
+    __slots__ = ['_id', '_name', '_type', '_protocol', '_address', '_port', '_management_port', '_status']
 
-    def __init__(self, s_id, s_name, s_type, s_protocol, s_address, s_port, m_port, s_token=None):
+    def __init__(self, s_id, s_name, s_type, s_protocol, s_address, s_port, m_port, s_startup_token=None):
```
praveen-garg, line 50: what is the use of this `s_startup_token=None` arg?

python/fledge/services/core/scheduler/scheduler.py
```diff
@@ -329,6 +329,12 @@ async def _start_task(self, schedule: _ScheduleRow) -> None:
         # jwt token for cross communication
         # args_to_exec.append("--token={}".format(t))
```
praveen-garg, line 330: remove this commented code

python/fledge/services/core/server.py
```diff
@@ -1246,6 +1282,18 @@ async def get_auth_token(cls, request: web.Request) -> web.Response:
             :Example:
                 curl -sX GET http://localhost:<core mgt port>/fledge/service/authtoken?_vid=<>
         """
+        bearer_token = request.headers.get('bearer_token', "")
+        if bearer_token:
+            services_list = ServiceRegistry.all()
+            is_bearer_token_valid = False
+            for svc in services_list:
+                if svc._token == bearer_token:
```
praveen-garg, line 1290: this check should be fixed

pintomax: This is not related to the code changed so far.

If bearer_token has in "sub" claim the service name then ServiceRegistry.getBearerToken(claims['sub'])
can be used.

Could you check this?

praveen-garg: Yes, And can be a common code as we see in other methods.

python/fledge/services/core/server.py
```diff
@@ -1633,6 +1681,120 @@ async def add_audit(cls, request):
...
+        bearer_token = parts[1]
+
+        if bearer_token is not None:
+            claims = cls.validate_token(bearer_token)
+            if claims.get('error') is not None:
+                msg = claims['error']
+                raise web.HTTPBadRequest(reason=msg, body=json.dumps({"error": msg}))
+
+            # Check input token exists in system for the service name given in claims['sub']
+            foundToken = ServiceRegistry.getBearerToken(claims['sub'])
```
praveen-garg, line 1717: also match the token

MarkRiddoch (approved):

MarkRiddoch (approved):

python/fledge/services/core/server.py
```diff
@@ -1120,10 +1125,17 @@ async def register(cls, request):
...
 
+            # If token then check single use token verification; if bad then return 4XX
             if token is not None:
                 if not isinstance(token, str):
                     msg = 'Token can be a string only'
-                    raise web.HTTPBadRequest(reason=msg, body=json.dumps(msg))
+                    raise web.HTTPBadRequest(reason=msg)
+
+                # Check startup token exists
+                if ServiceRegistry.checkStartupToken(service_name, token) == False:
```
praveen-garg, line 1135: @pintomax We must check if a startup token was issued and while registering then token is not present in payload. 

if ServiceRegistry.getStartupToken then ensure token is not data.get('token', None) and match. Here we are relying only if supplied by consumer service. and allow registration with None. The None is allowed only for those which were not issued any startup token.

pintomax: @praveen-garg is that resolved by your last commit?

praveen-garg: nope

pintomax: Would you mind adding that check while I'm testing all the connected services along with dispatcher?

C/common/bearer_token.cpp
```diff
@@ -0,0 +1,61 @@
+/*
+ * Fledge bearer token utilities
+ *
+ * Copyright (c) 2021 Dianomic Systems
```
MarkRiddoch, line 4: 2022

C/common/bearer_token.cpp
```diff
@@ -0,0 +1,61 @@
...
+ * Author: Massimiliano Pinto
+ */
+
+#include "bearer_token.h"
+
+using namespace std;
+using HttpServer = SimpleWeb::Server<SimpleWeb::HTTP>;
+
+/**
+ * Extract access bearer token from request object
```
MarkRiddoch, line 17: Should say what happens if the request has no bearer token

C/common/include/bearer_token.h
```diff
@@ -0,0 +1,21 @@
+#ifndef _BEARER_TOKEN_H
+#define _BEARER_TOKEN_H
+/*
+ * Fledge bearer token utilities
+ *
+ * Copyright (c) 2021 Dianomic Systems
```
MarkRiddoch, line 6: 2022

C/common/bearer_token.cpp
```diff
@@ -0,0 +1,61 @@
...
+ * Released under the Apache 2.0 Licence
+ *
+ * Author: Massimiliano Pinto
+ */
+
+#include "bearer_token.h"
+
+using namespace std;
+using HttpServer = SimpleWeb::Server<SimpleWeb::HTTP>;
+
```
MarkRiddoch, line 15: I don't really like these free floating C functions and the use of the vector. It would have been better to have a class to represent the bearer token, the constructor would be the thing to extract it from the request, etc.

C/common/management_client.cpp
```diff
@@ -1,25 +1,27 @@
 /*
  * Fledge storage service.
  *
- * Copyright (c) 2017-2018 OSisoft, LLC
+ * Copyright (c) 2017-2021 OSisoft, LLC
```
MarkRiddoch, line 4: 2022 - but you don't need to update this

C/common/management_client.cpp
```diff
@@ -864,3 +867,308 @@ bool ManagementClient::addAuditEntry(const std::string& code,
...
+				}
+				claims.clear();
+				verified = false;
+			}
+			else
+			{
+				verified = true;
+
+				// Set token claims in the input map
+				if (doc["aud"].IsString() &&
```
MarkRiddoch, line 963: If one of these is missing or not a string the code will abort - are you sure tis can never happen?

C/common/management_client.cpp
```diff
@@ -864,3 +867,308 @@ bool ManagementClient::addAuditEntry(const std::string& code,
...
+			claims.clear();
+			ret = false;
+			m_logger->error("Micro service bearer token '%s' not verified.",
+					bearer_token.c_str());
+		}
+	}
+	else
+	{
+		// A validated token exists: check expiration time
+		vector<string> elems = JWTTokenSplit(bearer_token, '.');
```
MarkRiddoch, line 999: A bearer token class would have made this so much cleaner

C/common/management_client.cpp
```diff
@@ -864,3 +867,308 @@ bool ManagementClient::addAuditEntry(const std::string& code,
...
+			m_logger->error("Failure while parsing JSON claims of bearer token, "
+					"error %s, input data %s",
+					GetParseError_En(doc.GetParseError()),
+					plainData.c_str());
+			claims.clear();
+			return false;
+		}
+
+		unsigned long expiration;
+		if (doc["exp"].IsUint())
```
MarkRiddoch, line 1026: Add HasMember check

C/common/management_client.cpp
```diff
@@ -864,3 +867,308 @@ bool ManagementClient::addAuditEntry(const std::string& code,
...
+			m_received_tokens.erase(bearer_token);
+
+			m_logger->error("Micro service bearer token '%s' has expired.",
+					bearer_token.c_str());
+			claims.clear();
+		}
+		else
+		{
+			// Token still valid, set token claims in the input map
+			claims["aud"] = doc["aud"].GetString();
```
MarkRiddoch, line 1055: Again this could abort if any of these are missing or the wrong type

C/common/management_client.cpp
```diff
@@ -864,3 +867,308 @@ bool ManagementClient::addAuditEntry(const std::string& code,
...
+}
+
+/**
+ * Refresh the JWT bearer token string
+ *
+ * @param request       HTTP request object
+ * @param claims        Map to fill with JWT public token claims
+ * @return              New bearer token on success, empty string otherwise
+ */
+string ManagementClient::refreshBearerToken(const string& bearer_token)
```
MarkRiddoch, line 1100: There are lots of copies going on here, because of object returns. Would it be better to pass std::string&  and avoid the copies. You could then return a boolean status and not have the look for empty string to detect failure.

C/common/management_client.cpp
```diff
@@ -864,3 +867,308 @@ bool ManagementClient::addAuditEntry(const std::string& code,
...
+			{
+				m_logger->error("Failed to parse token refresh result: %s",
+						response.c_str());
+			}
+			ret = false;
+		}
+		else if (doc.HasMember("bearer_token"))
+		{
+			// Set new token
+			newToken = doc["bearer_token"].GetString();
```
MarkRiddoch, line 1154: Check for IsString to avoid possible abort

python/fledge/services/core/server.py
MarkRiddoch: Please add the TODO here

MarkRiddoch (changes requested):
We should think about how we are going to debug services with this in place. Having the option of not passing the --token and running the service as unauthenticated would probably make sense in all cases other than when control is used.

MarkRiddoch (approved):

praveen-garg (commented):
Looks okay to me. We should fix repetitive python code, spilled over different def for parsing/ decoding the token, especially python/fledge/services/core/server.py

python/fledge/services/core/server.py
```diff
@@ -1673,3 +1789,53 @@ async def add_and_enable_dispatcher(cls):
         schedule.enabled = False
         # Save schedule
         await cls.scheduler.save_schedule(schedule, is_enabled)
+
+    @classmethod
+    def get_token_common(cls, request):
```
praveen-garg, line 1794: can we name this better?

MarkRiddoch (approved):

PR #550, opened 2022-01-20
FOGL-6146: getService of AssetTracker implemented

pintomax:
FOGL-6146: getService of AssetTracker implemented

Example method call:

AssetTracker::getAssetTracker()->getService("Ingest", "asset_1");

### DIFF
--- a/C/common/include/asset_tracking.h
+++ b/C/common/include/asset_tracking.h
@@ -102,13 +102,14 @@ public:
 		{
 			return getService("Egress", asset);
 		};
+	std::string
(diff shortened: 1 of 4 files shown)


C/common/include/asset_tracking.h
```diff
@@ -102,13 +102,14 @@ class AssetTracker {
 		{
 			return getService("Egress", asset);
 		};
+	std::string
+		getService(const std::string& event, const std::string& asset);
```
MarkRiddoch, line 106: Why has this become public?

MarkRiddoch (approved):

PR #555, opened 2022-01-27
FOGL-6156: Notification rule plugin supports DPImage data type

pintomax:
FOGL-6156: Notification rule plugin supports DPImage data type

Removed notification_shim.py
Removed new Interpreter


MarkRiddoch (approved):
It's a shame we have to duplicate the code here that is in the PythonReading class and we could not find a way to reuse that code. I understand why you have it like this. I think we should go ahead with what you have here and think about how we might share the code in future.

PR #558, opened 2022-01-31
FOGL-6115

AmandeepArora:
Changes and testing completed.


C/services/common-plugin-interfaces/python/pyobject_reading_parser.cpp
```diff
@@ -108,13 +181,21 @@ std::vector<Reading *>* Py2C_getReadings(PyObject *polledData)
...
 					newReadings->push_back(newReading);
+				}
 			}
 		}
 	}
 	return newReadings;
 }
+#endif
 
+#if 0
```
MarkRiddoch, line 198: If we really don't need this code then let's remove it to avoid confusion.

C/services/common-plugin-interfaces/python/pyobject_reading_parser.cpp
```diff
@@ -19,8 +19,9 @@
 
 extern "C" {
 
-static void logErrorMessage();
+// static void logErrorMessage();
 
+#if 0
 /**
```
MarkRiddoch, line 25: If we really don't need this code then let's remove it to avoid confusion.

C/services/common-plugin-interfaces/python/pyobject_reading_parser.cpp
```diff
@@ -52,7 +53,72 @@ Reading* Py2C_parseReadingObject(PyObject *element)
 	
 	return new PythonReading(element);
 }
+#endif
 
+#if 0
```
MarkRiddoch, line 58: If we really don't need this code then let's remove it to avoid confusion.

C/services/common-plugin-interfaces/python/pyobject_reading_parser.cpp
```diff
@@ -52,7 +53,72 @@ Reading* Py2C_parseReadingObject(PyObject *element)
...
+    //Logger::getLogger()->debug("fill=%s, newReading->getUserTimestamp()=%d, newReading->getTimestamp()=%d", fill?"True":"False", newReading->getUserTimestamp(), newReading->getTimestamp());
+    if (fill)
+    {
+        newReading->setUserTimestamp(newReading->getTimestamp());
+        //Logger::getLogger()->debug("Copied TS into user TS: newReading->getUserTimestamp()=%d", newReading->getUserTimestamp());
+    }
+}
+#endif
+
+#if 0
```
MarkRiddoch, line 121: If we really don't need this code then let's remove it to avoid confusion.

C/services/common-plugin-interfaces/python/pyobject_reading_parser.cpp
```diff
@@ -147,7 +228,9 @@ static void logErrorMessage()
 	Py_CLEAR(pValue);
 	Py_CLEAR(pTraceback);
 }
+#endif
 
+#if 0
```
MarkRiddoch, line 233: If we really don't need this code then let's remove it to avoid confusion.

MarkRiddoch (commented):
It's very hard to tell rom a static review if this will do the right thing, looks good however. I would remove the #if 0 code, just to avoid clutter and confusion. 
We seem to have an issue with a unit test the came form the 5665 branch which needs to be looked at.

I have fixed two issues and pushed onto this branch. The 2D float array issue still exists but is less pressing.

C/services/common-plugin-interfaces/python/pyobject_reading_parser.cpp
AmandeepArora: This file has been removed. Forgot to push it earlier.

pintomax (approved):

MarkRiddoch (approved):
Update vector of vector of doubles data points to use pointers to make memory management easier. Also fixed unit test for this.

AmandeepArora:
Thanks @MarkRiddoch for fixing the unit tests.

@Massimiliano - It seems DCO signoff is missing for this commit:  Commit sha: 22d8312, Author: pintomax, Committer: GitHub; The sign-off is missing.

AmandeepArora:
This is obsolete.

PR #563, opened 2022-02-02
Fixed the backup test, failing on Rpi Bullseye

YashTatkondawar:


praveen-garg (approved):
We may need to figure out more around this; and else part of backup api when no .db/.dump files found.

PR #574, opened 2022-02-15
FOGL-6230 ACL control API endpoints fixes; and eval removed

praveen-garg:
- [x] fix Control ACL endpoints 
- [x] fix Control scripts endpoints
- [x] allow removal of ACL from script / in PUT endpoint


pintomax:
Successfully tested 
curl -sX PUT http://localhost:8081/fledge/service/SIN1/ACL -d '{"acl_name": "A"}'

pintomax (dismissed):

pintomax:
Update is fine

mshariq-nerd (dismissed):

python/fledge/services/core/api/control_service/acl_management.py
```diff
@@ -147,19 +155,29 @@ async def add_acl(request: web.Request) -> web.Response:
 
 @has_permission("admin")
 async def update_acl(request: web.Request) -> web.Response:
-    """ Update an access control list. Only the set of service and URL's can be updated
```
ashish-jabble, line 150: This is as per specs
```
Update an ACL
PUT /fledge/ACL/{name}
Update an access control list. Only the set of services and URL’s in the ACL can be updated, this call can not be used to change the name or users of an ACL.
```
Not sure from where the requirement comes from for changing the name of ACL

Also check schema for ACL table we have a primary constraint for name as well. https://github.com/fledge-iot/fledge/blob/develop/scripts/plugins/storage/sqlite/init.sql#L617

cc/ @MarkRiddoch

praveen-garg: Also, the same case for Control script?

I can understand, that we may need to propagate changes to services/scripts on ACL rename. Should we keep this static then in GUI? or later add a different endpoint with propagation to references? 

Flexibility to fix naming is very basic to have.

MarkRiddoch (approved):

pintomax:
We need to wait a bit before merging this PR to develop.

Some C++ classes need code update: FOGL-6240

PR #581, opened 2022-02-21
FOGL-6209 management service bearer token based auth mechanism

praveen-garg:
* removed the interim solution from scheduler.


MarkRiddoch (approved):

pintomax (approved):

python/fledge/services/core/server.py
```diff
@@ -1316,21 +1304,35 @@ async def cert_login(ca_cert):
...
             try:
-                fname = _FLEDGE_ROOT + "/data/.{}".format('managementtoken')
-                with open(fname, 'r', encoding = 'utf-8') as f:
-                    t = f.read()
+                auth_header = request.headers.get('Authorization', None)
             except:
                 raise api_exception.VerificationFailed
             else:
-                if verification_id != t:
+                # check bearer token with service registry for given service
```
praveen-garg, line 1314: todo: add checks,
```
if auth_header:
    if not "Bearer " in auth_header:
        log bad Authorization header data
        return UNAUTHORIZED
else:
   return UNAUTHORIZED
```

@pintomax  probably this should move to a def _check_auth_header along with next repeated (among other def/s) code.

PR #583, opened 2022-02-22
FOGL-6244: check and set empty value for security category

pintomax:
FOGL-6244: check and set empty value for security category


python/fledge/services/core/api/control_service/acl_management.py
```diff
@@ -365,12 +365,14 @@ async def detach_acl_from_service(request: web.Request) -> web.Response:
...
-            else:
-                raise StorageServerError(delete_cat_result)
+            category_desc = "Security category for {} service".format(svc_name)
+            category_value = {}
+            await cf_mgr.create_category(category_name=security_cat_name,
+                                         category_description=category_desc,
+                                         category_value=category_value)
+
+            message = "ACL detached from {} service successfully".format(svc_name)
+            return web.json_response({"message": message})
```
praveen-garg, line 375: this line is redundant as this is in else part of try-except-else.

PR #584, opened 2022-02-22
FOGL-5144 patch: fix the security flaw that allowed registration without token 

praveen-garg:
… token at startup was assigned

### DIFF
--- a/python/fledge/services/core/server.py
+++ b/python/fledge/services/core/server.py
@@ -1125,7 +1125,10 @@ class Server:
             if not isinstance(service_management_port, int):
                 raise web.HTTPBadRequest(reason='Service management port can be a positive integer only')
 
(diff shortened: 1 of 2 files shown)


python/fledge/services/core/server.py
```diff
@@ -1796,6 +1799,12 @@ def get_token_common(cls, request):
             msg = "Authorization header is missing"
             raise web.HTTPBadRequest(reason=msg, body=json.dumps({"error": msg}))
 
+        if not "Bearer " in auth_header:
+            msg = "Invalid Authorization token"
+            # FIXME: raise UNAUTHORISED here and among other places 
+            #   and JSON body to have message key
```
praveen-garg, line 1805: Though `error` is a correct key, but as we are using `message` since this started to be in response body, better to be consistent.

PR #585, opened 2022-02-22
FOGL-5144 patch token check get and issuance set fixes

praveen-garg:


praveen-garg:
this PR fixes the issue #584 correctly.

pintomax (approved):

PR #595, opened 2022-03-10
FOGL-6246 API validation added in control script

ashish-jabble:
ACL API will fix until FOGL-6258 is resolved


ashish-jabble:
@MarkRiddoch Also confirm that response of get all scripts in which order we should preferred. As of now alphabetically Z-A order from API itself. And in GUI with order A-Z as they have implemented order By at their end.

MarkRiddoch (approved):
Looks good to me

PR #610, opened 2022-03-29
FOGL-6322: call southPlugin->start() only for async plugins

pintomax:
FOGL-6322: call southPlugin->start() only for async plugins

### DIFF
--- a/C/services/south/south.cpp
+++ b/C/services/south/south.cpp
@@ -432,12 +432,7 @@ void SouthService::start(string& coreAddress, unsigned short corePort)
 				else
 				{
 					Logger::getLogger()->debug("Plugin does not persist data");
-					try {
-						southPlugin->start();
-						started = true;
-					} catch (...) {
-						Logger::getLogger()->debug("Plugin start raised an exception");
-					}
+					started = true;
 				}
 				if (!started)
 				{


praveen-garg:
@Innovative-ashwin is this related to the issue you were facing while develop the plctag (poll) plugin?

C/services/south/south.cpp
```diff
@@ -432,12 +432,7 @@ void SouthService::start(string& coreAddress, unsigned short corePort)
 				else
 				{
 					Logger::getLogger()->debug("Plugin does not persist data");
-					try {
-						southPlugin->start();
-						started = true;
-					} catch (...) {
-						Logger::getLogger()->debug("Plugin start raised an exception");
-					}
+					started = true;
```
MarkRiddoch, line 435: This is fine for the non-async plugins, but how is plugin_start called if it is async?

pintomax: There is an else branch and around line 537

plugin->start() is called

MarkRiddoch (approved):

Innovative-ashwin:
> 



> 

@praveen-garg , I was getting the issue in plctag as plugin->start is  called in case of poll plugins as well  from this part of south.cpp:

  Logger::getLogger()->debug("Plugin does not persist data");
416                                         try {
417                                                 southPlugin->start();
418                                                 started = true;
419                                         } catch (...) {
420                                                 Logger::getLogger()->debug("Plugin start raised an exception");
421                                         }

pintomax:
@Innovative-ashwin @praveen-garg  
We can cherry pick this commit into develop if needed

PR #615, opened 2022-04-05
FOGL-6293 Support for managing control schedule tasks

ashish-jabble:
- [x] Provided single API for all schedules along with configuration in existing GET script endpoints.
- [x] Unit tests 

For UPDATE & DELETE - Use existing schedule & configuration endpoint.

**NOTE** - Reset control_script table if not empty


python/fledge/services/core/api/control_service/script_management.py
```diff
@@ -30,14 +30,144 @@
...
+            schedules.append({
+                'id': str(sch.schedule_id),
+                'name': sch.name,
+                'processName': sch.process_name,
+                'type': Schedule.Type(int(sch.schedule_type)).name,
+                'repeat': 0,
+                'time': 0,
+                'day': sch.day,
+                'exclusive': sch.exclusive,
+                'enabled': sch.enabled
```
mshariq-nerd, line 74: It should return not only the schedules but also the config data i.e. parameter. In GUI we can always filter these with the process_name. To avoid the multiple calls to fetch the resource details by the client we should have the CRUD operations with schedule and configuration data.

mshariq-nerd (commented):
Also better to have  https://github.com/fledge-iot/fledge/pull/614 as a base this branch. So that we have single point for testing.

ashish-jabble:
#614 is for macro substitution which is independent of this work. And that should go into develop without any hasle. We should not wait for this PR to get complete. As I have parallel work which is going for Fledge API proxy.

python/fledge/services/core/api/control_service/script_management.py
ashish-jabble: @MarkRiddoch 
Please review the requirement and document itself in control service support REST API doc.
IMHO, We may not require these new endpoints, we can use the existing endpoint and add the schedule & category info inside it.

mshariq-nerd:
I am not clear here basically as I need to POST the parameter that part of #614 and GET is in this. How I can work in GUI without both having on same place?  I need to test GET on this and POST on #614 that is what you are saying? Only the schedule data is not the requirement.

MarkRiddoch:
> I am not clear here basically as I need to POST the parameter that part of #614 and GET is in this. How I can work in GUI without both having on same place? I need to test GET on this and POST on #614 that is what you are saying? Only the schedule data is not the requirement.

The GET was only for illustration of what was stored. You should be able to do everything with a POST

MarkRiddoch:
@ashish-jabble I don't understand your note "Reset DB is required" - do you mean there is work required to the database reset 9and presumably purge) or do you mean that to use this you need to first reset the database?

ashish-jabble:
> @ashish-jabble I don't understand your note "Reset DB is required" - do you mean there is work required to the database reset 9and presumably purge) or do you mean that to use this you need to first reset the database?

Reset DB here I mean for newly tables (i.e control_script). As we have changed the category naming. So to avoid any vague behaviour reset only control_script table.

No downgrade/upgrade scripts are required. As this table checked in the new schema ONLY.

MarkRiddoch (approved):

PR #616, opened 2022-04-05
FOGL-6306 Fledge API Proxy

ashish-jabble:
FOGL-6306 - only Fledge Core API
~FOGL-6310 - C management API~
~~FOGL-6310-test => FOGL-6306 + FOGL-6310~~


python/fledge/services/core/api/proxy.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+__author__ = "Ashish Jabble"
+__copyright__ = "Copyright (c) 2022 Dianomic Systems Inc."
+__license__ = "Apache 2.0"
+__version__ = "${VERSION}"
+
+
+_logger = logger.setup(__name__, level=logging.INFO)
+SVC_TYPE = "BucketStorage"
```
MarkRiddoch, line 24: We should not have anything that mentions the service type as part of this is to allow new services in FogLAMP that Fledge does not know about

python/fledge/services/core/api/proxy.py
```diff
@@ -0,0 +1,110 @@
...
+    except Exception as ex:
+        msg = str(ex)
+        raise web.HTTPInternalServerError(reason=msg, body=json.dumps({"message": msg}))
+    else:
+        return web.json_response({"info": result})
+
+
+async def _get_service_record_info_along_with_bearer_token():
+    try:
+        service = ServiceRegistry.get(s_type=SVC_TYPE)
```
MarkRiddoch, line 59: We should not be testing service type for the reasons noted above.

python/fledge/services/core/proxy.py
```diff
@@ -0,0 +1,105 @@
...
+    """ Add API proxy for a service
+
+    :Example:
+             curl -sX POST http://localhost:<SVC_MGT_PORT>/fledge/proxy -d '{"service_name": "BucketStorage", "POST": {"/fledge/bucket": "/bucket"}}'
+             curl -sX POST http://localhost:<SVC_MGT_PORT>/fledge/proxy -d '{"service_name": "BucketStorage", "GET": {"/fledge/bucket/{uniqueID}": "/bucket/{uniqueID}"}}'
+             curl -sX POST http://localhost:<SVC_MGT_PORT>/fledge/proxy -d '{"service_name": "BucketStorage", "GET": {"/fledge/bucket/{uniqueID}": "/bucket/{uniqueID}"}, "PUT": {"/fledge/bucket/{uniqueID}": "/bucket/{uniqueID}"}, "DELETE": {"/fledge/bucket/{uniqueID}": "/bucket/{uniqueID}"}}'
+            curl -sX POST http://localhost:<SVC_MGT_PORT>/fledge/proxy -d '{"service_name": "Bucket #1", "DELETE": {"/fledge/bucket/([0-9][0-9]*)$": "/bucket/([0-9][0-9]*)$"}, "GET": {"/fledge/bucket/([0-9][0-9]*)$": "/bucket/([0-9][0-9]*)$"}, "POST": {"/fledge/bucket": "/bucket"}, "PUT": {"/fledge/bucket/([0-9][0-9]*)$": "/bucket/([0-9][0-9]*)$", "/fledge/bucket/match": "/bucket/match"}}'
+   """
+    data = await request.json()
+    svc_name = data.get('service_name', None)
```
MarkRiddoch, line 41: I think the spec said "service" rather than "service_name", but I have changed the code on the other side to match this so just leave as is.

MarkRiddoch (changes requested):

python/fledge/services/core/proxy.py
ashish-jabble: Sure, Initially it was **service_name** in doc and  having lowercase http method names given in payload.
Yesterday the payload was updated to service along with uppercase http method names (which I had asked as query in doc itself)

python/fledge/services/core/api/proxy.py
ashish-jabble: This was added only to get the info for a service (like service port, address, bearer token) so that Admin API -> forward those calls to microservice API.

Are we going to pass service name or any identifier in request header for public API exposed endpoints?

For example:  To achieve below we require identifier related to microservice

curl -sX GET localhost:8081/foglamp/bucket/1   => GET <MICROSVC_PROTOCOL>://<MICROSVC_ADDRESS>:<MICROSVC_PORT>/bucket/1

OR Am I missing?

MarkRiddoch: We pass the service name when we setup the proxy, that should be enough to get the information you need I believe

MarkRiddoch (dismissed):

MarkRiddoch (approved):

PR #619, opened 2022-04-07
FOGL-6337: Adding some debug logs around import_array call

AmandeepArora:

I added a small change to give better information about the “numpy.core.multiarray failed to import“ error. So it is not really a fix per se but it did help the situation and issue is not seen by me and Yash anymore. 

Also we added “python3-numpy” to list of run-time dependencies in fledge-pkg’s CONTROL file. This is in a separate PR that Yash would send out for review.


MarkRiddoch (approved):

YashTatkondawar:
Opened a PR for packaging changes - https://github.com/fledge-iot/fledge-pkg/pull/46

PR #626, opened 2022-04-15
FOGL-6358: fix memory leak in python north plugin send

pintomax:
FOGL-6358: fix memory leak in python north plugin send


AmandeepArora (approved):

YashTatkondawar (approved):
Tested fledge1 http-north sending data to fledge2, and I see no memory growth and also no fatals on adding http-north service.

PR #631, opened 2022-04-21
FOGL-5665: Memory leak fix

pintomax:
FOGL-5665: Memory leak fix


MarkRiddoch (approved):
Looks good, thank you.

PR #647, opened 2022-05-06
Option to add north as service or task in lab system tests

YashTatkondawar:


praveen-garg (approved):

tests/system/lab/test
```diff
@@ -170,6 +219,7 @@ setup_north_pi_egress () {
 if [[  ${VERIFY_EGRESS_TO_PI} == 1 ]]
   then
      setup_north_pi_egress
+     python3 verify_clean_pi.py --pi-admin=${PI_USER} --pi-passwd=${PI_PASSWORD} --pi-host=${PI_IP} --pi-port=443 --pi-db=foglamp --asset-name=sinusoid
```
praveen-garg, line 222: should we have sys exit instead of (or along with assert) in python def? as assert will not work if cleanup failed. 

Also can we add some logging for python script execution, or we are able to see print statements in console log for each run?

praveen-garg (approved):

PR #650, opened 2022-05-11
FOGL-6408. Added image height and image width parameters inside configuration of test card plugin. 

deepanshu-yadav:
Hi @MarkRiddoch, 

I have changed the test card plugin a lit bit for making height and width of image configurable. 

This is the resultant image of 256X256 24 bit depth from the plugin. 

image_testcard1_testcard_2022-05-11 10:09:14 339254


MarkRiddoch (commented):
You will get some strange outputs with this as the 256 by 256 was chosen such that the loops that produced the colour band would go form 0 to 256 for each colour. You may want to check that for widths and height both less than 256 and greater than 256.

Also please fix DCO, all checkins should be signed. And also unit test failures.

MarkRiddoch (approved):

AmandeepArora (approved):

PR #653, opened 2022-05-12
FOGL-6407: can't free borrowed references; 

AmandeepArora:
Checked with valgrind also - no memory leak related to readings.

This also fixes the crash seen with 3 instances of csv-writer in a single south service instance.


MarkRiddoch (approved):

deepanshu-yadav (approved):
Solves the error in FOGL-6406 and also error in FOGL-6367.

PR #655, opened 2022-05-17
FOGL-6321: Check reconfig of north tasks

YashTatkondawar:

### DIFF
--- a/tests/system/python/packages/test_pi_webapi.py
+++ b/tests/system/python/packages/test_pi_webapi.py
@@ -26,7 +26,7 @@ TEMPLATE_NAME = "template.json"
 ASSET = "FOGL-2964-e2e-CoAP"
 DATAPOINT = "sensor"
 DATAPOINT_VALUE = 20
-north_service_name = "NorthReadingsToPI_WebAPI"
+NORTH_TASK_NAME = "NorthReadingsToPI_WebAPI"
 
 # This  gives the path of directory where fledge is cloned. test_file < packages < python < system < tests < ROOT
 PROJECT_ROOT = Path(__file__).parent.parent.parent.parent.parent
@@ -74,7 +74,7 @@ def verify_statistics_map(fledge_url, skip_verify_north_interface):
     assert 1 <= actual_stats_map['READINGS']
     if not skip_verify_north_interface:
         assert 1 <= actual_stats_map['Readings Sent']
-        assert 1 <= actual_stats_map[north_service_name]
+        assert 1 <= actual_stats_map[NORTH_TASK_NAME]
 
 
 def verify_asset(fledge_url):
@@ -194,10 +194,6 @@ class TestPackagesCoAP_PI_WebAPI:
             skip_verify_north_interface: Flag for assertion of data using PI web API
             Assertions:
                 on endpoint GET /fledge/ping
-                on endpoint GET /fledge/south
-                on endpoint GET /fledge/north
-                on endpoint GET /fledge/filter
-                on endpoint GET /fledge/service
                 on endpoint GET /fledge/statistics
                 on endpoint GET /fledge/asset
                 on endpoint GET /fledge/track"""
@@ -216,7 +212,7 @@ class TestPackagesCoAP_PI_WebAPI:
 
         # Good reconfiguration to check data is sent
         data = {"SendFullStructure": "false"}
-        put_url = "/fledge/category/{}".format(north_service_name)
+        put_url = "/fledge/category/{}".format(NORTH_TASK_NAME)
         resp = utils.put_request(fledge_url, urllib.parse.quote(put_url), data)
         assert "false" == resp["SendFullStructure"]["value"]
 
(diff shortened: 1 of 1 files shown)


MarkRiddoch (approved):

tests/system/python/packages/test_pi_webapi.py
```diff
@@ -15,50 +15,88 @@
...
 import utils
-
+from pathlib import Path
+import urllib.parse
 
 TEMPLATE_NAME = "template.json"
 ASSET = "FOGL-2964-e2e-CoAP"
 DATAPOINT = "sensor"
 DATAPOINT_VALUE = 20
+north_service_name = "NorthReadingsToPI_WebAPI"
```
praveen-garg, line 29: service or task?

NORTH_TASK_NAME

tests/system/python/packages/test_pi_webapi.py
```diff
@@ -15,50 +15,88 @@
...
-def get_statistics_map(fledge_url):
-    _connection = http.client.HTTPConnection(fledge_url)
-    _connection.request("GET", '/fledge/statistics')
-    r = _connection.getresponse()
-    assert 200 == r.status
-    r = r.read().decode()
-    jdoc = json.loads(r)
-    return utils.serialize_stats_map(jdoc)
+# This  gives the path of directory where fledge is cloned. test_file < packages < python < system < tests < ROOT
+PROJECT_ROOT = Path(__file__).parent.parent.parent.parent.parent
```
praveen-garg, line 32: is there not a better way to do it?

tests/system/python/packages/test_pi_webapi.py
```diff
@@ -15,50 +15,88 @@
...
-    if event:
-        uri += '?event={}'.format(event)
-    _connection.request("GET", uri)
-    r = _connection.getresponse()
-    assert 200 == r.status
-    r = r.read().decode()
-    jdoc = json.loads(r)
-    return jdoc
+@pytest.fixture
+def reset_fledge(wait_time):
```
praveen-garg, line 37: don't we have this in conftest?

tests/system/python/packages/test_pi_webapi.py
```diff
@@ -128,58 +168,99 @@ def test_end_to_end(self, start_south_north, read_data_from_pi_web_api, fledge_u
...
-        conn.request("GET", '/fledge/asset/{}'.format(asset_name))
-        r = conn.getresponse()
-        assert 200 == r.status
-        r = r.read().decode()
-        retval = json.loads(r)
-        assert {DATAPOINT: DATAPOINT_VALUE} == retval[0]["reading"]
+            assert old_ping_result['dataSent'] < new_ping_result['dataSent']
+
+        # Bad reconfiguration to check data is not sent
+        data = {"PIWebAPIUserId": "Admin"}
```
praveen-garg, line 240: I suggest to replace `Admin` with `foo` or `Inv@lidRandomUserID`

tests/system/python/packages/test_pi_webapi.py
```diff
@@ -128,58 +168,99 @@ def test_end_to_end(self, start_south_north, read_data_from_pi_web_api, fledge_u
...
+        put_url = "/fledge/category/{}".format(north_service_name)
+        resp = utils.put_request(fledge_url, urllib.parse.quote(put_url), data)
+        assert "Admin" == resp["PIWebAPIUserId"]["value"]
+
+        old_ping_result = verify_ping(fledge_url, skip_verify_north_interface, wait_time, retries)
+
+        # Wait for the OMF schedule to run and then send the new data.
+        # Otherwise, it sends the data with old config.
```
praveen-garg, line 248: not clear what do you mean `Otherwise, it sends the data`? as you have configured with bad value it will not be sent anyway for this run.

I understand we need the next task iteration schedule to run hence the wait (default 15s hence 15s+ wait)?

tests/system/python/packages/test_pi_webapi.py
```diff
@@ -128,58 +168,99 @@ def test_end_to_end(self, start_south_north, read_data_from_pi_web_api, fledge_u
...
+            read_data_from_pi_web_api: Fixture to read data from PI web API
+            skip_verify_north_interface: Flag for assertion of data using PI web API
+            Assertions:
+                on endpoint GET /fledge/ping
+                on endpoint GET /fledge/south
+                on endpoint GET /fledge/north
+                on endpoint GET /fledge/filter
+                on endpoint GET /fledge/service
+                on endpoint GET /fledge/statistics
+                on endpoint GET /fledge/asset
```
praveen-garg, line 202: Which filter, service? asset for south?

tests/system/python/packages/test_pi_webapi.py
YashTatkondawar: The reset fixture in conftest is source code based and as this is a package-based test, it could not be used.

tests/system/python/packages/test_pi_webapi.py
praveen-garg: that sounds like a bug to me. if Re-config is successful, the immediate next task schedule should respect that. May be scheduler has already queued.

PR #662, opened 2022-05-27
FOGL-6464 Update storage client for new schema API

MarkRiddoch:

NOte, this can not be merged until a number of other Jira's are complete FOGL-6451 and FOGL-6384

### DIFF
--- a/C/common/include/storage_client.h
+++ b/C/common/include/storage_client.h
@@ -44,6 +44,14 @@ class StorageClient {
 		~StorageClient();
 		ResultSet	*queryTable(const std::string& tablename, const Query& query);
 		ReadingSet	*queryTableToReadings(const std::string& tableName, const Query& query);
+		int 		insertTable(const std::string& schema, const std::string& tableName, const InsertValues& values);
+		int		updateTable(const std::string& schema, const std::string& tableName, const InsertValues& values, const Where& where);
+		int		updateTable(const std::string& schema, const std::string& tableName, const JSONProperties& json, const Where& where);
+		int		updateTable(const std::string& schema, const std::string& tableName, const InsertValues& values, const JSONProperties& json, const Where& where);
+		int		updateTable(const std::string& schema, const std::string& tableName, const ExpressionValues& values, const Where& where);
+		int		updateTable(const std::string& schema, const std::string& tableName, std::vector<std::pair<ExpressionValues *, Where *>>& updates);
+		int		updateTable(const std::string& schema, const std::string& tableName, const InsertValues& values, const ExpressionValues& expressoins, const Where& where);
+		int		deleteTable(const std::string& schema, const std::string& tableName, const Query& query);
 		int 		insertTable(const std::string& tableName, const InsertValues& values);
 		int		updateTable(const std::string& tableName, const InsertValues& values, const Where& where);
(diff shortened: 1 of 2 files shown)


C/common/storage_client.cpp
```diff
@@ -484,13 +484,26 @@ ReadingSet* StorageClient::queryTableToReadings(const std::string& tableName,
...
+ */
+int StorageClient::insertTable(const string& schema, const string& tableName, const InsertValues& values)
 {
 	try {
 		ostringstream convert;
 
 		convert << values.toJSON();
 		char url[128];
-		snprintf(url, sizeof(url), "/storage/table/%s", tableName.c_str());
+		snprintf(url, sizeof(url), "/storage/schema/%s/table/%s", schema.c_str(), tableName.c_str());
```
Innovative-ashwin, line 506: We can add a debug log here printing the "url" and message that insertion is done for particular schema and table, similar for other functions

C/common/storage_client.cpp
```diff
@@ -484,13 +484,26 @@ ReadingSet* StorageClient::queryTableToReadings(const std::string& tableName,
...
+ * @return int		The number of rows inserted
+ */
+int StorageClient::insertTable(const string& schema, const string& tableName, const InsertValues& values)
 {
 	try {
 		ostringstream convert;
 
 		convert << values.toJSON();
 		char url[128];
-		snprintf(url, sizeof(url), "/storage/table/%s", tableName.c_str());
```
Innovative-ashwin, line 493: Are we sure that no other client ( python client may be )  will ever try to post to  /storage/table/  and will always be posting to /storage/schema<schema_name><table><tablename> , if that is sure then there might  be changes in python client as well, storage service needs to be updated in a way so  as to support /storage/table url or not?

C/common/include/storage_client.h
```diff
@@ -44,6 +46,14 @@ class StorageClient {
 		~StorageClient();
 		ResultSet	*queryTable(const std::string& tablename, const Query& query);
```
Innovative-ashwin, line 47: Why queryTable is not changing? existing queryTable sends a PUT to URL /storage/table , in case , this URL is not going to be supported and replaced with /storage/schema/.../table/... , then this function also needs to be modified

C/common/storage_client.cpp
MarkRiddoch: > Are we sure that no other client ( python client may be ) will ever try to post to /storage/table/ and will always be posting to /storage/schema<schema_name>
> 
The is the C++ client, it doesn't matter what the Postgres client does to these code changes.



When we need the python code to use the new schema we will have to update the Python client.
> , if that is sure then there might be changes in python client as well, storage service needs to be updated in a way so as to support /storage/table url or not?

Yes, the service must support both URLs for backward compatibility. 

> What I mean to ask is that does /storage/table URL need to be supported by the storage service , if not then all the clients posting to /storage/table will have to be modified to past to new URL ( /storage/schema/.../table/...)

C/common/storage_client.cpp
MarkRiddoch: Not inclined to add this and we will end up swamping the logs

C/common/include/storage_client.h
MarkRiddoch: This was an oversight, thank you

PR #672, opened 2022-06-08
FOGL-5195: Add support for control in Python south plugins

AmandeepArora:
Also added support for plugin mode to have multiple flags set (separated by pipe symbol '|') e.g. "poll|control" in python plugins.

### DIFF
--- a/C/services/common-plugin-interfaces/python/include/python_plugin_common_interface.h
+++ b/C/services/common-plugin-interfaces/python/include/python_plugin_common_interface.h
@@ -357,11 +357,25 @@ static PLUGIN_INFORMATION *Py2C_PluginInfo(PyObject* pyRetVal)
 		}
 		else if(!strcmp(ckey, "mode"))
 		{
+			stringstream ss(valStr); 
+			string s;
+
 			info->options = 0;
-			if (!strcmp(valStr, "async"))
+			
+			// Tokenizing w.r.t. pipe '|'
+			while(getline(ss, s, '|'))
 			{
-				info->options |= SP_ASYNC;
+				Logger::getLogger()->debug("%s: mode: Found token %s", __FUNCTION__, s.c_str());
+				if (s.compare("async")==0)
+				{
+					info->options |= SP_ASYNC;
+				}
+				else if (s.compare("control")==0)
+				{
+					info->options |= SP_CONTROL;
+				}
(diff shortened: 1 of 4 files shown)


AmandeepArora:
While testing, I faced the issue as mentioned in FOGL-6506. Tested plugin_operation path using a temporary hack in code.

C/services/common-plugin-interfaces/python/include/python_plugin_common_interface.h
```diff
@@ -357,11 +357,26 @@ static PLUGIN_INFORMATION *Py2C_PluginInfo(PyObject* pyRetVal)
...
-				info->options |= SP_ASYNC;
+				Logger::getLogger()->debug("%s: mode: Found token %s", __FUNCTION__, s.c_str());
+				if (s.compare("async")==0)
+				{
+					info->options |= SP_ASYNC;
+				}
+				else if (s.compare("control")==0)
+				{
+					info->options |= SP_CONTROL;
+				}
```
MarkRiddoch, line 377: It would probably be a good idea to have an else branch that logs an unknown option.

C/services/south-plugin-interfaces/python/python_plugin_interface.cpp
```diff
@@ -145,6 +147,238 @@ void *PluginInterfaceInit(const char *pluginName, const char * pluginPathName)
...
+	// Fetch required method in loaded object
+	pFunc = PyObject_GetAttrString(it->second->m_module, "plugin_write");
+	if (!pFunc)
+	{
+		Logger::getLogger()->fatal("Cannot find method 'plugin_write' "
+					   "in loaded python module '%s'",
+					   it->second->m_name.c_str());
+	}
+
+	if (!pFunc || !PyCallable_Check(pFunc))
```
MarkRiddoch, line 208: If you return above you avoid getting the second long message the plunk is NULL. You will have to del with the GIL release of course.

MarkRiddoch (commented):
A couple of minor things, also I thought I heard you mention documentation updates in the scrum call, are they under a different PR?

AmandeepArora:
@MarkRiddoch : Review comments incorporated and documentation changes also done.

MarkRiddoch (approved):

PR #695, opened 2022-06-21
FOGL-6483: Changes to propagate Management client host/port to plugins in plugin_init

AmandeepArora:


C/services/south/south.cpp
```diff
@@ -704,7 +704,9 @@ bool SouthService::loadPlugin()
 			// Removes all the m_items already present in the category
 			m_config.removeItems();
 			m_config = m_mgtClient->getCategory(m_name);
-
+			m_config.addItem("mgmt_client_url_base", "Management client host and port",
```
MarkRiddoch, line 707: Longer term we should use the existing mechanism we have for C++ notification plugins everywhere, bu this sis fine for the time.

MarkRiddoch (approved):

C/services/south/south.cpp
AmandeepArora: Yes, sure.

PR #697, opened 2022-06-22
Fogl 6333

deepanshu-yadav:
Did a test to send readings to north under impaired network.
Currently no way to verify that there is no gap in readings sent to pi as extracting bulk amount of data needs exploration.

### DIFF
--- a/tests/system/python/conftest.py
+++ b/tests/system/python/conftest.py
@@ -720,7 +720,7 @@ def pytest_addoption(parser):
     parser.addoption("--rate_limit", action="store", type=int, default=10,
                      help="The limit in packet transfer rate in kbps")
 
-    parser.addoption("--south-service-wait-time", action="store", type=int, default="20",
+    parser.addoption("--delay", action="store", type=int, default="20",
                      help="The delay in packet transfer to induce in the network. "
                           "Given in milli seconds.")
 
--- a/tests/system/python/conftest.py
+++ b/tests/system/python/conftest.py
@@ -717,15 +717,12 @@ def pytest_addoption(parser):
                           " to keep on running "
                           "after switching off the south service.")
 
-    parser.addoption("--interface-for-impairment", action="store", default="eth0",
-                     help="The interface on which network impairment will be applied.")
-
-    parser.addoption("--rate-limit", action="store", type=int, default=None,
-                     help="The limit in packet transfer rate in kbps.")
-
-    parser.addoption("--packet-delay", action="store", type=int, default=None,
-                     help="The delay in packet transfer to induce in the network. "
-                          "Given in milliseconds.")
+    parser.add_argument('throttled-network-config', action='store', type=json.loads,
+                        help="Give config '{'rate_limit': '100',"
+                             "            'packet_delay': '50',"
+                             "            'interface': 'eth0'}' "
+                             "for causing a delay of 50 milliseconds "
+                             "and rate restriction of 100 kbps on interface eth0.")
 
     parser.addoption("--start-north-as-service", action="store", type=bool, default=True,
                      help="Whether start the north as a service.")
@@ -992,18 +989,8 @@ def north_catch_up_time(request):
 
 
 @pytest.fixture
-def interface_for_impairment(request):
-    return request.config.getoption("--interface-for-impairment")
-
-
-@pytest.fixture
-def rate_limit(request):
-    return request.config.getoption("--rate-limit")
-
-
-@pytest.fixture
-def packet_delay(request):
-    return request.config.getoption("--packet-delay")
+def throttled_network_config(request):
+    return request.config.getoption("--throttled-network-config")
 
 
 @pytest.fixture
--- a/tests/system/python/conftest.py
+++ b/tests/system/python/conftest.py
@@ -724,6 +724,9 @@ def pytest_addoption(parser):
                      help="The delay in packet transfer to induce in the network. "
                           "Given in milliseconds.")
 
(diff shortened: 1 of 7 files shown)


tests/system/python/conftest.py
```diff
@@ -951,3 +978,34 @@ def pytest_itemcollected(item):
...
+    return request.config.getoption("--rate-limit")
+
+
+@pytest.fixture
+def packet_delay(request):
+    return request.config.getoption("--packet-delay")
+
+
+@pytest.fixture
+def start_north_as_service(request):
```
praveen-garg, line 1010: not needed as a fixture, we can pass local scope var to call corresponding fixture

tests/system/python/packages/test_north_pi_webapi_nw_throttle.py
```diff
@@ -0,0 +1,272 @@
...
+    conn.request("PUT", '/fledge/schedule/disable', json.dumps({"schedule_name": sch_name}))
+    r = conn.getresponse()
+    assert 200 == r.status
+    r = r.read().decode()
+    jdoc = json.loads(r)
+    assert "scheduleId" in jdoc
+    return
+
+
+def enable_schedule(fledge_url, sch_name):
```
praveen-garg, line 119: We already have these as fixtures

tests/system/python/packages/test_north_pi_webapi_nw_throttle.py
```diff
@@ -0,0 +1,272 @@
...
+    add_south(south_plugin, None, fledge_url, config=_config,
+              service_name=SOUTH_SERVICE_NAME, installation_type='package')
+    if not start_north_as_service:
+        start_north_task_omf_web_api(fledge_url, pi_host, pi_port, pi_user=pi_admin, pi_pwd=pi_passwd,
+                                     start_task=False, taskname=NORTH_TASK_NAME)
+    else:
+        start_north_omf_as_a_service(fledge_url, pi_host, pi_port, pi_user=pi_admin, pi_pwd=pi_passwd,
+                                     start=False, service_name=NORTH_TASK_NAME)
+
+    add_filter("python35", None, "py35", {}, fledge_url, None, installation_type='package',
```
praveen-garg, line 199: Instead of adding flag and then adding filter + pipeline to north instance; can not we use add_filter as is and just upload the script?

tests/system/python/packages/test_north_pi_webapi_nw_throttle.py
```diff
@@ -0,0 +1,272 @@
...
+
+    data = {"name": "py35", "plugin": "python35", "filter_config": {"enable": "true"}}
+    utils.post_request(fledge_url, "/fledge/filter", data)
+
+    data = {"pipeline": ["py35"]}
+    put_url = "/fledge/filter/{}/pipeline?allow_duplicates=true&append_filter=true" \
+        .format(NORTH_TASK_NAME)
+    utils.put_request(fledge_url, urllib.parse.quote(put_url, safe='?,=,&,/'), data)
+
+    url = fledge_url + urllib.parse.quote('/fledge/category/{}_py35/script/upload'
```
praveen-garg, line 210: NORTH_TASK_NAME -> NORTH_INSTANCE_NAME (as it can be a service, and actually by default its a service)

deepanshu-yadav: Done.

tests/system/python/packages/test_north_pi_webapi_nw_throttle.py
```diff
@@ -0,0 +1,272 @@
...
+        time.sleep(north_catch_up_time)
+
+        # clear up all the distortions on this network.
+        reset_network(interface=interface_for_impairment)
+        verify_ping(fledge_url, north_catch_up_time)
+        # verify_asset(fledge_url)
+        # verify_statistics_map(fledge_url, skip_verify_north_interface)
+        # verify_asset_tracking_details(fledge_url, skip_verify_north_interface)
+        #
+        if not skip_verify_north_interface:
```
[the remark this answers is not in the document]
deepanshu-yadav: Done.

tests/system/python/packages/test_north_pi_webapi_nw_throttle.py
deepanshu-yadav: We should not start the north instance before applying filter. Otherwise we will not be able to insert reading  id from the very first reading itself.  Also I wanted to install filter only. Then we need to upload a script to the filter. This is a very specific thing we have to do. So we cannot utilize the existing fixture add_filter.

tests/system/python/conftest.py
deepanshu-yadav: But why not keep it configurable.

tests/system/python/conftest.py
```diff
@@ -703,6 +706,30 @@ def pytest_addoption(parser):
...
+                          " to keep on running "
+                          "after switching off the south service.")
+
+    parser.addoption("--interface-for-impairment", action="store", default="eth0",
+                     help="The interface on which network impairment will be applied.")
+
+    parser.addoption("--rate-limit", action="store", type=int, default=None,
+                     help="The limit in packet transfer rate in kbps.")
+
+    parser.addoption("--packet-delay", action="store", type=int, default=None,
```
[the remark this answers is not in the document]
deepanshu-yadav: Done Updated REAME as well.

praveen-garg (approved):

PR #699, opened 2022-06-23
FOGL-6530 Proxy API mutlipart form support and other cleanup fixes

ashish-jabble:
Also ran custom job with the given change and all system tests are passing. As such no regression found.


MarkRiddoch (approved):
Thanks, works well

PR #713, opened 2022-07-05
FOGL-6587: schema support for sqlitelb plugin

pintomax:
FOGL-6587: schema support for sqlitelb plugin


pintomax:
Bucket service now starts with "sqlitelb" plugin

PR #717, opened 2022-07-06
FOGL-6611: addition of fledge.acl_usage table

pintomax:
FOGL-6611: addition of fledge.acl_usage table


MarkRiddoch (changes requested):
What about upgrade/downgrade scripts and changing the schema version?

PR #751, opened 2022-07-26
FOGL-6665: Improve error log performance

AmandeepArora:


scripts/common/get_logs.sh
```diff
@@ -0,0 +1,125 @@
...
+    -logfile) logfile="$2"; shift 2;;
+  esac
+done
+
+sum=$(($offset + $limit))
+echo "" >&2
+echo "****************************************************************************************" >&2
+echo "************************************* START ********************************************" >&2
+
+if [ -f /tmp/fledge_syslog_script_runs ]; then
```
MarkRiddoch, line 37: There is a danger here if we have two users logged in and looking at syslog

MarkRiddoch:
What sort of improvements are you seeing? Is writing the file slowing in down very much? Thinking of an SD card I can see the storage writes might be an issue.

scripts/common/get_logs.sh
AmandeepArora: Since these requests go thru' fledge core, would the requests from 2 browsers be processed sequentially by core or they could interleave. Even if they interleave, the script_runs count would increase doubly fast, which shouldn't be a problem. Or do you think a worse race condition could occur in this case?

AmandeepArora:
> What sort of improvements are you seeing? Is writing the file slowing in down very much? Thinking of an SD card I can see the storage writes might be an issue.

On my well spec'ed machine with NVMe SSD and around 900K syslog lines (175 MB in size), it takes 40-50 msec to get syslog entries in most cases. And if the complete syslog file has to be scanned, it takes 500-700 ms.

On a RPi 4 with 1 GB RAM and 16 GB high speed SD card, and 695K lines in syslog (110 MB in size), it takes 110-170ms to get syslog entries in most cases. And if the complete syslog file has to be scanned, it takes 850-950 ms.

And as expected, the first time a particular "service+log level" filter is applied for fetching logs, then it takes more time while trying last 2000, 4000, lines etc. from syslog file. Otherwise we are able to avoid grepping of whole syslog file every time a syslog query is handled. 

Did you mean also the writing of small temp files by this get_logs.sh bash script causing further slowdown? I intended to keep these temp files in some tmpfs partition (RAM/swap) rather than some disk. So may be I can move these files to '/run' probably?

python/fledge/services/core/api/support.py
```diff
@@ -28,6 +30,10 @@
 if any(x in platform.platform() for x in ['centos', 'redhat']):
     _SYSLOG_FILE = '/var/log/messages'
 
+# FLEDGE_ROOT env variable
+_FLEDGE_ROOT = os.getenv("FLEDGE_ROOT", default='/usr/local/fledge')
+_SCRIPTS_DIR = os.path.expanduser(_FLEDGE_ROOT + '/scripts')
```
ashish-jabble, line 35: Use common utility
```
from fledge.common.common import _FLEDGE_ROOT
```

```
_SCRIPTS_DIR = "{}/scripts".format(_FLEDGE_ROOT)
```

python/fledge/services/core/api/support.py
```diff
@@ -160,6 +166,9 @@ async def get_syslog_entries(request):
         template = __GET_SYSLOG_CMD_TEMPLATE
         lines = __GET_SYSLOG_TOTAL_MATCHED_LINES
         non_total_template = __GET_SYSLOG_TEMPLATE_WITH_NON_TOTALS
```
ashish-jabble, line 168: Local variable 'non_total_template' value is not used anymore. Remove it from here and its reference.

python/fledge/services/core/api/support.py
```diff
@@ -160,6 +166,9 @@ async def get_syslog_entries(request):
...
+
+        levels = ""
+        level = "debug"
```
ashish-jabble, line 171: By default level should be ALL.
And levels = "(INFO|WARNING|ERROR|FATAL|DEBUG)"

python/fledge/services/core/api/support.py
```diff
@@ -191,9 +204,17 @@ async def get_syslog_entries(request):
             response['count'] = total_lines
             cmd = template.format(valid_source[source], _SYSLOG_FILE, total_lines - offset, limit)
         else:
-            cmd = non_total_template.format(valid_source[source], _SYSLOG_FILE, offset, limit)
-        a = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE).stdout.readlines()
+            scriptPath = os.path.join(_SCRIPTS_DIR, "common", "get_logs.sh")
+            # cmd = non_total_template.format(valid_source[source], _SYSLOG_FILE, offset, limit)
+            pattern = '({})\[.*\].*{}:'.format(valid_source[source], levels)
+            cmd = '{} -offset {} -limit {} -pattern \'{}\' -logfile {} -source {} -level {}'.format(scriptPath, offset, limit, pattern, _SYSLOG_FILE, source, level)
+            _logger.info('********* non_totals=true: new shell command: {}'.format(cmd))
```
ashish-jabble, line 211: log should be with debug if really required

python/fledge/services/core/api/support.py
```diff
@@ -191,9 +204,17 @@ async def get_syslog_entries(request):
...
+
+        t1 = datetime.datetime.now()
+        a = subprocess.Popen([cmd], shell=True, stdout=subprocess.PIPE).stdout.readlines()
+        t2 = datetime.datetime.now()
         c = [b.decode() for b in a]  # Since "a" contains return value in bytes, convert it to string
```
ashish-jabble, line 216: Please rename t1, a, t2, c  variables to some more meaningful

python/fledge/services/core/api/support.py
```diff
@@ -191,9 +204,17 @@ async def get_syslog_entries(request):
...
+        _logger.info('********* Time taken for grep/tail/head subprocess: {} msec'.format((t2 - t1).total_seconds()*1000))
```
ashish-jabble, line 217: log should be with debug if really required

scripts/common/get_logs.sh
```diff
@@ -0,0 +1,125 @@
+#!/bin/bash
+
+__author__="Amandeep Singh Arora"
+__version__="1.0"
+
+> /tmp/fledge_syslog.log
+
+exec 2<&-
+exec 2<>/tmp/fledge_syslog.log
```
ashish-jabble, line 9: We should avoid the file naming with Product prefix.

scripts/common/get_logs.sh
```diff
@@ -0,0 +1,125 @@
...
+
+RECALC_AFTER_N_SCRIPT_RUNS=100
+NUM_LOGFILE_LINES_TO_CHECK_INITIALLY=2000
+
+offset=100
+limit=5
+pattern=""
+level="debug"
+logfile="/var/log/syslog"
+sourceApp="fledge"
```
ashish-jabble, line 19: It's better to add comments what these variables are used for.

scripts/common/get_logs.sh
```diff
@@ -0,0 +1,125 @@
...
+echo "script_runs=$script_runs" >&2
+if [[ $script_runs -gt ${RECALC_AFTER_N_SCRIPT_RUNS} ]]; then
+	echo "Resetting script_runs" >&2
+	rm /tmp/fledge_syslog_script_runs
+	script_runs=0
+fi
+echo -n "$script_runs" > /tmp/fledge_syslog_script_runs
+echo "offset=$offset, limit=$limit, sum=$sum, pattern=$pattern, sourceApp=$sourceApp, level=$level, script_runs=$script_runs" >&2
+
+factor=2
```
ashish-jabble, line 53: Also describe what factor is

ashish-jabble (commented):
Please add more comments in get_logs.sh.

python/fledge/services/core/api/support.py
AmandeepArora: I would like to retain it for now while this change is being tested by Ashwin and others. I can maybe comment it for now.

python/fledge/services/core/api/support.py
AmandeepArora: Do u mean "request.query['level']" would have value ALL in this case? ALL also means "debug and above", correct?

I am using level just as a string to put into /tmp/fledge_syslog_factor file. So it has no more meaning than that.

And I am using levels to compose grep pattern. And setting it to "(INFO|WARNING|ERROR|FATAL|DEBUG)" would just mean unnecessarily complicating grep's work with no benefit.

python/fledge/services/core/api/support.py
AmandeepArora: yes, kept it info for testing round.

python/fledge/services/core/api/support.py
AmandeepArora: t1 and t2 are timestamps before and after shell script execution. I thought it would be pretty obvious.

Will change 'a' to 'rv' (return value) and 'c' to 'rv_str'?

AmandeepArora: 'a' and 'c' were already there before this PR.

python/fledge/services/core/api/support.py
AmandeepArora: ok

scripts/common/get_logs.sh
AmandeepArora: Changed "fledge_" to "fl_" throughout the file.

scripts/common/get_logs.sh
AmandeepArora: Their meaning is pretty obvious if you look around a bit in the script.

scripts/common/get_logs.sh
AmandeepArora: Added comment.

AmandeepArora:
> Please add more comments in get_logs.sh.

Added.

python/fledge/services/core/api/support.py
AmandeepArora: Unit test cases fail when I change 'a' to 'rv', so reverting.

python/fledge/services/core/api/support.py
AmandeepArora: done

python/fledge/services/core/api/support.py
```diff
@@ -208,14 +204,14 @@ async def get_syslog_entries(request):
...
+            _logger.debug('********* non_totals=true: new shell command: {}'.format(cmd))
 
         t1 = datetime.datetime.now()
-        a = subprocess.Popen([cmd], shell=True, stdout=subprocess.PIPE).stdout.readlines()
+        rv = subprocess.Popen([cmd], shell=True, stdout=subprocess.PIPE).stdout.readlines()
         t2 = datetime.datetime.now()
-        c = [b.decode() for b in a]  # Since "a" contains return value in bytes, convert it to string
-        _logger.info('********* Time taken for grep/tail/head subprocess: {} msec'.format((t2 - t1).total_seconds()*1000))
-        response['logs'] = c
+        rv_str = [b.decode() for b in a]  # Since "a" contains return value in bytes, convert it to string
```
ashish-jabble: Here is the reason for the unit test failure on this commit is due to incomplete variable refactoring. See code snippet
```
[b.decode() for b in a] 
```
reference of `a` is still there. It should be with `rv`

AmandeepArora: yeah, thanks for pointing out.

python/fledge/services/core/api/support.py
ashish-jabble: Okay

scripts/common/get_logs.sh
MarkRiddoch: We can run multiple threads to make API requests in theory. It may be that Python GIL actively prevents this, but possibly not as I belie these commands are run using asyncio mechanisms.

MarkRiddoch:
> > What sort of improvements are you seeing? Is writing the file slowing in down very much? Thinking of an SD card I can see the storage writes might be an issue.
> 
> On my well spec'ed machine with NVMe SSD and around 900K syslog lines (175 MB in size), it takes 40-50 msec to get syslog entries in most cases. And if the complete syslog file has to be scanned, it takes 500-700 ms.
> 
> On a RPi 4 with 1 GB RAM and 16 GB high speed SD card, and 695K lines in syslog (110 MB in size), it takes 110-170ms to get syslog entries in most cases. And if the complete syslog file has to be scanned, it takes 850-950 ms.
> 
> And as expected, the first time a particular "service+log level" filter is applied for fetching logs, then it takes more time while trying last 2000, 4000, lines etc. from syslog file. Otherwise we are able to avoid grepping of whole syslog file every time a syslog query is handled.
> 
> Did you mean also the writing of small temp files by this get_logs.sh bash script causing further slowdown? I intended to keep these temp files in some tmpfs partition (RAM/swap) rather than some disk. So may be I can move these files to '/run' probably?

These look like good improvements. The problem with putting them in special location can be two fold, more complex setup and permissions. We also need to think about running under containers on both Linux and in devices like the Cisco routers we run in at some sites. Keeping it simple is probably best.

MarkRiddoch (approved):

MarkRiddoch:
What is stopping us checking this in?

AmandeepArora:
I am waiting for @ashwinscale to approve these changes.

ashwinscale (approved):
lgtm

PR #754, opened 2022-07-27
FOGL-6724 Fix import of persisted plugin data

MarkRiddoch:


python/fledge/services/core/api/plugins/data.py
```diff
@@ -79,8 +79,12 @@ async def add(request: web.Request) -> web.Response:
             msg = "{} key already exist.".format(key)
             return web.HTTPConflict(reason=msg, body=json.dumps({"message": msg}))
         data = await request.json()
-        payload = PayloadBuilder().INSERT(key=key, data=data)
-        await storage_client.insert_into_tbl('plugin_data', payload.payload())
+        data = data["data"]
```
ashish-jabble, line 82: So, we do not need to update the curl Example as it is already with **data** KV pair.

ashish-jabble (approved):
Dry run, LGTM!

PR #756, opened 2022-07-28
FOGL-6742: isnull and not null added to sqlitelb and postgres

pintomax:
FOGL-6742: isnull and not null added to sqlitelb and postgres

### DIFF
--- a/C/plugins/storage/postgres/connection.cpp
+++ b/C/plugins/storage/postgres/connection.cpp
@@ -2752,7 +2752,9 @@ bool Connection::jsonModifiers(const Value& payload, SQLBuffer& sql)
  * Convert a JSON where clause into a PostresSQL where clause
  *
  */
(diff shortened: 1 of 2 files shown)


pintomax:
curl -X PUT http://localhost:46095/storage/table/asset_tracker/query -d '
{
	"where" : {
		"column" : "ts",
		"condition" : "notnull",
		"and" : {
			"column" : "service",
			"condition" : "=",
			"value" : "SIN3"
		}
	}
}' | jq

curl -X PUT http://localhost:46095/storage/table/asset_tracker/query -d '
{
	"where" : {
		"column" : "ts",
		"condition" : "isnull",
		"and" : {
			"column" : "service",
			"condition" : "=",
			"value" : "SIN2"
		}
	}
}' | jq

example calls made with a new row into asset_tracker with ts = NULL, Postgres and sqlitelb

ashish-jabble (commented):
Add some unit tests. So that we are sure there are no regression around the new changes.

PR #761, opened 2022-08-01
FOGL-6721: un-deprecate asset tracking records

pintomax:
FOGL-6721: initial implementation of un-deprecate assets

deprecated assets are un-deprecated only at service start.

Current change is for "sqlite" storage plugin


MarkRiddoch (approved):

ashwinscale (approved):
LGTM. Review by test.

PR #765, opened 2022-08-03
FOGL-6693

ashish-jabble:
**Fixed items**
- [x] Add new endpoint to get persistent plugins
```
$ curl -sX GET "http://localhost:8081/fledge/service/{service_name}/persist"
{"persistent": ["OMF"]}
```
- [x] Forbidden handling added in POST and DELETE when service is running
- [x] HTTP status code fixes in POST when malformed payload.

### DIFF
--- a/python/fledge/plugins/common/utils.py
+++ b/python/fledge/plugins/common/utils.py
@@ -8,13 +8,16 @@
 
 import datetime
 
-__author__ = "Amarendra Kumar Sinha"
+from fledge.services.core.api import utils as api_utils
+
+__author__ = "Amarendra Kumar Sinha, Ashish Jabble"
 __copyright__ = "Copyright (c) 2017 OSIsoft, LLC"
 __license__ = "Apache 2.0"
 __version__ = "${VERSION}"
 
 DEPRECATED_BIT_POSITION = 7
 DEPRECATED_BIT_MASK_VALUE = 1 << DEPRECATED_BIT_POSITION
+PERSIST_DATA_BIT_POSITION = 3
 
 
 def get_diff(old, new):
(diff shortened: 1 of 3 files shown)


MarkRiddoch (approved):

python/fledge/plugins/common/utils.py
```diff
@@ -51,3 +54,22 @@ def bit_at_given_position_set_or_unset(n, k):
...
+
+def get_persist_plugins():
+    """ Get a list of south, north, filter types plugins that can persist data for a service
+    :return: list - plugins
+    :example: ["OMF"]
+    """
+    plugin_list = []
+    supported_persist_dirs = ["south", "north", "filter"]
+    for plugin_type in supported_persist_dirs:
+        libs = api_utils.find_c_plugin_libs(plugin_type)
```
praveen-garg, line 67: does this mean we cannot persist data for python based plugins?

python/fledge/services/core/api/plugins/data.py
```diff
@@ -70,6 +87,10 @@ async def add(request: web.Request) -> web.Response:
     try:
         service = request.match_info.get('service_name', None)
         plugin = request.match_info.get('plugin_name', None)
+        svc_records = ServiceRegistry.all()
+        for service_record in svc_records:
+            if service_record._name == service and int(service_record._status) == 1:
+                raise web.HTTPForbidden(reason=FORBIDDEN_MSG)
```
praveen-garg, line 93: Better check schedule to be disabled; OR may be Unresponsive check as well.

python/fledge/services/core/api/plugins/data.py
```diff
@@ -111,6 +132,10 @@ async def delete(request: web.Request) -> web.Response:
         service = request.match_info.get('service_name', None)
         service = urllib.parse.unquote(service) if service is not None else None
         plugin = request.match_info.get('plugin_name', None)
+        svc_records = ServiceRegistry.all()
+        for service_record in svc_records:
+            if service_record._name == service and int(service_record._status) == 1:
+                raise web.HTTPForbidden(reason=FORBIDDEN_MSG)
```
praveen-garg, line 138: as above

PR #772, opened 2022-08-04
FOGL-5954 Add definition of all the flags values for C++ plugins

MarkRiddoch:


ashish-jabble (commented):
LGTM!

It's better to add Python based `mode` brief description as well. See `mode` is missing
https://github.com/fledge-iot/fledge/blob/develop/docs/plugin_developers_guide/02_writing_plugins.rst#plugin-information.

Also confirm if it requires to check-in under 2.0.0RC after merge.

MarkRiddoch:
> LGTM!
> 
> It's better to add Python based `mode` brief description as well. See `mode` is missing https://github.com/fledge-iot/fledge/blob/develop/docs/plugin_developers_guide/02_writing_plugins.rst#plugin-information.
> 
> Also confirm if it requires to check-in under 2.0.0RC after merge.

Agreed, it was going to be my next addition to this Jira.

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -57,14 +57,15 @@ A typical Python implementation of this would simply return a fixed dictionary o
 
 These are the properties returned by the JSON document:
 
-- **Name** - A textual name that will be used for reporting purposes for this plugin.
-- **Version** - This property allows the version of the plugin to be communicated to the plugin loader. This is used for reporting purposes only and has no effect on the way Fledge interacts with the plugin.
-- **Type** - The type of the plugin, used by the plugin loader to determine if the plugin is being used correctly. The type is a simple string and may be South, North, Storage, Filter, Rule or Delivery.
+- **name** - A textual name that will be used for reporting purposes for this plugin.
+- **version** - This property allows the version of the plugin to be communicated to the plugin loader. This is used for reporting purposes only and has no effect on the way Fledge interacts with the plugin.
+- **mode** - A set of options the define how the plugin operations, these are separated from each other using the | symbol.
```
ashish-jabble, line 62: > the define

that/they define?

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -57,14 +57,15 @@ A typical Python implementation of this would simply return a fixed dictionary o
...
+- **type** - The type of the plugin, used by the plugin loader to determine if the plugin is being used correctly. The type is a simple string and may be South, North, Storage, Filter, Rule or Delivery.
```
ashish-jabble, line 63: lower case? (south, north, filter). Not sure about storage, rule, delivery types, Do we have support of these?

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -81,6 +82,27 @@ These are the properties returned by the JSON document:
...
++---------+---------------------------------------------------------------------------------------+
+| Mode    | Description                                                                           |
++=========+=======================================================================================+
+| poll    | The plugin is a polled plugin and *plugin_poll* will be called periodically to obtain |
+|         | new values.                                                                           |
++---------+---------------------------------------------------------------------------------------+
+| async   | The plugin is an asynchronous plugin, *plugin_poll* will not be called and the        |
+|         | plugin will be supplied with a callback function that it calls each time it has a     |
+|         | new value to pass to the Fledge system. The *plugin_register_ingest* entry point      |
+|         | will be called to register the callback with the plugin.                              |
```
ashish-jabble, line 97: 1) We can also add some info about **plugin_start**. It is called only once for a plugin and also have the responsibility to install the plugin code into python event handling for Asyncio.
2) Table will break when it comes t white-labelled as Fledge text is used inside. Better to use a list-table directive syntax and having defined widths and header rows

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -81,6 +82,27 @@ These are the properties returned by the JSON document:
...
++---------+---------------------------------------------------------------------------------------+
+| none    | This is equivalent to poll.                                                           |
++---------+---------------------------------------------------------------------------------------+
+| control | The plugin support a control flow from Fledge to the device the plugin is connected   |
+|         | to. The plugin must supply the control entry points *plugin_write* and                |
+|         | *plugin_operation*                                                                    |
```
ashish-jabble, line 103: Table will break when it comes t white-labelled as Fledge text is used inside.
Better to use a list-table directive syntax and having defined widths and header rows

ashish-jabble (changes requested):

docs/plugin_developers_guide/02_writing_plugins.rst
MarkRiddoch: We don't support storage, otherwise yes we do for the others.

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -81,6 +82,26 @@ These are the properties returned by the JSON document:
...
++---------+---------------------------------------------------------------------------------------+
+| Mode    | Description                                                                           |
++=========+=======================================================================================+
+| poll    | The plugin is a polled plugin and *plugin_poll* will be called periodically to obtain |
+|         | new values.                                                                           |
++---------+---------------------------------------------------------------------------------------+
+| async   | The plugin is an asynchronous plugin, *plugin_poll* will not be called and the        |
+|         | plugin will be supplied with a callback function that it calls each time it has a     |
+|         | new value to pass to the system. The *plugin_register_ingest* entry point will be     |
+|         | called to register the callback with the plugin.                                      |
```
ashish-jabble, line 97: *plugin_start* is also required for Python async plugins. We should add brief note about it.

ashish-jabble (approved):

PR #777, opened 2022-08-07
FOGL-6525 fix make for aarch64 aiohttp pip package and missing sqlite…

praveen-garg:
…3 command

pip3 upgrade is required for aarch64/arm64 env for aiohttp package to install.
make install after sqlite3_build_prepare should be enough to install sqlite3 as a command, We can investigate that later. 
This should fix the issues QA seen in dev env to build on aarch64 and  also missing sqlite3 command for creating backup & reset after make/make install.

### DIFF
--- a/requirements.sh
+++ b/requirements.sh
@@ -197,7 +197,6 @@ if [[ $YUM_PLATFORM = true ]]; then
 	service rsyslog start
 
 	sqlite3_build_prepare
-
 	# Attempts a second execution of make if the first fails
 	set +e
 	make
@@ -206,10 +205,11 @@ if [[ $YUM_PLATFORM = true ]]; then
 
 		set -e
 		make
+		# TODO: Use make install to install sqlite3 as a command
 	fi
 	cd $fledge_location
 	set -e
-
+	
 	# Upgrade curl if needed
 	curl_upgrade_evaluates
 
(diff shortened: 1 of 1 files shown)


praveen-garg:
close #774

praveen-garg:
https://github.com/fledge-iot/fledge-pkg/pull/59 is for packaging fix.

requirements.sh
```diff
@@ -248,18 +248,20 @@ elif apt --version 2>/dev/null; then
...
 	# for Kerberos authentication, avoid interactive questions
 	DEBIAN_FRONTEND=noninteractive apt install -yq krb5-user
 	DEBIAN_FRONTEND=noninteractive apt install -yq libcurl4-openssl-dev
 
-	apt install -y cpulimit
+	apt install -y cpulimit 
+	
 else
-	echo "Requirements cannot be automatically installed, please refer README.rst to install requirements manually"
-yum install -y rh-python36
```
ashish-jabble, line 264: python36 package does not require on yum?

requirements.sh
```diff
@@ -248,18 +248,20 @@ elif apt --version 2>/dev/null; then
 	apt install -y cmake g++ make build-essential autoconf automake uuid-dev
 	apt install -y libtool libboost-dev libboost-system-dev libboost-thread-dev libpq-dev libz-dev
 	apt install -y python-dev python3-dev python3-pip python3-numpy
+        pip3 install --upgrade pip
 
 	sqlite3_build_prepare
 	make
-
+        
+	apt install -y sqlite3 # make install after sqlite3_build_prepare should be enough to install sqlite3 as a command
```
ashish-jabble, line 256: So we require sqlite3 command in any case?. Only `libsqlite3-dev` is not required as we compile our custom sqlite dev package from https://github.com/dianomic/sqlite3-pkg

ashish-jabble (approved):

requirements.sh
praveen-garg: We should see make install OR make && make install install sqlite3; if that works and handle the cases for create back up and reset etc. We can rely on that and remove pkg dependency in dev env but still the prod. 

I assume libsqlite dev is required for sqlite3 to compile. No harm to keep

requirements.sh
praveen-garg: This is in else block which is meant for unsupported env. Don't why it was added in no yum /RedHat family block

PR #782, opened 2022-08-09
FOGL-6774 Configuration update handling for rest_api category and having authentication config item in case of non-admin user

ashish-jabble:
**Note**:
Existing system and unit health tests are okay with the given change.


mshariq-nerd (approved):

mshariq-nerd:
Approved. Looks fine in its scope, though I feel complete ADMIN API config response belongs to Admin. Only some part of it can be used by normal user. IMO for normal user only user required config should be returned.
cc/ @MarkRiddoch

MarkRiddoch (approved):
Agree, we should have a more general model for this. I think we may want to review the types of user and what they can do for the next release, post 2.0, but at least this fixes an obvious hole.

PR #787, opened 2022-08-16
Pi system tests fixes

deepanshu-yadav:
1. Made asset hierarchy unique for every system tests that sends data to PI using PI Web API. 
2. For system tests sending data to PI using connector relay only delete the PI point.
3. For asset having only single data point,  PI point is created with their asset name only. Handled them passing '' in data points list.   
4. Call clean PI system for every tests including lab tests. 
5. Made asset name in system tests unique so that unique data point is created. 
6. Made asset template unique as well by making the hierarchy list unique for every tests. 

We should now run them all @Mohit04tomar 
cc  @praveen-garg .


tests/system/common/clean_pi_system.py
```diff
@@ -54,8 +54,11 @@ def search_for_pi_point(host, admin, password, asset_name, data_point_name):
     except Exception:
         print("Could not get Points data.")
         return False, None
+    if data_point_name != '':
```
praveen-garg, line 57: Please add comments and Jira # reference as Known behaviour for single datapoint for an asset reading

tests/system/common/clean_pi_system.py
```diff
@@ -54,8 +54,11 @@ def search_for_pi_point(host, admin, password, asset_name, data_point_name):
...
+        name_to_search = asset_name + '.' + data_point_name
+    else:
+        name_to_search = asset_name
```
praveen-garg, line 60: we search for asset or datapoint?

praveen-garg (commented):
please add an explanation to each dp_list for '' appended.

deepanshu-yadav:
> please add an explanation to each dp_list for '' appended.

done

PR #789, opened 2022-08-18
FOGL-6665

AmandeepArora:
Added changes for configuring separate syslog files for fledge/foglamp with log rotation. Logs would be placed under ${FLEDGE_ROOT}/data/logs folder. File “debug.log” would contains fledge logs of DEBUG and higher levels, & similarly for info.log, warning.log and error.log files. So now extracting fledge (service) logs is quite fast, except when filtering debug logs and there being lots of debug logs.

Still pending/questions: 
* There is a helper script "scripts/extras/setup_logger.sh" to install these rsyslog & logrotate config for fledge/foglamp, need to run this as root or with sudo. This can be run in postinst script when installing packages, or manually, if required in dev environment. 
* Should be still retain fledge logs in system /var/log/syslog file also?
* Should we maintain older logic of grep/head/tail from /var/log/syslog file or always have these logs in "data/logs" folder?
* Need to check for non-debian environments.
* "data/logs" folder is already being used for something else (package logs?), need to resolve this.


MarkRiddoch (changes requested):
Why have we done this?

This is not what the Jira requested and has broken lots of other functionality. I explicitly said do not mix this up with the container thing Ashwin raised.

Support bundles will now be broken, documentation is broken, we are writing log data to the must critical of filestore locations, the place we keep our data.

It also does not resolve the original issue when Fledge is producing lots of lots it got very slow - it still will with this.

AmandeepArora:
Hi @MarkRiddoch, this is a separate branch (& PR) compared to my previous work (FOGL-6665 branch) which was against the original JIRA. These are experimental changes that Ashwin asked for, that we can evaluate and see if this improves the user experience for syslog filtering as provided thru' GUI.

We can have the logs at an alternate location rather than ${FLEGDE_ROOT}/data/logs.

Also this solution would avoid the performance penalty if the system log file significant number of logs from non-fledge tasks/services. 

This solution simplifies the grep portion of grep/head/tail command.

MarkRiddoch:
Closing this as Aman left many years ago and this was never fully finished.

PR #797, opened 2022-08-25
function call fixes in memory leak system test 

Mohit04tomar:
I have updated the **test_memcheck.sh** file. 
The  **adding_sinusoid** is replaced by **add_sinusoid**


praveen-garg (approved):

praveen-garg (dismissed):

ashish-jabble (dismissed):
echo statements can be better in test.

praveen-garg (approved):

PR #801, opened 2022-08-29
FOGL-6765: Add details on AFLocation hint behavior

RVerhoeff:
This is the page that has the most detailed description of the AFLocation hint. The [OMF Hint Filter webpage](https://github.com/fledge-iot/fledge-filter-omfhint/blob/develop/docs/index.rst) points to this one. There is a related [Pull Request #16](https://github.com/fledge-iot/fledge-filter-omfhint/pull/16) in the [fledge-filter-omfhint](https://github.com/fledge-iot/fledge-filter-omfhint) repository to add the AF Location hint because it was not there before.

Signed-off-by: Ray Verhoeff <ray@dianomic.com>


MarkRiddoch (dismissed):

deepanshu-yadav:
I don't know whether it is related to this . But we can also specify the convention we are using to create PI Points in PI. 
For example if an asset is as follows : 

`{ sinusoid: ["sinusoid"] }`. Then PI point will be something like  `1100001553_measurement_type1sinusoid` 

And if it contains more than one data points like `{ sinusoid: ["sinusoid1", "sinusoid2"] }` . There will be two PI points  

`1100001553_measurement_type1sinusoid.sinusoid1`  and `1100001553_measurement_type1sinusoid.sinusoid2`

docs/OMF.rst
```diff
@@ -459,12 +459,29 @@ Asset Framework Location Hint
...
+
+- An asset in a Fledge Reading is used to create a `Container in the OSIsoft Asset Framework <https://docs.osisoft.com/bundle/omf-with-pi-web-api/page/container-messages.html>`_.
+  A *Container* is an AF Element with an AF Attribute that is mapped to a PI Point using the OSIsoft PI Point Data Reference.
+  The names of both the AF Element and the AF Attribute are the Fledge Reading asset name.
+- If you edit the AF Location hint, the Container will be moved to the new location in the AF hierarchy.
+- If you disable the OMF Hint filter, the Container will not move.
+- If you wish to move a Container, you can do this with the PI System Explorer.
+  Right-click on the AF Element that represents the Container.
+  Choose Copy.
+  Select the AF Element that will serve as the new parent of the Container.
```
deepanshu-yadav, line 479: Would be better if we attach screenshots?

RVerhoeff: I am trying to avoid describing the PI points created by OMF and stick with the AF Element and AF Attributes that are mapped to them. You raise a good point about a Fledge asset with more than one data point. I presume these become more than one AF Attribute within an AF Element. My main point is that an administrator can move the AF Element; we don't have to do it for them. Regarding screenshots, it might help but it would also make that section of documentation large and hard to navigate. We would also be documenting AVEVA's product.

MarkRiddoch (approved):

deepanshu-yadav (approved):

PR #802, opened 2022-08-30
FOGL-6825 Correct documentation, South never runs as a task

MarkRiddoch:


Mohit04tomar (approved):

Mohit04tomar (commented):
in line 'which continuously wait for new data', Should wait be replaced with wait/poll?

PR #804, opened 2022-09-01
FOGL-6851 Add documentation regarding installing pip packages via the GUI

MarkRiddoch:


ashish-jabble (commented):
@MarkRiddoch This should really under building pipelines section? or in other section As Developer is another menu option in GUI on enabling from Settings.
IMO, we should add another index Develop features and describe there and we can link that in building pipelines if really required.

Also Develop features for production release as well or only with Dev/nightly/develop?

MarkRiddoch:
> @MarkRiddoch This should really under building pipelines section? or in other section As Developer is another menu option in GUI on enabling from Settings. IMO, we should add another index Develop features and describe there and we can link that in building pipelines if really required.

It is in the section because this is the section that talks about using Python packages and is all we have at the moment. We should add a proper section on the GUI with references for the different pages, but that is a bigger job and will duplicate other sections.

We need to have this documented in release 2.0, so thesis the best option we have for now.

> 
> Also Develop features for production release as well or only with Dev/nightly/develop?

The developer features are for targeted for the 2.0 release of FogLAMP.

ashish-jabble (approved):

PR #806, opened 2022-09-05
FOGL-6854 When a bearer token can not be refreshed by a service request

MarkRiddoch:
a restart of the service

### DIFF
--- a/python/fledge/services/common/microservice.py
+++ b/python/fledge/services/common/microservice.py
@@ -119,6 +119,10 @@ class FledgeMicroservice(FledgeProcess):
         pass
 
     @abstractmethod
+    async def restart(self, request):
+        pass
+
+    @abstractmethod
     async def change(self, request):
         pass
 
--- a/python/fledge/services/core/server.py
+++ b/python/fledge/services/core/server.py
@@ -1242,6 +1242,38 @@ class Server:
             raise web.HTTPNotFound(reason=str(ex))
 
     @classmethod
+    async def restart_service(cls, request):
+        """ Restart a service
+
+        :Example:
+            curl -X PUT  http://localhost:<core mgt port>/fledge/service/dc9bfc01-066a-4cc0-b068-9c35486db87f/restart
+        """
+
+        try:
+            service_id = request.match_info.get('service_id', None)
+
+            try:
+                services = ServiceRegistry.get(idx=service_id)
+            except service_registry_exceptions.DoesNotExist:
+                raise ValueError('Service with {} does not exist'.format(service_id))
+
+            ServiceRegistry.restart(service_id)
+
+            if cls._storage_client_async is not None and services[0]._name not in ("Fledge Storage", "Fledge Core"):
+                try:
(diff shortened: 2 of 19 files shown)


pintomax (dismissed):

python/fledge/services/core/server.py
```diff
@@ -1241,6 +1241,38 @@ async def unregister(cls, request):
...
+                services = ServiceRegistry.get(idx=service_id)
+            except service_registry_exceptions.DoesNotExist:
+                raise ValueError('Service with {} does not exist'.format(service_id))
+
+            ServiceRegistry.restart(service_id)
+
+            if cls._storage_client_async is not None and services[0]._name not in ("Fledge Storage", "Fledge Core"):
+                try:
+                    cls._audit = AuditLogger(cls._storage_client_async)
+                    await cls._audit.information('SRVRS', {'name': services[0]._name})
```
ashish-jabble, line 1265: SRVRS - It's a new audit log entry, We need change in schema with upgrade/downgrade scripts

python/fledge/services/core/server.py
```diff
@@ -1241,6 +1241,38 @@ async def unregister(cls, request):
...
+                except Exception as ex:
+                    _logger.exception(str(ex))
+
+            _resp = {'id': str(service_id), 'message': 'Service restart requested'}
+
+            return web.json_response(_resp)
+        except ValueError as ex:
+            raise web.HTTPNotFound(reason=str(ex))
```
ashish-jabble, line 1273: Also catch general exception

ashish-jabble (changes requested):
Python unit tests are failing

python/fledge/services/common/microservice.py
```diff
@@ -118,6 +118,10 @@ def _get_service_registration_payload(self):
     async def shutdown(self, request):
         pass
 
+    @abstractmethod
+    async def restart(self, request):
+        pass
+
```
ashish-jabble, line 124: Also I believe we do not need this, as restart is not common to all micro services. Instead only with core.

pintomax (dismissed):

pintomax (dismissed):

ashish-jabble:
@MarkRiddoch Please revert the checked in unit test_microservice.py file as we don't need restart def to pass on; as common def has been removed in the [commit](https://github.com/fledge-iot/fledge/pull/806/commits/f47ee3b3345b2ec7733b20dc09c5889d030da5b3#diff-212d4bd9669aeb1af80a85c7c8f12cfbd57112b02fc3bd933d2b4989968e9d18). Instead we need to fix the [audit system tests](https://github.com/fledge-iot/fledge/blob/develop/tests/system/python/api/test_audit.py#L25).
Also I am not able to restart the service if the service state is in the failed state. On calling restart endpoint it gives 500 with empty message.

MarkRiddoch:
> Also I am not able to restart the service if the service state is in the failed state. On calling restart endpoint it gives 500 with empty message.

It is not meant to. This is for the case when the service itself is requesting a restart

pintomax (approved):

PR #809, opened 2022-09-06
Fogl 6744 storage asset tracker

Innovative-ashwin:


MarkRiddoch (dismissed):

MarkRiddoch (approved):

C/common/storage_asset_tracking.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+	                storageAssetTrackerTuplesCache.erase(it);
+			return NULL;
+                }
+                else if ((*it)->m_maxCount == ptr->m_maxCount)
+                {
+                // case where counts are same but datapoints are different
+                // "a", "b", "c" and "a", "b", "foo"
+                // keep both the records
+        	        if (compareDatapoints(ptr->m_datapoints,(*it)->m_datapoints))
```
AmandeepArora, line 145: compare should still return a match if it is the same set of DPs but in a different order. Please ensure this.

C/common/storage_asset_tracking.cpp
```diff
@@ -0,0 +1,274 @@
...
+               		{
+				//  record to be addded 
+				Logger::getLogger()->debug("%s:%d:%s tuple present and case where counts are same but datapoints are different, update cache ",  __FILE__,__LINE__, __FUNCTION__);
```
AmandeepArora, line 148: Please remove __FILE__ from logs as complete path of source file would be stored into binary/library, which is not desirable.

C/services/south/ingest.cpp
```diff
@@ -499,14 +503,35 @@ void Ingest::processQueue()
...
 				for (vector<Reading *>::iterator it = q->begin();
 							 it != q->end(); ++it)
 				{
 					Reading *reading = *it;
 					string assetName = reading->getAssetName();
+                                        const std::vector<Datapoint *> dpVec = reading->getReadingData();
+					std::string temp;
+
+                                        for ( auto dp : dpVec)
+                                        {
```
AmandeepArora, line 526: I think we should maintain DPs separately here rather than combining them in a single string to take care of change in order of DPs.

C/services/south/ingest.cpp
```diff
@@ -499,14 +503,35 @@ void Ingest::processQueue()
...
+				StorageAssetTracker *satracker = StorageAssetTracker::getStorageAssetTracker();
+				if ( satracker == nullptr)
+                                {
+                                        Logger::getLogger()->error("%s:%d could not initialize satracker ", __FILE__, __LINE__);
+					return;
+                                }
 
 				string lastAsset = "";
 				int *lastStat = NULL;
+				std::map <std::string , std::set<std::string> > assetDatapointMap;
```
AmandeepArora, line 515: How do we handle the case and store information in this map container where one reading for asset a1 has DPs - dp1, dp2 & dp3. And then another reading for same asset a1 has DPs - dp3, dp4 & dp5.

C/services/south/ingest.cpp
```diff
@@ -499,14 +503,35 @@ void Ingest::processQueue()
...
 					string assetName = reading->getAssetName();
+                                        const std::vector<Datapoint *> dpVec = reading->getReadingData();
+					std::string temp;
+
+                                        for ( auto dp : dpVec)
+                                        {
+                                                temp.append("\"");
+                                                temp.append(dp->getName());
+                                                temp.append("\"");
+                                                assetDatapointMap[assetName].insert(temp);
```
AmandeepArora, line 530: This insertion doesn't make sense in this loop. Please check.

AmandeepArora (changes requested):

C/common/management_client.cpp
```diff
@@ -1540,3 +1541,418 @@ ACL ManagementClient::getACL(const string& aclName)
...
+                }
+                else
+                {
+                        const rapidjson::Value& trackArray = doc["track"];
+                        if (trackArray.IsArray())
+                        {
+                                // Process every row and create the AssetTrackingTuple object
+                                for (auto& rec : trackArray.GetArray())
+                                {
+					 m_logger->error("%s:%d Inside for loop of trackArray ", __FILE__, __LINE__);
```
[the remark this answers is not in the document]
Innovative-ashwin: fixed

C/common/storage_asset_tracking.cpp
Innovative-ashwin: already using a set to order the dp before passing them to the function findStorageAssetTrackingCache

C/services/south/ingest.cpp
Innovative-ashwin: have modified data structure

C/services/south/ingest.cpp
Innovative-ashwin: have modified data structure, its now map <asset , object> where object si a structure of { set<string> and count>

the set <string> will contain one reading dataPoints ( that too in sorted order)

C/common/management_client.cpp
```diff
@@ -1540,3 +1541,418 @@ ACL ManagementClient::getACL(const string& aclName)
...
+ * @param    event              The associated event type
+ * @param    dp			The datapoints Type
+ * @param    c			The count of datapoints
+ * @return              	A pointer to AssetTrackingTuple objects allocated on heap
+ */
+StorageAssetTrackingTuple* ManagementClient::getStorageAssetTrackingTuple(const std::string& serviceName,
+                                                        const std::string& assetName,
+                                                        const std::string& event, const std::string& dp, const unsigned int& c)
+{
+	m_logger->error("%s:%d function %s start  service = %s , assetName = %s ", __FILE__, __LINE__,__FUNCTION__, serviceName.c_str(), assetName.c_str());
```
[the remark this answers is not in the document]
Innovative-ashwin: Have removed this

C/common/storage_asset_tracking.cpp
Innovative-ashwin: Fixed

C/services/south/ingest.cpp
```diff
@@ -522,15 +527,35 @@ void Ingest::processQueue()
...
 					std::string temp;
 
+					std::set<std::string> tempSet;
+					// first sort the individual datapoints 
+					// e.g. dp2, dp3, dp1 push them in a set,to make them 
+					// dp1,dp2,dp3
                                         for ( auto dp : dpVec)
                                         {
+						temp.clear();
                                                 temp.append("\"");
```
AmandeepArora, line 537: Why do we need enclosing double quotes?

C/services/south/ingest.cpp
```diff
@@ -522,15 +527,35 @@ void Ingest::processQueue()
...
+						if ( i> 0) temp.append(",");
+						temp.append(setItr);
+						++i;
+					}
+
+					// Push them in a set so as to avoid duplication of datapoints
+					// a reading of d1, d2, d3 and another d2,d3,d1 , second will be discarded
+
+					assetDatapointMap[assetName].setDp.insert(temp);
+                                        assetDatapointMap[assetName].count = i;
```
AmandeepArora, line 558: Do we always insert the dpset into map? Or only when the count has increased from previous value? If we insert always, then finally before updating cache, do we find the dpset with the maximum count for each asset seen?

C/services/south/ingest.cpp
```diff
@@ -739,23 +759,43 @@ void Ingest::processQueue()
...
-						std::string temp;
-						temp.append("\"");
-						temp.append(dp->getName());
-						temp.append("\"");
-						assetDatapointMap[assetName].insert(temp);
-					}
+                                        for ( auto dp : dpVec)
+                                        {
+						temp.clear();
+                                                temp.append("\"");
```
AmandeepArora, line 779: do we need enclosing quotes?

C/services/south/ingest.cpp
```diff
@@ -797,43 +836,34 @@ void Ingest::processQueue()
...
-                                  	}    
+                                  		if (rv == NULL)
+                                  		{
+	                                       		// Record not found in cache , please update cache
+							Logger::getLogger()->debug("%s:%d record not found in cache  add it", __FUNCTION__, __LINE__);
+                                                	satracker->addStorageAssetTrackingTuple(storageTuple);
+                                  		}
+                                  		else
+                                  		{
+							//record not found undeprecate the record
```
AmandeepArora, line 861: update comment "record found"

C/services/south/ingest.cpp
Innovative-ashwin: as per requirement in JIIRA

C/services/south/ingest.cpp
Innovative-ashwin: whether to insert in db or not is checked in find function in the storageassettracker

C/services/south/ingest.cpp
Innovative-ashwin: as per requirement on jIRA

AmandeepArora: As per JIRA, "datapoints" is an array of dp name strings, and in JSON, strings are double quoted, there is no such requirement in C++ code, only when conversion to JSON happens. Can you show me sample JSON from ur testing?

C/common/storage_asset_tracking.cpp
```diff
@@ -0,0 +1,274 @@
...
+ *
+ * @param tuple        StorageAssetTrackingTuple Type
+ * @return	       A pointer to StorageAssetTrackingTuple in cache or null
+ */
+
+
+StorageAssetTrackingTuple* StorageAssetTracker::findStorageAssetTrackingCache(StorageAssetTrackingTuple& tuple)	
+{
+	StorageAssetTrackingTuple *ptr = &tuple;
+	std::unordered_multiset<StorageAssetTrackingTuple*>::const_iterator it = storageAssetTrackerTuplesCache.find(ptr);
```
AmandeepArora, line 121: this can return multiple iterators for a given service/asset/plugin/... tuple? If so, handle that.

C/services/south/ingest.cpp
```diff
@@ -452,6 +455,12 @@ void Ingest::waitForQueue()
  */
 void Ingest::processQueue()
 {
+
+	typedef struct{
+                std::set<std::string> setDp;
+                unsigned int          count;
```
AmandeepArora, line 461: count needs to be maintained for each dplist we are maintaining.

python/fledge/services/core/server.py
```diff
@@ -1620,15 +1620,28 @@ async def get_track(cls, request):
...
 
+        jsondata=data.get("data")
+
         try:
-            result = await cls._asset_tracker.add_asset_record(asset=data.get("asset"),
-                                                               plugin=data.get("plugin"),
-                                                               service=data.get("service"),
-                                                               event=data.get("event"))
+            if jsondata is None:
+
```
deepanshu-yadav, line 1633: We can use 
```
asset_dict = {'asset': data.get("asset"), 
                      'plugin': data.get("plugin")       
                        'service': data.get("service")
                       'event': data.get("event")
 }
if jsondata is not None:
    asset_dict['jsondata'] = data.get("data")
result = await cls._asset_tracker.add_asset_record(**asset_dict)

```

Innovative-ashwin: that did not work for assettracker class .

C/common/storage_asset_tracking.cpp
Innovative-ashwin: handled that

C/services/south/ingest.cpp
Innovative-ashwin: fixed

C/services/south/ingest.cpp
Innovative-ashwin: This might be a choice issue , whether to keep it here or adding in management-client.cpp

AmandeepArora (approved):

deepanshu-yadav (approved):
API working now as expected.

PR #819, opened 2022-09-12
FOGL-6898 Add description for securioty category

MarkRiddoch:


pintomax (approved):

deepanshu-yadav:
This may cause errors https://github.com/fledge-iot/fledge/blob/51e4e8bc9cda367922befc05b73560d1479868cc/python/fledge/services/core/api/control_service/acl_management.py#L320  it assumes security category contains these two values. One more item will be added. 

Let me test it. then only merge.

PR #820, opened 2022-09-12
Pair system test fix

deepanshu-yadav:
Tests are passing now.


tests/system/python/pair/test_pyton_north_service_pair.py
```diff
@@ -21,7 +21,7 @@
 
 # Local machine
 local_south_plugin = "sinusoid"
-local_south_asset_name = "sinusoid"
+local_south_asset_name = "python-north-service-pair"
```
praveen-garg, line 24: Why should this be the same? As remote north?

tests/system/python/pair/test_pyton_north_service_pair.py
```diff
@@ -55,7 +55,7 @@ def reset_fledge_local(wait_time):
 
 @pytest.fixture
 def setup_local(reset_fledge_local, add_south, add_north, fledge_url, remote_ip):
-    local_south_config = {"assetName": {"value": remote_south_asset_name}}
+    local_south_config = {"assetName": {"value": local_south_asset_name}}
```
praveen-garg, line 58: Local and remote both same?

tests/system/python/pair/test_pyton_north_service_pair.py
deepanshu-yadav: Remote asset name does not matter. the local set up sends data to remote by http north and then to PI. So if the local asset name is for example asset1 then pi point created will be asset1.

tests/system/python/pair/test_pyton_north_service_pair.py
deepanshu-yadav: Remote asset name does not matter. The local set up sends data to remote by http north and then to PI. So if the local asset name is for example asset1 then pi point created will be asset1.

Mohit04tomar (approved):

PR #821, opened 2022-09-12
FOGL-6799 Add Dryrun option to Python tasks

MarkRiddoch:


pintomax (approved):

pintomax (approved):

python/fledge/tasks/north/sending_process.py
```diff
@@ -925,6 +925,9 @@ async def _start(self):
 
             exec_sending_process = self._config['enable']
 
+            if self.isDryRun():
+                return
```
ashish-jabble, line 929: With --dryrun Sending process logs are coming with Started....Stop...Execution Completed, which means EGRESS of readings is going on if north server is reachable.

python/fledge/tasks/north/sending_process.py
```diff
@@ -976,6 +975,11 @@ async def run(self):
         global _log_performance
         global _LOGGER
 
+        # Return early if task runs with --dryrun arg
+        # In future if any runtime configuration required for a task then it should be created before it
+        if self.is_dry_run():
```
praveen-garg, line 980: move above globals?

praveen-garg (commented):
> below logs I am getting on startup, not sure how to handle that. Still investigating!

We should be seeing these in develop too, NO? Scheduler on start calls _get_schedules (mind _ before it) and then schedule the first task and then next task... 

What is the concern here, to investigate?!

ashish-jabble:
> We should be seeing these in develop too, NO?

Nope! 
With the given branch now on every startup you see the logs ONLY; not the actual task run as it early exit with dryrun mechanism.

praveen-garg:
> not the actual task run as it early exit with dryrun mechanism.

You mean, now tasks are always running with dry run flag? I'm not following the earlier comment

ashish-jabble:
> You mean, now tasks are always running with dry run flag? I'm not following the earlier comment

On every boot yes all of the tasks except STARTUP type called with dry-run flag. As asked in JIRA

ashish-jabble (approved):

PR #828, opened 2022-09-19
Fogl 6877 API level changes for getting health of logging.

deepanshu-yadav:
Example 
```
$ curl -sX GET http://localhost:8081/fledge/health/logging | jq 
{
  "levels": [
    {
      "name": "sss",
      "level": "warning"
    }
  ],
  "disk": {
    "used": 95711760,
    "usage": 83,
    "available": 20493872
  }
}

```

### DIFF
--- a/python/fledge/services/core/api/health.py
+++ b/python/fledge/services/core/api/health.py
@@ -102,7 +102,6 @@ async def get_logging_health(request: web.Request) -> web.Response:
         for row in results["rows"]:
             for item_name, item_info in row["value"].items():
                 if item_name == "logLevel" and item_info['value'] not in excluded_log_levels:
-                    _LOGGER
                     service_name = row["key"].replace("Advanced", "").strip()
                     log_level = item_info['value']
                     log_levels.append({"name": service_name, "level": log_level})
--- a/python/fledge/services/core/api/health.py
+++ b/python/fledge/services/core/api/health.py
@@ -102,7 +102,8 @@ async def get_logging_health(request: web.Request) -> web.Response:
         for row in results["rows"]:
             for item_name, item_info in row["value"].items():
                 if item_name == "logLevel" and item_info['value'] not in excluded_log_levels:
-                    service_name = row["key"]
+                    _LOGGER
+                    service_name = row["key"].replace("Advanced", "").strip()
                     log_level = item_info['value']
                     log_levels.append({"name": service_name, "level": log_level})
 
--- a/python/fledge/services/core/api/health.py
+++ b/python/fledge/services/core/api/health.py
@@ -96,22 +96,17 @@ async def get_logging_health(request: web.Request) -> web.Response:
         from fledge.services.core import connect
 
         services_info = serv_api.get_service_records()
-        levels_array = []
-        excluded_services = ["Storage", "Core"]
-        for services_info in services_info['services']:
-            if services_info['type'] not in excluded_services:
-                service_name = services_info["name"]
-                cf_mgr = ConfigurationManager(connect.get_storage_async())
-                category_name = service_name + "Advanced"
-                config_item = "logLevel"
-                category_item = await cf_mgr.get_category_item(category_name, config_item)
-                log_level = category_item["value"]
-                level_dict = dict()
-                level_dict["name"] = service_name
-                level_dict["level"] = log_level
-                levels_array.append(level_dict)
+        log_levels = []
+        excluded_services = ["Core"]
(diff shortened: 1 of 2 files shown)


python/fledge/services/core/api/health.py
```diff
@@ -26,6 +26,113 @@
...
+    try:
+        response['disk'] = {}
+        used, available, usage = await get_disc_usage('/var/log')
+        # fill all the fields after values are retrieved
+        response['disk']['used'] = used
+        response['disk']['usage'] = usage
+        response['disk']['available'] = available
+
+    except Exception as ex:
+        msg = "Failed to get disk stats of log!{}".format(str(ex))
```
AmandeepArora, line 129: Change to ""Failed to get disk stats for /var/log !

AmandeepArora (approved):

MarkRiddoch (dismissed):

python/fledge/services/core/api/health.py
```diff
@@ -21,11 +21,116 @@
...
+    """
+    response = {}
+    try:
+        from fledge.services.core.api import service as serv_api
+        from fledge.common.configuration_manager import ConfigurationManager
+        from fledge.services.core import connect
+
+        services_info = serv_api.get_service_records()
+        log_levels = []
+        excluded_services = ["Core"]
```
praveen-garg, line 100: exclude management service?

python/fledge/services/core/api/health.py
```diff
@@ -21,11 +21,116 @@
...
+        cf_mgr = ConfigurationManager(connect.get_storage_async())
+        for s in services_info['services']:
+            if s['type'] not in excluded_services:
+                service_name = s["name"]
+                cat_name = service_name + "Advanced"
+                if s['type'] == "Storage":
```
praveen-garg, line 106: Could you also check for notification & bucket advanced category?
key? I believe we have random things.

deepanshu-yadav:
Hi @MarkRiddoch , 

Tested for south, north, Dispatcher and notification services. Excluded Core and Management. 
One thing left is Bucket. Bucket service creates a Advanced category named "BucketServer Advanced" not 
"BucketServerAdvanced" unlike other services. 

I can handle it here and we can create BucketServerAdvanced just like any other service. Please guide.

MarkRiddoch:
> Hi @MarkRiddoch ,
> 
> Tested for south, north, Dispatcher,Storage and notification services. Excluded Core and Management. One thing left is Bucket. Bucket service creates a Advanced category named "BucketServer Advanced" not "BucketServerAdvanced" unlike other services.
> 
> I can handle it here or we can create BucketServerAdvanced just like any other service. Please guide.

I suggest it is best to handle it here by allowing any service to add white space between its name and the Advanced part.

python/fledge/services/core/api/health.py
MarkRiddoch: I don't think there is any need to exclude the management service

python/fledge/services/core/api/health.py
```diff
@@ -91,25 +91,37 @@ async def get_logging_health(request: web.Request) -> web.Response:
...
+        schedules_info['schedules'] = schedules
+        schedules_info['schedules'].append({"type": "STARTUP",
+                                            "processName": "storage",
+                                            "name": "Storage"})
         log_levels = []
-        excluded_services = ["Core", "Management"]
         cf_mgr = ConfigurationManager(connect.get_storage_async())
-        for s in services_info['services']:
-            if s['type'] not in excluded_services:
+        for s in schedules_info['schedules']:
```
MarkRiddoch, line 115: This is fine, but every time we add a new service type we will need to update this.

python/fledge/services/core/api/health.py
```diff
@@ -21,11 +21,113 @@
...
+        payload = PayloadBuilder().SELECT("key", "value").payload()
+        _storage_client = connect.get_storage_async()
+        results = await _storage_client.query_tbl_with_payload('configuration', payload)
+        log_levels = []
+        for row in results["rows"]:
+            for item_name, item_info in row["value"].items():
+                if item_name == "logLevel":
+                    service_name = row["key"]
+                    log_level = item_info['value']
+                    log_levels.append({"name": service_name, "level": log_level})
```
MarkRiddoch, line 106: Only if log_level is not one of error or warning

MarkRiddoch (changes requested):

MarkRiddoch (approved):

PR #833, opened 2022-09-20
FOGL-6938 Fix crash with incorrect argument count for get_plugin_info

MarkRiddoch:

### DIFF
--- a/C/plugins/utils/get_plugin_info.cpp
+++ b/C/plugins/utils/get_plugin_info.cpp
@@ -24,15 +24,20 @@ typedef PLUGIN_INFORMATION *(*func_t)();
  *
  * @param argv[1]  relative/absolute path to north/south C plugin shared library
  *
- * @param argv[2]  symbol to extract value from (typically 'plugin_info')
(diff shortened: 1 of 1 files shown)


C/plugins/utils/get_plugin_info.cpp
```diff
@@ -24,15 +24,20 @@ typedef PLUGIN_INFORMATION *(*func_t)();
  *
  * @param argv[1]  relative/absolute path to north/south C plugin shared library
  *
- * @param argv[2]  symbol to extract value from (typically 'plugin_info')
+ * @param argv[2]  symbol to extract value from (defaults 'plugin_info')
```
ashish-jabble, line 27: So, it's optional now!

ashish-jabble (approved):

AmandeepArora (approved):

C/plugins/utils/get_plugin_info.cpp
MarkRiddoch: Yes, it will not effect how we use it internally, but is easier for the user.

PR #834, opened 2022-09-21
FOGL-1904 For Python tasks configuration available at boot startup with dryrun mechanism

ashish-jabble:

- [x] Updated system tests as now Purge configuration available on startup

### DIFF
--- a/python/fledge/services/core/server.py
+++ b/python/fledge/services/core/server.py
@@ -895,7 +895,28 @@ class Server:
                                  "Therefore, starting the service...")
                     loop.run_until_complete(cls.add_and_enable_dispatcher())
                     _logger.info("Dispatcher service started.")
-
+                schedule_list = loop.run_until_complete(cls.scheduler.get_schedules())
+                for sch in schedule_list:
+                    # STARTUP type exclusion
+                    if int(sch.schedule_type) != 1:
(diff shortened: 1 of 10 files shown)


python/fledge/services/core/server.py
```diff
@@ -895,7 +895,28 @@ def _start_core(cls, loop=None):
                                  "Therefore, starting the service...")
                     loop.run_until_complete(cls.add_and_enable_dispatcher())
                     _logger.info("Dispatcher service started.")
-
+                schedule_list = loop.run_until_complete(cls.scheduler.get_schedules())
+                for sch in schedule_list:
+                    # STARTUP type exclusion
+                    if int(sch.schedule_type) != 1:
+                        _logger.debug("{}-{}-{}-{}".format(str(sch.schedule_id), sch.name, int(sch.schedule_type), sch.process_name))
+                        # TODO: Errors on start up enable below line (if statement) otherwise it will queue the tasks for backup & restore (as inbuilt schedules)
```
MarkRiddoch, line 903: We should probably include backup at least, but maybe also restore. They may in the future add configuration. 

I've added these to FOGL-6799

MarkRiddoch (approved):

python/fledge/services/core/server.py
ashish-jabble: I am fixing one by one as there are issues in FOGL-6799 which needs to be fixed. Pushed changes into this branch. As this PR is created on top of FOGL-6799.

PR #857, opened 2022-10-20
FOGL-7058 Resolve warning on Ubuntu 18.04 platform

MarkRiddoch:


Innovative-ashwin (approved):

C/services/south/south.cpp
```diff
@@ -142,8 +142,8 @@ pid_t pid;
 	close(2);
 	// redirect fd's 0,1,2 to /dev/null
 	(void)open("/dev/null", O_RDWR);  	// stdin
```
Innovative-ashwin, line 144: why this line (void)open("/dev/null", O_RDWR);   is not removed as in north.cpp?

PR #863, opened 2022-10-27
FOGL-7062 Add support to sqlitelb to create readings table

MarkRiddoch:
Used when we set sqlitelb as the readings plugin while keeping sqlite as the configuration plugin


pintomax (approved):

C/plugins/storage/sqlitelb/common/connection.cpp
```diff
@@ -587,6 +606,151 @@ Connection::Connection()
 		//Release sqlStmt buffer
 		delete[] sqlReadingsStmt;
 
+		if (initialiseReadings)
```
deepanshu-yadav, line 609: Why are doing this isn't init.sql of sqlitelb storage does this?  I am suggesting the following  thing 

when storage.sh fetches the required storage plugin the storage.cpp should also return readings plugin along with storage plugin name. The script will execute (storage.sh) will execute corresponding shell scripts of the required plugins. 

For example if we give storage plugin sqlite and readings plugin sqlitelb then it should execute the shell script  
 sqlite.sh  and sqlitelb.sh respectively.

Makefile
```diff
@@ -25,7 +25,7 @@ ifneq ("$(PLATFORM_RH)","")
 else
 	PIP_INSTALL_REQUIREMENTS := pip3 install -Ir
 	PYTHON_BUILD_PACKAGE = python3 setup.py build -b ../$(PYTHON_BUILD_DIR)
-	CMAKE := cmake
+	CMAKE := cmake -DCMAKE_BUILD_TYPE=Debug
 endif
```
deepanshu-yadav, line 29: Should be removed afterwards i suppose?

C/plugins/storage/sqlitelb/common/connection.cpp
MarkRiddoch: It doesn't, but it assumes the sqlitelb is being used for both config and readings, that is not the case the is solving. Tis is for when we switch sqlitelb in for just the readings. Ideally we want to run the init_readings.ql, but there are issues with that, so this is a short term fix until we get a chance to rethink the initialisation. The same thing holds true if Postgres is used for readings only.

RVerhoeff (approved):

praveen-garg:
@MarkRiddoch This is waiting for some other fixes?

pintomax (approved):

PR #868, opened 2022-11-02
FOGL-7060 New user role types & some restrictions added to API as per user

ashish-jabble:
**Items**
- [x] Update the CRUD API to support these user types.
- [x] Implement testing on all the API endpoints such that a view user can only call GET end points.
- [x] Implement testing on all API endpoints such that the data_view user can only access GET endpoints where the stub of the URL is /fledge/asset
- [x] New System tests added for different user types & existing system tests updated 

**Storage engine**
- [x] Sqlite
- [x] Sqlitelb
- [x] PostgreSQL

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,2 +1,2 @@
 fledge_version=2.0.0
-fledge_schema=58
+fledge_schema=59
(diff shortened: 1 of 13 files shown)


MarkRiddoch (commented):
This looks good to me, we probably should get the other database plugins supported before checking in to avoid having different schema versions on different plugins.

As for testing, I would say use the API to do the tests rather than the GUI, just seems simpler.

MarkRiddoch (approved):

PR #871, opened 2022-11-03
FOGL-7076:added check for update cache on count increase

Innovative-ashwin:
Signed-off-by: Innovative-ashwin <sinha.ashwini111@gmail.com>


praveen-garg (approved):
lgtm! I can see now that asset tracker only get entries if datapoint count changes.

PR #874, opened 2022-11-04
Reuse empty readings tables in sqlite storage plugin

gnandan:

### DIFF
--- a/C/plugins/storage/sqlite/common/readings_catalogue.cpp
+++ b/C/plugins/storage/sqlite/common/readings_catalogue.cpp
@@ -1835,6 +1835,7 @@ ReadingsCatalogue::tyReadingReference  ReadingsCatalogue::getReadingReference(Co
 
 		AttachDbSync *attachSync = AttachDbSync::getInstance();
 		attachSync->lock();
+		ReadingsCatalogue::tyReadingReference emptyTableReference = {-1, -1};
 
 		auto item = m_AssetReadingCatalogue.find(asset_code);
 		if (item != m_AssetReadingCatalogue.end())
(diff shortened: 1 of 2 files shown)


C/plugins/storage/sqlite/common/readings_catalogue.cpp
```diff
@@ -1933,6 +1960,52 @@ ReadingsCatalogue::tyReadingReference  ReadingsCatalogue::getReadingReference(Co
...
+
+	ConnectionManager *manager = ConnectionManager::getInstance();
+	Connection *connection = manager->allocate();
+	dbHandle = connection->getDbHandle();
+	
+	for (auto &item : m_AssetReadingCatalogue)
+	{
+		int tableId = item.second.first;
+		int dbId = item.second.second;
+		sql_cmd = "SELECT COUNT(*) FROM (SELECT 0 FROM readings_" + to_string(dbId) + ".readings_" + to_string(dbId) + "_" + to_string(tableId) + " LIMIT 1)";
```
MarkRiddoch, line 1985: Why the nested select?

MarkRiddoch (changes requested):

C/plugins/storage/sqlite/common/readings_catalogue.cpp
```diff
@@ -1874,7 +1875,16 @@ ReadingsCatalogue::tyReadingReference  ReadingsCatalogue::getReadingReference(Co
 
 						startReadingsId = 1;
 
-						success = createNewDB(dbHandle,  dbId, startReadingsId, NEW_DB_ATTACH_REQUEST);
+						if (getEmptyReadingTableReference(emptyTableReference))
+						{
+							emptyTableReference.tableId;
```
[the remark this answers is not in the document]
gnandan: I don't want to assign anything here because I am passing variable emptyTableReference by reference to function getEmptyReadingTableReference so I will get back updated values if any. To avoid confusion I changed if-else construct to if

C/plugins/storage/sqlite/common/readings_catalogue.cpp
gnandan: asset_reading_catalogue table is just having only details to get the table not the actual table. That's why inner query it getting actual table and outer query is checking the number of rows from the table returned by inner query

MarkRiddoch (approved):

praveen-garg (approved):

PR #880, opened 2022-11-15
DataPointUtility and JsonToDatapoint

YmaIneo:
Added features:
 * Utility methods for searching and creating datapoints (namespace DatapointUtility, files datapoint_utility.h/cpp)
 * Method to convert JSOn format to datapoint format (included in the Datapoint class)

Signed-off-by: Yannick Marchetaux <yannick.marchetaux@equans.com>


C/common/datapoint.cpp
```diff
@@ -353,3 +353,62 @@ int bscount = 0;
...
 	return rval;
 }
+
+/**
+ * Parsing a Json string
+ * 
+ * @param json : string json 
+ * @return vector of datapoints
+*/
+std::vector<Datapoint*>* Datapoint::parseJson(std::string json) {
```
jchabod, line 363: Please use a const reference rather than a copy( `const std::string& json`)

C/common/datapoint_utility.cpp
```diff
@@ -0,0 +1,172 @@
...
+}
+
+/**
+ * Generate default attribute dict on Datapoint
+ * 
+ * @param dps dict of values 
+ * @param key key of dict
+ * @return pointer of the created datapoint
+ */
+Datapoint * DatapointUtility::createDictElement(Datapoints * dps, const string & key) {
```
jchabod, line 162: Shouldn't you also add a `createListElement`?`

C/common/datapoint.cpp
```diff
@@ -353,3 +353,62 @@ int bscount = 0;
...
+	
+	rapidjson::Document document;
+
+    const auto & parseResult = document.Parse(const_cast<char*>(json.c_str()));
```
jchabod, line 367: `const_cast` does not seem required here

jchabod:
Thanks for update.
I am not part of FLEDGE process, so I do not know if I can actually approve this PR. FYI, I simply added remarks because I was interested in that PR.

I would say that is OK for me except that :
- There remain mix-up between spaces and tabs
- I don't know if this "intensive" use of `auto` is acceptable (it is forbidden in projects I am usually involved in, but I do not know if this is applicable here)
- It could be possible to avoid code duplication by re-using `findDatapointElement` in other similar function, but once again, I am not aware of coding guidelines for this project.

YmaIneo:
> Thanks for update. I am not part of FLEDGE process, so I do not know if I can actually approve this PR. FYI, I simply added remarks because I was interested in that PR.
> 
> I would say that is OK for me except that :
> 
> * There remain mix-up between spaces and tabs
> * I don't know if this "intensive" use of `auto` is acceptable (it is forbidden in projects I am usually involved in, but I do not know if this is applicable here)
> * It could be possible to avoid code duplication by re-using `findDatapointElement` in other similar function, but once again, I am not aware of coding guidelines for this project.

Thanks for your comment. 

I have made the following changes :
* Replacement of spaces by tabs
* Refactoring of methods (use method `findDatapointElement` in other  similar function)
* Add method `findDictOrListElement` and `createDictOrListElement` (Refactoring)
* Removal of the use of "auto".

I'm contributing to the "fledge power" project. This PR will allow us to simplify our plugins and services.

C/common/datapoint_utility.cpp
```diff
@@ -0,0 +1,224 @@
...
+	}
+	return nullptr;
+}
+
+/**
+ * Search a string from a key
+ *
+ * @param dict : parent dictionary
+ * @param key : key to research
+ * @return correponding string otherwise empty string
```
jchabod, line 104: - mistyping for corresponding

C/common/datapoint.cpp
```diff
@@ -353,3 +353,59 @@ int bscount = 0;
...
+ * Parsing a Json string
+ * 
+ * @param json : string json 
+ * @return vector of datapoints
+*/
+std::vector<Datapoint*> *Datapoint::parseJson(const std::string& json) {
+	
+	rapidjson::Document document;
+
+    const auto& parseResult = document.Parse(json.c_str());
```
jchabod, line 367: There remain some mix-up between space an tabs

C/common/datapoint_utility.cpp
```diff
@@ -0,0 +1,224 @@
...
+	const DatapointValue::dataTagType dType(data.getType());
+	if (dType == DatapointValue::T_STRING) {
+		return data.toStringValue();
+	}
+
+	return "";
+}
+
+/**
+ * Method to delete elements in a vector
```
jchabod, line 124: It seems important to explain that the deleted elements are also freed

C/common/datapoint_utility.cpp
```diff
@@ -0,0 +1,224 @@
...
+ * @param dps dict of values 
+ * @param key key of dict 
+*/
+void DatapointUtility::deleteValue(Datapoints *dps, const string& key) {
+	Datapoints::iterator it1 = dps->end();
+	Datapoint *d = nullptr;
+	for (Datapoints::iterator it = dps->begin(); it != dps->end(); it++){
+		if ((*it)->getName() == key) {
+			it1 = it;
+			d = *it;
```
jchabod, line 135: If you delete the content here you won't need `it1`, `d`and following `if` branch

C/common/datapoint_utility.cpp
```diff
@@ -0,0 +1,224 @@
...
+ * Generate default attribute integer on Datapoint
+ * 
+ * @param dps dict of values 
+ * @param key key of dict
+ * @param valueDefault value attribute of dict
+ * @return pointer of the created datapoint
+ */
+Datapoint *DatapointUtility::createIntegerElement(Datapoints *dps, const string& key, long valueDefault) {
+
+	deleteValue(dps, key);
```
jchabod, line 156: Ok with that, but this method implies that the insertion gets a `O(n)` complexity.

YmaIneo: I don't understand, can you give me more details?

jchabod: I Just mean that unconditionnaly calling delete will imply a complete loop over all elements, which leads to a complexity of `O(n)`. That may lead to performances issues. however, I don't see any real option as nothing is designed to avoid this linear complexity...

jchabod (approved):

C/common/datapoint_utility.cpp
YmaIneo: Ah ok. thanks for the details. 
When calling the delete method, the loop will browse the elements until it finds the one with the key. It doesn't go through all the elements.
If not, I have no other idea.

MarkRiddoch (approved):
Thank you for this contribution

PR #885, opened 2022-11-18
FOGL-7082 deprecated asset record handling for Ingest & store event internally in API along with unit tests

ashish-jabble:


tests/unit/python/fledge/services/core/api/test_asset_tracker_api.py
```diff
@@ -24,6 +24,10 @@
 __version__ = "${VERSION}"
 
 
+async def mock_coro(return_value):
```
Innovative-ashwin, line 27: suggest to write  a more meaningful name than mock_coro

Innovative-ashwin (approved):

MarkRiddoch (approved):

PR #888, opened 2022-11-22
FOGL-7128 Inhibit functionality to find empty readings table as it is causing a regression

MarkRiddoch:


ashish-jabble:
@MarkRiddoch Please rebase with develop branch correctly. Documentation and Asset browser changes are not required.

C/plugins/storage/sqlite/common/readings_catalogue.cpp
```diff
@@ -1969,6 +1969,9 @@ bool ReadingsCatalogue::getEmptyReadingTableReference(tyReadingReference &emptyT
 	string sql_cmd;
 	sqlite3_stmt *stmt;
 
+	// Disable functionality temporarily to avoid regression 
+	return false;
```
ashish-jabble, line 1973: As you have made a temporary disabled the empty readings tables feature. Do not we need the functionality of Reuse empty readings tables in sqlite storage plugin?

To handle the issue, do we need to reopen 7064 JIRA or to be handle in new JIRA or in this JIRA?

MarkRiddoch: Yes, we should reinstate the functionality, but we can not afford to have a broken storage plugin whilst we do this

ashish-jabble (approved):

PR #893, opened 2022-12-02
FOGL-6891: Handle storage service restart

AmandeepArora:
Storage service restart is handled in south service, north service & notification service. But this hasn't been tested a lot. Also lots of debug logs are still present in this current code.


C/common/logger.cpp
```diff
@@ -19,7 +19,7 @@
 using namespace std;
 
 // uncomment line below to get uSec level timestamps
-// #define ADD_USEC_TS
+#define ADD_USEC_TS
```
MarkRiddoch, line 22: Probably best not to checkin with this uncommented

C/common/storage_client.cpp
```diff
@@ -49,8 +55,42 @@ StorageClient::StorageClient(const string& hostname, const unsigned short port)
...
+ */
+StorageClient::StorageClient(const string& hostname, const unsigned short port, std::map<std::thread::id, std::atomic<int>>* m) : m_streaming(false), m_management(NULL)
+{
+	m_host = hostname;
+	m_pid = getpid();
+	m_logger = Logger::getLogger();
+	m_urlbase << hostname << ":" << port;
+	m_logger->info("StorageClient c'tor 2: this=%p, m_urlbase=%s", this, m_urlbase.str().c_str());
+	
+	m_logger->info("StorageClient c'tor 2: current thread ID=%u, m=%p", std::this_thread::get_id(), m);
```
MarkRiddoch, line 72: This is debug rather than info information

C/common/storage_client.cpp
```diff
@@ -112,6 +157,38 @@ HttpClient *StorageClient::getHttpClient(void) {
...
+		// m_seqnum_map.clear();
+		
+		m_host = storage_service_host;
+		m_urlbase.str("");
+		m_logger->info("StorageClient::handle_storage_service_restart(): this=%p, m_urlbase=%s", this, m_urlbase.str().c_str());
+		m_urlbase << storage_service_urlbase.str();
+		storage_service_urlbase.str("");
+		storage_service_host = "";
+
+		/* else
```
MarkRiddoch, line 183: If it is not needed please remove

C/common/storage_client.cpp
```diff
@@ -180,6 +258,8 @@ bool StorageClient::readingAppend(const vector<Reading *>& readings)
 		ss << m_pid << "#" << thread_id << "_" << m_seqnum_map[thread_id].load();
 		sto_mtx_client_map.unlock();
 
+		m_logger->info("ss=%s", ss.str().c_str());
```
MarkRiddoch, line 261: This really does not look like useful info for the user, should be debug at best but probably should be removed.

C/common/storage_client.cpp
```diff
@@ -1498,17 +1612,87 @@ void StorageClient::handleException(const exception& ex, const char *operation,
...
 		if (m_management)
 		{
-			// Get a handle on the storage layer
+			PRINT_FUNC;
 			ServiceRecord storageRecord("Fledge Storage");
-			if (!m_management->getService(storageRecord))
+			do
+			{
+				PRINT_FUNC;
+				system("ps -ef | grep fledge.services.storage | grep -v grep | wc -l > /tmp/storage_service_instance_count");
```
MarkRiddoch, line 1627: This is quite ugly, not very portable and makes the assumption the storage service is on the same machine as the client that is using it. We have been very careful not to add that restriction so that we may distribute services across machines or containers at some point.

I suspect you can find out what you need to by just using calls to the API and looking at error codes. A TCP connection to the API port can tell you if it is running or not.

C/services/storage/storage.cpp
```diff
@@ -212,14 +214,14 @@ unsigned short servicePort;
 	}
 	if (config->hasValue("logLevel"))
 	{
-		logger->setMinLevel(config->getValue("logLevel"));
+		//logger->setMinLevel(config->getValue("logLevel"));
```
MarkRiddoch, line 217: Please reinstate

C/services/storage/storage.cpp
```diff
@@ -212,14 +214,14 @@ unsigned short servicePort;
...
+		logger->setMinLevel("info");
 	}
 	else
 	{
-		logger->setMinLevel("warning");
+		logger->setMinLevel("info");
```
MarkRiddoch, line 222: Please set back to warning before checking any code into develop

python/fledge/common/configuration_manager.py
```diff
@@ -33,7 +33,7 @@
 
 import logging
 
-_logger = logger.setup(__name__)
+_logger = logger.setup(__name__, level=logging.DEBUG)
```
MarkRiddoch, line 36: Please remove DEBUG before checking in

python/fledge/common/storage_client/storage_client.py
```diff
@@ -16,14 +16,15 @@
...
+import logging
 from abc import ABC, abstractmethod
 
 from fledge.common import logger
 from fledge.common.service_record import ServiceRecord
 from fledge.common.storage_client.exceptions import *
 from fledge.common.storage_client.utils import Utils
 
-_LOGGER = logger.setup(__name__)
+_LOGGER = logger.setup(__name__, level=logging.DEBUG)
```
MarkRiddoch, line 27: Please reinstate original before checking into develop

python/fledge/common/storage_client/storage_client.py
```diff
@@ -53,11 +54,16 @@ def __init__(self, core_management_host, core_management_port, svc=None):
         try:
             if svc:
                 self.service = svc
+                # _LOGGER.info("StorageClientAsync: __init__: self.service={}".format(self.service))
```
MarkRiddoch, line 57: Please remove commented out code

python/fledge/services/core/api/scheduler.py
```diff
@@ -22,6 +24,7 @@
 __version__ = "${VERSION}"
 
 __DEFAULT_LIMIT = 20
+_logger=logger.setup("core_api_scheduler", level=logging.DEBUG)
```
MarkRiddoch, line 27: Please remove before any merge to develop

python/fledge/services/core/scheduler/scheduler.py
```diff
@@ -135,8 +155,11 @@ def __init__(self, core_management_host=None, core_management_port=None, is_safe
 
         # Initialize class attributes
         if not cls._logger:
-            cls._logger = logger.setup(__name__, level=logging.INFO)
-            # cls._logger = logger.setup(__name__, level=logging.DEBUG)
+            # cls._logger = logger.setup(__name__, level=logging.INFO)
```
MarkRiddoch, line 158: Please reinstate original before merging

python/fledge/services/core/scheduler/scheduler.py
```diff
@@ -439,6 +479,9 @@ async def _check_schedules(self):
             if schedule.exclusive and schedule_execution.task_processes:
                 continue
 
+            # if schedule.type == Schedule.Type.STARTUP and self._scheduler_restart:
```
MarkRiddoch, line 482: Please remove commented out code if it is not required

python/fledge/services/core/server.py
```diff
@@ -56,7 +57,7 @@
 __license__ = "Apache 2.0"
 __version__ = "${VERSION}"
 
-_logger = logger.setup(__name__, level=20)
+_logger = logger.setup(__name__, level=logging.DEBUG)
```
MarkRiddoch, line 60: Please reinstate before merging

python/fledge/services/core/service_registry/monitor.py
```diff
@@ -61,6 +74,34 @@ async def _sleep(self, sleep_time):
     async def _monitor_loop(self):
         """async Monitor loop to monitor registered services"""
         # check health of all micro-services every N seconds
+
+        '''
+        def start_storage():
+            with  open("/tmp/monitor.log", "w+") as file:
```
MarkRiddoch, line 80: What is this file used for? It looks like it is for your debugging purposes and should not be part of the final merge.

python/fledge/services/core/service_registry/monitor.py
```diff
@@ -123,23 +185,61 @@ async def _monitor_loop(self):
...
+                    # self._logger.debug("Resolving pending notification for ACL change "
+                    #                    "for service {} ".format(service_record._name))
                     if not self._acl_handler:
                         self._acl_handler = ACLManager(connect.get_storage_async())
                     await self._acl_handler.\
                         resolve_pending_notification_for_acl_change(service_record._name)
 
                     check_count[service_record._id] = 1
 
+                # self._logger.debug("step D")
```
MarkRiddoch, line 197: Please remove this if it is really not required

MarkRiddoch (changes requested):
Unit tests are failing.

Please remove code that is not really needed.

There is lots of log output at the info level that looks at best to be debug and much that looks like it is not really needed. Merging this will cause a huge increase in logging output at info level that is meant to be a level to give the user more information, not for the code developer.

The changes are impacting the default log levels of various areas of the code and these should be reverted.

The code is littered with PRINT_FUNC, this looks like your debugging and should be removed.

The use of ps in the storage client limits the deployment options to having all services in the same machine or container, this should be avoided.

AmandeepArora:
This is really POC code only.

PR #901, opened 2022-12-14
FOGL-7227 Run purge operations on readings plugins if different from the

MarkRiddoch:
storage plugin.

No one issue that could be improved is that two warnings are issued when doing a reset operations, one for each plugin.


ashwinscale (approved):
LGTM review by test

pintomax (approved):

praveen-garg:
> one issue that could be improved is that two warnings are issued when doing a reset operations, one for each plugin.


@Mohit04tomar ^

PR #900, opened 2022-12-14
FOGL-7179: Set session level time zone for Postgres DB connection to 'UTC'

AmandeepArora:
With this setting, all timezone-aware timestamps are returned in 'UTC' rather than configured time zone.

Signed-off-by: Amandeep Singh Arora <aman@dianomic.com>

PR #902, opened 2022-12-15
FOGL-7179: Request Postgres DB to return UTC timestamps for audit log query

AmandeepArora:
This is an alternate solution specifically for queries to a non-reading table to Postgres DB.

Alternate solution is in https://github.com/fledge-iot/fledge/pull/900.

If PR 900 is acceptable, that might be the better and universal solution, assuming no further application code assumes such retrieved timestamps to be in local timezone and attempts a subsequent conversion to UTC again.

### DIFF
--- a/C/plugins/storage/postgres/connection.cpp
+++ b/C/plugins/storage/postgres/connection.cpp
@@ -443,6 +443,13 @@ SQLBuffer	jsonConstraints;	// Extra constraints to add to where clause
 								sql.append("\"");
 								sql.append((*itr)["column"].GetString());
 								sql.append("\"");
+								Logger::getLogger()->info("itr->HasMember(\"timezone\")=%s", itr->HasMember("timezone")?"True":"False");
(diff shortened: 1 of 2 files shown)


AmandeepArora:
For now, I am not going to fix the python unit tests that are failing in this PR. If we are going to pick this changeset rather than the one in PR-900, then I would fix these UTs.

C/plugins/storage/postgres/connection.cpp
```diff
@@ -443,6 +443,13 @@ SQLBuffer	jsonConstraints;	// Extra constraints to add to where clause
 								sql.append("\"");
 								sql.append((*itr)["column"].GetString());
 								sql.append("\"");
+								Logger::getLogger()->info("itr->HasMember(\"timezone\")=%s", itr->HasMember("timezone")?"True":"False");
```
MarkRiddoch, line 446: I would make this debug if you intend to keep it, certainly not a message that has meaning to a user

AmandeepArora:
Accepting the solution in PR-900, closing this PR.

PR #904, opened 2022-12-16
FOGL-7128: Search empty readings table before creating new database i…

gnandan:
…f reading tables are not available


praveen-garg:
@gnandan This needs conflict resolution. @MarkRiddoch can you please review this?

MarkRiddoch:
We are about to do another release, so, unfortunately I would suggest we hold this until after that release.

C/common/config_category.cpp
```diff
@@ -4,7 +4,7 @@
  * Copyright (c) 2018 Dianomic Systems
  *
  * Released under the Apache 2.0 Licence
- *
+
```
MarkRiddoch, line 7: Please revert this file back to what it was

C/plugins/north/OMF/plugin.cpp
```diff
@@ -2,6 +2,7 @@
  * Fledge PI Server north plugin.
  *
  * Copyright (c) 2018-2022 Dianomic Systems
+
```
MarkRiddoch, line 5: Please revert this file back to what it was

MarkRiddoch (changes requested):
Just a few minor things. Slightly concerned as to thread safety, but nothing to indicate it isn't. Have all the system tests been run against this branch? I think we should before merging this to develop.

MarkRiddoch (approved):

C/plugins/storage/sqlite/common/readings_catalogue.cpp
```diff
@@ -1960,52 +1973,73 @@ ReadingsCatalogue::tyReadingReference  ReadingsCatalogue::getReadingReference(Co
...
+	if (m_EmptyAssetReadingCatalogue.size() == 0)
+	{
+		loadEmptyAssetReadingCatalogue();
+	}
+	
+	auto it = m_EmptyAssetReadingCatalogue.begin();
+	if (it != m_EmptyAssetReadingCatalogue.end())
+	{
+		asset = it->first;
+		emptyTableReference.tableId = it->second.first;
```
Innovative-ashwin, line 2037: if there will be only one value in the map , and iteration over a set of values not require d, then what is the need of map data structure?

gnandan: Purge task keeps purging data on a regular interval, that's why map is being used to track of all the empty table.

Innovative-ashwin (approved):

PR #908, opened 2022-12-21
FOGL-6979 - Allow for 0 rows updated in an update query to be treated as a normal condition

MarkRiddoch:
We need to see if this fixes the issue with FogLAMP Manage deployments


pintomax (approved):

C/plugins/storage/postgres/connection.cpp
```diff
@@ -1286,6 +1287,21 @@ SQLBuffer	sql;
 					return false;
 				}
 			}
+			if (iter->HasMember("modifier") && (*iter)["modifier"].IsArray())
+			{
+				const Value& modifier = (*iter)["modifier"];
+				for (Value::ConstValueIterator modifiers = modifier.Begin(); modifiers != modifier.End(); ++modifiers)
+                		{
+					if (iter->IsString())
```
ashish-jabble, line 1295: `modifiers->`

C/plugins/storage/postgres/connection.cpp
```diff
@@ -1286,6 +1287,21 @@ SQLBuffer	sql;
...
+					{
+						string mod = iter->GetString();
```
ashish-jabble, line 1297: `modifiers->`

ashwinscale (approved):
LGTM review by test

praveen-garg:
@MarkRiddoch SQLite LB plugin is not required to be fixed here?

MarkRiddoch:
> @MarkRiddoch SQLite LB plugin is not required to be fixed here?

Yes it does. It used to be the case that the common directory was shared, hence common. It appears that somewhere along the line this connection between the two was dropped and now they have individual common directories - which is a bit of a farce. The least risky thing to do at this stage is apply the change to the sqlitelb common directory as well.

pintomax (approved):

praveen-garg (approved):

PR #914, opened 2022-12-23
FOGL-7278 Use old OMF types for EDS 

MarkRiddoch:


Mohit04tomar (approved):

C/plugins/north/OMF/plugin.cpp
```diff
@@ -877,6 +877,7 @@ uint32_t plugin_send(const PLUGIN_HANDLE handle,
 		Logger::getLogger()->fatal("OMF Endpoint is not available");
 		return 0;
 	}
+Logger::getLogger()->fatal("FIXME: version is %s", version.c_str());
```
praveen-garg, line 880: Fixed in #920

PR #917, opened 2022-12-23
Add missing packages table to init.sql

MarkRiddoch:


ashish-jabble (commented):
No upgrade is required?
It's better to get them handled in current schema (i.e 59) upgrade sql files as a NOTE

MarkRiddoch:
> No upgrade is required?
> It's better to get them handled in current schema (i.e 59) upgrade sql files as a NOTE

I did consider it, however it is only an issue when you saw from Postgres's to sqlitelb, in which case you will run init.sql. n upgrade script might be nice, but in reality I don't think it would ever be run.

ashish-jabble:
Sure, If upgrade is not a concern then it's fine.

ashish-jabble (approved):

PR #918, opened 2022-12-23
FOGL-7217: initial fix: do reconf in a separate thread

AmandeepArora:
Do reconf in a separate thread, so that service's management interface is not stalled for long in case of long plugin_poll calls


C/services/south/south.cpp
```diff
@@ -547,6 +550,22 @@ void SouthService::start(string& coreAddress, unsigned short corePort)
 						}
 						else // V2 poll method
 						{
+							PRINT_FUNC;
```
MarkRiddoch, line 553: Please remove debug PRINT_FUNC;

C/services/south/south.cpp
```diff
@@ -547,6 +550,22 @@ void SouthService::start(string& coreAddress, unsigned short corePort)
...
+									lock_guard<mutex> guard(m_pendingNewConfigMutex);
+									numPendingReconfs = m_pendingNewConfig.size();
+								}
+								if (numPendingReconfs)
+								{
+									Logger::getLogger()->debug("SouthService::start(): %d entries in m_pendingNewConfig, poll thread yielding CPU", numPendingReconfs);
+									std::this_thread::sleep_for(std::chrono::milliseconds(200));
+								}
+								else
+									break;
```
MarkRiddoch, line 567: Is this the way shutdown gets done in a tidy fashion? If so a comment might be useful.

C/services/south/south.cpp
```diff
@@ -815,7 +834,7 @@ void SouthService::restart()
 /**
  * Configuration change notification
```
MarkRiddoch, line 835: Please provide fill doxygen comment header

C/services/south/south.cpp
```diff
@@ -940,6 +959,86 @@ void SouthService::configChange(const string& categoryName, const string& catego
...
+ */
+static void reconfThreadMain(void *arg)
+{
+	SouthService *ss = (SouthService *)arg;
+	Logger::getLogger()->info("reconfThreadMain(): Spawned new thread for plugin reconf");
+	while(1)
+	{
+		ss->handlePendingReconf();
+	}
+	Logger::getLogger()->info("reconfThreadMain(): plugin reconf thread exiting");
```
MarkRiddoch, line 974: And how does this ever happen?

C/services/south/south.cpp
```diff
@@ -940,6 +959,86 @@ void SouthService::configChange(const string& categoryName, const string& catego
...
+}
+
+/**
+ * Handle configuration change notification
```
MarkRiddoch, line 978: Add some useful detail here. Called by the reconfigure thread?

C/services/south/south.cpp
```diff
@@ -940,6 +959,86 @@ void SouthService::configChange(const string& categoryName, const string& catego
...
+ */
+void SouthService::handlePendingReconf()
+{
+	Logger::getLogger()->debug("SouthService::handlePendingReconf: Going into cv wait");
+	mutex mtx;
+	unique_lock<mutex> lck(mtx);
+	m_cvNewReconf.wait(lck);
+	Logger::getLogger()->debug("SouthService::handlePendingReconf: cv wait has completed");
+
+	while(1)
```
MarkRiddoch, line 988: It's not clear what calls what and what is waiting and how the waits are interrupted for reconfiguration or shutdown.

C/services/south/south.cpp
```diff
@@ -940,6 +959,86 @@ void SouthService::configChange(const string& categoryName, const string& catego
...
+			if (numPendingReconfs)
+				Logger::getLogger()->debug("SouthService::handlePendingReconf(): will process %d entries in m_pendingNewConfig", numPendingReconfs);
+			else
+			{
+				Logger::getLogger()->debug("SouthService::handlePendingReconf DONE");
+				break;
+			}
+		}
+
+		for (unsigned int i=0; i<numPendingReconfs; i++)
```
MarkRiddoch, line 1003: If you have more than one pending reconfiguration you don't need to worry about it, only the last one. Each reconfiguration request comes with a complete configuration category.

C/services/south/south.cpp
```diff
@@ -940,6 +959,86 @@ void SouthService::configChange(const string& categoryName, const string& catego
...
+		}
+	}
+}
+
+/**
+ * Configuration change notification using a separate thread
+ */
+void SouthService::configChange(const string& categoryName, const string& category)
+{
+	PRINT_FUNC;
```
MarkRiddoch, line 1032: Please remove debug

C/services/south/south.cpp
```diff
@@ -815,7 +834,7 @@ void SouthService::restart()
 /**
  * Configuration change notification
  */
-void SouthService::configChange(const string& categoryName, const string& category)
+void SouthService::configChangeReal(const string& categoryName, const string& category)
```
MarkRiddoch, line 837: I really do not like the naming of this.

MarkRiddoch (changes requested):

C/services/south/south.cpp
AmandeepArora: At south service shutdown only. Just indicates that this thread exited cleanly.

C/services/south/south.cpp
AmandeepArora: New reconf requests are queued up in SouthService::configChange() and m_cvNewReconf.notify_all() gets called to wake up the reconf thread's function SouthService::handlePendingReconf() and it processes all queued up reconf requests.

C/services/south/south.cpp
AmandeepArora: I am thinking that sometimes reconf could have side effects and hence it is probably better to process all reconf requests in order. e.g. a device register value set from 1->2->3 might indicate something special, etc.

C/services/south/south.cpp
AmandeepArora: Even now, service shutdown gets done eventually in case of service disable, but may get pended by a few plugin_poll calls. But at foglamp shutdown, it seems services get just 10 seconds for shutdown, else their process is killed by core, which doesn't seem right.

C/services/south/south.cpp
AmandeepArora: Changed.

AmandeepArora:
There is another PR for the plugin code: https://github.com/dianomic/foglamp-south-suez/pull/14

MarkRiddoch (approved):

C/services/south/south.cpp
AmandeepArora: It is not required to terminate, that's all this thread does. It just waits for new reconf operations. It waits for m_cvNewReconf condition variable inside handlePendingReconf() before it processes all pending reconf requests at that time.

PR #919, opened 2022-12-23
FOGL-7285 System tests for PIWebAPI are failing with PIWebAPI 2021 sp2.

Mohit04tomar:
Signed-off-by: Mohit Tomar <m.singh19994@gmail.com>

### DIFF
--- a/tests/system/python/conftest.py
+++ b/tests/system/python/conftest.py
@@ -638,8 +638,8 @@ def pytest_addoption(parser):
     parser.addoption("--pi-token", action="store", default="omf_north_0001",
                      help="OMF Producer Token")
     parser.addoption("--pi-use-legacy", action="store", default="true",
-                     help="Set true if you want to send linked data types otherwise set true.")
-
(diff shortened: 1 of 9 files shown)


tests/system/python/conftest.py
```diff
@@ -672,6 +637,8 @@ def pytest_addoption(parser):
                      help="PI Server user login password")
     parser.addoption("--pi-token", action="store", default="omf_north_0001",
                      help="OMF Producer Token")
+    parser.addoption("--pi-use-legacy", action="store", default="true",
+                     help="Set true if you want to send linked data types otherwise set true.")
```
Mohit04tomar, line 641: Set false if you want to send linked data types otherwise set true.

praveen-garg: Set false to override the default plugin behaviour i.e. for OMF version >=1.2.x to send linked data types

praveen-garg (approved):

PR #922, opened 2022-12-27
FOGL-7260 - CentOS Stream9 fixes in core & their system tests

ashish-jabble:


python/fledge/common/utils.py
```diff
@@ -85,3 +85,42 @@ def decorator(Class):
...
+    """ General information to identifying the operating system """
+    import ast
+    import re
+    os_details = {}
+    try:
+        filename = '/etc/os-release'
+        f = open(filename)
+    except FileNotFoundError:
+        filename = '/usr/lib/os-release'
+        f = open(filename)
```
praveen-garg, line 100: use with context or close after reading?

python/fledge/common/utils.py
```diff
@@ -85,3 +85,42 @@ def decorator(Class):
...
+
+    for line_number, line in enumerate(f, start=1):
+        line = line.rstrip()
+        if not line or line.startswith('#'):
+            continue
+        m = re.match(r'([A-Z][A-Z_0-9]+)=(.*)', line)
```
praveen-garg, line 106: add comment for regex explaination

python/fledge/common/utils.py
```diff
@@ -85,3 +85,42 @@ def decorator(Class):
...
+            name, val = m.groups()
+            if val and val[0] in '"\'':
+                val = ast.literal_eval(val)
+            os_details.update({name: val})
+    return os_details
+
+
+def is_debian():
+    """
+        To check if the Operating system is of Debian type or Not
```
praveen-garg, line 117: of Debian type or Not
Debian based or Not

Actually we are checking if Redhat family or not

praveen-garg (commented):
minor comments otherwise lgtm!

python/fledge/common/utils.py
praveen-garg: is_redhat_based/is_redhat_family will keep it more cleaner along with existing changed for yum vs apt.

e.g. 
`pkg_mgt = 'yum' if is_redhat_based() else 'apt'`

python/fledge/services/core/api/update.py
```diff
@@ -126,10 +125,10 @@ async def get_updates(request: web.Request) -> web.Response:
...
          curl -sX GET http://localhost:8081/fledge/update |jq
     """
-    _platform = platform.platform()
-    update_cmd = "sudo apt update"
-    upgradable_pkgs_check_cmd = "apt list --upgradable | grep \^fledge"
-    if "centos" in _platform or "redhat" in _platform:
+    if utils.is_debian():
+        update_cmd = "sudo apt update"
+        upgradable_pkgs_check_cmd = "apt list --upgradable | grep \^fledge"
+    else:
```
praveen-garg, line 131: that way we can also keep here default as is; and  only `"centos" in _platform or "redhat" in _platform` with if is replaced.

praveen-garg (approved):

MarkRiddoch (approved):

PR #923, opened 2022-12-27
FOGL-7298 PI system Connector relay OMF version fix

deepanshu-yadav:
Hi @MarkRiddoch 

We no longer get the following error message with this pull request. 
```
 {"code":7,"message":"Invalid message headers.","source":"omfversion","description":"Header omfversion has invalid value: 1.2. Expected value: 1.0"} - 10.0.0.95:5460 /ingress/messages
```

However at the Connector relay side we can see the elements and it attributes but no PI points are being created for the element . 

Also we need to set Legacy or complex data type = true in order to send data to connector relay. 
cc @ashwinscale


C/plugins/north/OMF/plugin.cpp
```diff
@@ -980,6 +980,15 @@ uint32_t plugin_send(const PLUGIN_HANDLE handle,
 		connInfo->omfversion = EDS_OMF_VERSION;
 	}
 #endif
+
+// Version for Connector Relay is 1.0 only.
+#ifdef CR_OMF_VERSION
```
MarkRiddoch, line 985: I don't believe the connector relay ever went beyond OMF 1.0, so we probably do not need to have the ifdef here.

MarkRiddoch (changes requested):
Force the legacy type support for CR endpoints as well. It should not be left to the user to do this.

praveen-garg:
```
if (version.find("2019") != std::string::npos)
	{
		connInfo->omfversion = "1.0";
	}
	else if (version.find("2020") != std::string::npos)
	{
		connInfo->omfversion = "1.1";
	}
```

@MarkRiddoch @deepanshu-yadav  Should we override with a INFO/WARN to legacy true? So that existing users does not need see failure on upgrade? /cc @ashwinscale

deepanshu-yadav:
> ```
> if (version.find("2019") != std::string::npos)
> 	{
> 		connInfo->omfversion = "1.0";
> 	}
> 	else if (version.find("2020") != std::string::npos)
> 	{
> 		connInfo->omfversion = "1.1";
> 	}
> ```
> 
> @MarkRiddoch @deepanshu-yadav Should we override with a INFO/WARN to legacy true? So that existing users does not need see failure on upgrade? /cc @ashwinscale

Done!

praveen-garg (dismissed):

C/plugins/north/OMF/plugin.cpp
deepanshu-yadav: Done

deepanshu-yadav:
> Force the legacy type support for CR endpoints as well. It should not be left to the user to do this.

Done

praveen-garg (approved):

praveen-garg:
Tested with PI Web API 2019, upgrade works okay. No user intervention required to fix LegacyType.

MarkRiddoch (approved):

PR #924, opened 2022-12-27
FOGL-7222 - Authentication system API tests & other non -notification tests fixes on CentOS9 stream

ashish-jabble:
`fledge_nightly_build_centos_stream_9/27/` - Now those tests are passing

Note: Tests file will be reverted in FOGL-7260 branch & other TODO comments for x509 strict check

### DIFF
--- a/python/fledge/common/web/ssl_wrapper.py
+++ b/python/fledge/common/web/ssl_wrapper.py
@@ -9,7 +9,7 @@
 import time
 import datetime
 import subprocess
-from fledge.common import logger
+from fledge.common import logger, utils
 
 __author__ = "Amarendra Kumar Sinha"
 __copyright__ = "Copyright (c) 2019 Dianomic Systems"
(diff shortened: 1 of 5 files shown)


python/fledge/common/web/ssl_wrapper.py
```diff
@@ -80,21 +80,27 @@ def verify_against_revoked(cls):
     @classmethod
     def verify_against_ca(cls):
         echo_process = subprocess.Popen(['echo', cls.user_cert], stderr=subprocess.PIPE, stdout=subprocess.PIPE)
-        a = subprocess.Popen(["openssl", "verify", "-CAfile", cls.ca_cert, "-x509_strict"], stdin=echo_process.stdout, stderr=subprocess.PIPE, stdout=subprocess.PIPE)
+        args = ["openssl", "verify", "-CAfile", cls.ca_cert]
+        cmd = "openssl verify -CAfile {}".format(cls.ca_cert)
```
praveen-garg, line 84: We can avoid this and use ' '.join(args) where required

python/fledge/common/web/ssl_wrapper.py
```diff
@@ -80,21 +80,27 @@ def verify_against_revoked(cls):
...
+        # TODO: Add JIRA for to handle x509 strict check when open ssl version is 3 or so
+        # This is completely a workaround as of now CentOS9 stream has openssl version with 3.0
+        if utils.get_open_ssl_version(version_string=False)[0] < 3:
+            x509_strict = "-x509_strict"
+            args.append(x509_strict)
```
praveen-garg, line 89: can directly append instead of having extra var

python/fledge/common/web/ssl_wrapper.py
```diff
@@ -80,21 +80,27 @@ def verify_against_revoked(cls):
     @classmethod
     def verify_against_ca(cls):
         echo_process = subprocess.Popen(['echo', cls.user_cert], stderr=subprocess.PIPE, stdout=subprocess.PIPE)
-        a = subprocess.Popen(["openssl", "verify", "-CAfile", cls.ca_cert, "-x509_strict"], stdin=echo_process.stdout, stderr=subprocess.PIPE, stdout=subprocess.PIPE)
+        args = ["openssl", "verify", "-CAfile", cls.ca_cert]
+        cmd = "openssl verify -CAfile {}".format(cls.ca_cert)
+        # TODO: Add JIRA for to handle x509 strict check when open ssl version is 3 or so
+        # This is completely a workaround as of now CentOS9 stream has openssl version with 3.0
```
praveen-garg, line 86: ; as of now CentOS Stream 9 has openssl version 3.0

python/fledge/common/web/ssl_wrapper.py
```diff
@@ -80,21 +80,27 @@ def verify_against_revoked(cls):
     @classmethod
     def verify_against_ca(cls):
         echo_process = subprocess.Popen(['echo', cls.user_cert], stderr=subprocess.PIPE, stdout=subprocess.PIPE)
-        a = subprocess.Popen(["openssl", "verify", "-CAfile", cls.ca_cert, "-x509_strict"], stdin=echo_process.stdout, stderr=subprocess.PIPE, stdout=subprocess.PIPE)
+        args = ["openssl", "verify", "-CAfile", cls.ca_cert]
+        cmd = "openssl verify -CAfile {}".format(cls.ca_cert)
+        # TODO: Add JIRA for to handle x509 strict check when open ssl version is 3 or so
```
praveen-garg, line 85: for to -> for handling/ to handle ... when OpenSSL version is >= 3.x

praveen-garg (commented):
This is good to go as an interim solution; as missing X509 cert attributes are not critical for self-signed certificates.

python/fledge/common/web/ssl_wrapper.py
ashish-jabble: When joining strings Python only has to allocate new memory for the final string, which is much more efficient, but could take longer to compute.
For example foo += 'bar' is just as readable (according to me) and more faster than foo = ''.join([foo, 'bar']).

python/fledge/common/web/ssl_wrapper.py
ashish-jabble: This variable is also used in `cmd` for logging purpose.

python/fledge/common/web/ssl_wrapper.py
ashish-jabble: Will address in main branch

python/fledge/common/web/ssl_wrapper.py
ashish-jabble: Will address in main branch

PR #925, opened 2022-12-28
FOGL-7299: Draft Solution for providing option to increase and decrea…

gnandan:
…se maximum attached database for sqlite plugin


gnandan:
Also we need to provide compiled sqlite3 libs with package, then only this will work

praveen-garg:
@ashish-jabble Please do basic testing if this allows storage tuning?

requirements.sh
```diff
@@ -110,7 +110,7 @@ sqlite3_build_prepare(){
 	cd ${SQLITE_PKG_REPO_NAME}
 	cd src
 	echo "Compiling SQLite3 static library for Fledge ..."
-	./configure --enable-shared=false --enable-static=true --enable-static-shell CFLAGS="-DSQLITE_MAX_COMPOUND_SELECT=900 -DSQLITE_MAX_ATTACHED=62 -DSQLITE_ENABLE_JSON1 -DSQLITE_ENABLE_LOAD_EXTENSION -DSQLITE_ENABLE_COLUMN_METADATA -fno-common -fPIC"
+	./configure --enable-static-shell CFLAGS="-DSQLITE_MAX_COMPOUND_SELECT=900 -DSQLITE_MAX_ATTACHED=125 -DSQLITE_ENABLE_JSON1 -DSQLITE_ENABLE_LOAD_EXTENSION -DSQLITE_ENABLE_COLUMN_METADATA -fno-common -fPIC"
```
MarkRiddoch, line 113: The previous value had been very carefully chosen given the number of file descriptors needed for each database attached (3), the number of default threads and the number of per process file descriptors that are available. This new value will blow the file descriptor limit and will fail if the descriptors are used.

MarkRiddoch (changes requested):
You are not actually doing anything with the limit that is configured. The more serious issue is running out of file descriptors, which in practice happens before the current number of database that can be attached is reached. 

What has prompted you to do this change?

gnandan:
> You are not actually doing anything with the limit that is configured. The more serious issue is running out of file descriptors, which in practice happens before the current number of database that can be attached is reached.
> 
> What has prompted you to do this change?

HI Mark,
Actually, requirements.sh was just building the sqlite3 package but not really installing it (No 'make install' is there).  So even after configuring SQLITE_MAX_ATTACHED=62 actual value was the default one that's is 10 databases out of which we only have 7 databases for readings because 3 databases  'tmp' ,'main' and fledge.db are always there so readings can use only 7 Databases.  For FOGL-7128 : I was able to successfully reuse empty tables but I was only able to use upto 7 reading tables after that I was getting error for not attaching more than 10 database. That's why I changed it to attach more databases.  If we are ok  to use just 7 readings DBs at most then FOGL-7128 will work fine along with reuse of empty tables

Also I made Attached DB Limit configurable so that user can increase or decrease this limit if required using fledge GUI, Default value is still 10  DBs so no change in that. We can change SQLITE_MAX_ATTACHED=62  again, however, we should have the ways to use it.

praveen-garg:
Hi @MarkRiddoch Could you please check [the reply](https://github.com/fledge-iot/fledge/pull/925#issuecomment-1371797064) from @gnandan

PR #933, opened 2023-01-12
FOGL-7279 Add display name and deprecated support to configuration ca…

MarkRiddoch:
…tegory get/set item attribute

### DIFF
--- a/C/common/include/config_category.h
+++ b/C/common/include/config_category.h
@@ -126,7 +126,9 @@ class ConfigCategory {
 					MAXIMUM_ATTR,
 					LENGTH_ATTR,
 					VALIDITY_ATTR,
-					GROUP_ATTR};
+					GROUP_ATTR,
+					DISPLAY_NAME_ATTR,
+					DEPRECATED_ATTR};
 		std::string			getItemAttribute(const std::string& itemName,
 								 ItemAttribute itemAttribute) const;
 
@@ -173,7 +175,7 @@ class ConfigCategory {
 		};
 		std::vector<CategoryItem *>	m_items;
 		std::string			m_name;
-		std::string         m_parent_name;
+		std::string         		m_parent_name;
 		std::string			m_description;
 		std::string			m_displayName;
 
--- a/C/common/include/config_category.h
+++ b/C/common/include/config_category.h
@@ -128,7 +128,8 @@ class ConfigCategory {
 					VALIDITY_ATTR,
 					GROUP_ATTR,
 					DISPLAY_NAME_ATTR,
-					DEPRECATED_ATTR};
+					DEPRECATED_ATTR,
+					RULE_ATTR};
 		std::string			getItemAttribute(const std::string& itemName,
 								 ItemAttribute itemAttribute) const;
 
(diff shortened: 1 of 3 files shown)


C/common/include/config_category.h
```diff
@@ -126,7 +126,9 @@ class ConfigCategory {
 					MAXIMUM_ATTR,
 					LENGTH_ATTR,
 					VALIDITY_ATTR,
-					GROUP_ATTR};
+					GROUP_ATTR,
+					DISPLAY_NAME_ATTR,
+					DEPRECATED_ATTR};
```
ashish-jabble, line 131: What about `rule` optional config item?

MarkRiddoch: When you say rule do you mean the validity rule? If so this is already covered as VALIDITY_ATTR.

ashish-jabble: We have supported `rule` optional attribute as well. Validity optional attribute is different from it.
**rule**
```
The validation rule will take the form of an expression. If it evaluates to true then the proposed value of the configuration item will be allowed. If it evaluates to false then it will not be.
```

**validity**
```
An expression used to determine if the configuration item is valid. Used in the UI to gray out one value based on the value of others.
```

We have also added its documentation https://fledge-iot.readthedocs.io/en/develop/plugin_developers_guide/02_writing_plugins.html#configuration-lifecycle

tests/unit/C/common/test_config_category.cpp
```diff
@@ -309,6 +309,17 @@ const char* bigCategory =
 			"\"description\": \"Defines\"} " \
 		"}";
 
+const char *optionals = 
+	"{\"item1\" : { "\
+			"\"type\": \"integer\", \"displayName\": \"Item1\", " \
+			"\"value\": \"3\", \"default\": \"3\", " \
+			"\"description\": \"First Item\", " \
+			"\"group\" : \"Group1\", " \
+			"\"rule\" : \"1 = 0\", " \
```
ashish-jabble, line 318: `==`

MarkRiddoch: Sorry, I don't understand your comment?

ashish-jabble: Example:  IN *optionals* const char, To test that a proposed value is even a rule of the form

`"rule" : "value % 4 == 3"` OR `"rule" : "value  == 3"`

The rule will be executed as part of the validation process in the configuration manager.
Note, if other validation is called for, e.g. minimum or maximum value, then these will be run in addition to the custom rule.

So meaning that rule expression only apply/checks with **value** KV pair. Category won't be created/updated if it does not meet the rule.

MarkRiddoch: It really doesn't matter, the test is simply to test if you can retrieve the rule. The code being tested does not look at the content.

ashish-jabble: That's absolutely correct, but it always be better to have clear test data.

ashish-jabble (approved):

PR #941, opened 2023-01-19
Optimisation for adding milliseconds to timestamps

MarkRiddoch:

Aman, it struck me this morning that given your optimisation for caching the seconds we could; also improve on the sub-second portion by avoiding sprint, that is known to be slow, for both the number conversation and the string concatenation. Also in the FMT_ISO8601MS case doing the conversion to the C++ string is slow.

New verses original times with this code for the abstracted portion
```
original test1: 10000 iterations took 0.002458 seconds
original test2: 10000 iterations took 0.005415 seconds
     new test1: 10000 iterations took 0.001551 seconds
     new test2: 10000 iterations took 0.001933 seconds
```

### DIFF
--- a/C/common/reading.cpp
+++ b/C/common/reading.cpp
@@ -287,24 +287,34 @@ char  assetTime[DATE_TIME_BUFFER_LEN + 20];
 	if (dateFormat != FMT_ISO8601 && addMS)
 	{
 		// Add microseconds
-		snprintf(micro_s,
-			 sizeof(micro_s),
-			 ".%06lu",
-			 m_timestamp.tv_usec);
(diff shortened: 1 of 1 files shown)


AmandeepArora (commented):
Hi Mark, I tried something similar in all those 3 places. I was not able to get a reliable/significant speedup with that. My initial changeset improved timings from 950-1050 ms to around 400 ms. But with/without this change the timings remained around 400 ms only when averaged over processing of 110 million readings across 6 python filters. The difference was an average 406 ms with snprintf vs 402 ms without snprintf.

AmandeepArora (commented):
Otherwise changes look good.

pintomax (approved):

PR #953, opened 2023-01-30
FOGL-7375 : Modified to return default value as blank for type:script

gnandan:

### DIFF
--- a/C/common/config_category.cpp
+++ b/C/common/config_category.cpp
@@ -1566,10 +1566,20 @@ ostringstream convert;
 	else if (m_itemType == JsonItem ||
 		 m_itemType == NumberItem ||
 		 m_itemType == DoubleItem ||
-		 m_itemType == ScriptItem ||
 		 m_itemType == CodeItem)
 	{
 		convert << ", \"default\" : \"" << JSONescape(m_default) << "\" }";
+	}
+	else if ( m_itemType == ScriptItem )
+	{
+		if (m_default == "\"\"" )
+		{
+			convert << ", \"default\" : \"\" }";
+		}
+		else
+		{
+			convert << ", \"default\" : \"" << JSONescape(m_default) << "\" }";
+		}
 	}
 	return convert.str();
 }


MarkRiddoch (changes requested):
No, this is not what is needed. The default should not be hard coded in the config  category class. The filter sets its default to the empty string. Please find out what is adding the extra quotes and fix the problem where it is broken.

gnandan:
> No, this is not what is needed. The default should not be hard coded in the config category class. The filter sets its default to the empty string. Please find out what is adding the extra quotes and fix the problem where it is broken.

Hi Mark,

Constructor of  CategoryItem itself is adding extra quotes in case of blank values to default.

Below are the details form config_category.cpp file at line number# 1267

ConfigCategory::CategoryItem::CategoryItem(const string& name,
					   const Value& item)
{
....
if (m_default.empty())
			{
				m_default = "\"\""; <=== Line number 1267 of config_category.cpp
			}
.....
}

So it's ok to modify constructor itself ?

MarkRiddoch:
> 

Fixing the constructor would be the right place to fix this.

gnandan:
> > 
> 
> Fixing the constructor would be the right place to fix this.

It is done

gnandan:
> 

Reverted changes as fixing in constructor failing during reconfigure because JSON is having blank value for default instead of quotes 
```
Value after fix ==> "default" : 
expected value  ==>  "default": "" 
```

MarkRiddoch:
> 

We do not have to do this for any other configuration type, why do we need to do this for scripts? Where did it fail when you fixed the constructor?

gnandan:
> 



> We do not have to do this for any other configuration type, why do we need to do this for scripts? Where did it fail when you fixed the constructor?

Well if  config item is of type script / code then constructor explicitly sets it to quotes rather 
```
if (m_itemType == ScriptItem ||
		    m_itemType == CodeItem)
		{
			// Get content of script type item as is
			rapidjson::StringBuffer strbuf;
			rapidjson::Writer<rapidjson::StringBuffer> writer(strbuf);
			item["default"].Accept(writer);
			if (m_default.empty())
			{
				m_default = "\"\"";
			}
		}
```

If I change m_default = "" then final value in JSON is shown as 
```
"default": <blank space> instead of "default" : ""
```
So  it gets crashed due to Invalid JSON. That's why in case of script/code we have should not use escape character sequence " "\" otherwise in final JSON default will like "default" :  "\"\"" 

So instead of blank it translates to quotes and GUI get quotes as a value of  length 2 instead of empty value of length zero. That's why GUI fails here.

Let me know further action on this.

PR #958, opened 2023-02-02
FOGL-7343 Document new user roles

MarkRiddoch:


docs/securing_fledge.rst
```diff
@@ -130,7 +131,12 @@ Whenever a user logs into Fledge the age of their password is checked against th
 User Management
 ===============
 
-Once mandatory authentication has been enabled and the currently logged in user has the role *admin*, a new option appears in the GUI, *User Management*.
+The user management option becomes active once the Fledge has been configured to require authentication of users. This is enabled via the *Admin API* page of the *Configuration* menu item. A new menu item *User Management* will appear in the left hand menu.
+
+.. note::
+
+   After setting the Authentication option to Mandatory in the configuration page the Fledge instance should be restarted.
```
ashish-jabble, line 138: `mandatory` - lowercase

docs/securing_fledge.rst
```diff
@@ -142,11 +148,17 @@ The user management pages allows
   - Deleting users.
   - Resetting user passwords.
   - Changing the role of a user.
+  - Changing the details of a user
+
+Fledge currently supports four roles for users:
+
+  - **Administrator**: a user with admin role is able to fully configure Fledge, view the data read by the Fledge instance  and also manage Fledge users
```
ashish-jabble, line 155: Add full stop to end a sentence.

docs/securing_fledge.rst
```diff
@@ -142,11 +148,17 @@ The user management pages allows
...
 
-Fledge currently supports two roles for users,
+  - **Editor**: a user with this role is able to configure Fledge and view the data read by Fledge. The user can not manage other users or add new users.
```
ashish-jabble, line 157: `A user` - caps (A)

docs/securing_fledge.rst
```diff
@@ -142,11 +148,17 @@ The user management pages allows
...
 
-  - **admin**: a user with admin role is able to fully configure Fledge and also manage Fledge users
-  - **user**: a user with this role is able to configure Fledge but can not manage users
+  - **Viewer**: a user that can only view the configuration of the Fledge instance. The user has no ability to modify the Fledge instance in any way.
```
ashish-jabble, line 159: > view the configuration

It pretends to be only to view the configuration, though with this type of role ALL read access operations are allowed and can be viewed.

docs/securing_fledge.rst
```diff
@@ -142,11 +148,17 @@ The user management pages allows
...
+
+  - **Data Viewer**: a user that can only view the data in Fledge and not the configuration of Fledge itself. The user has no ability to modify the Fledge instance in any way.
```
ashish-jabble, line 161: > view the data

We should tell to users - From GUI tabs like `Assets & Readings, Dashboards etc.` can be viewed.

docs/securing_fledge.rst
```diff
@@ -159,6 +171,15 @@ To add a new user from the *User Management* page select the *Add User* icon in
 
 You can select a role for the new user, a user name and an initial password for the user. Only users with the role *admin* can add new users.
 
+Update User Details
+-------------------
+
+The edit user option allows the name, authentication method and description of a user to be updated.
```
ashish-jabble, line 177: This option is only available for `admin` role types.

And the User with the other roles can update their profile ( only allowed fields are - `Name, Password`)

docs/securing_fledge.rst
MarkRiddoch: I would rather not lit specific items that are or are not available as it will need constant changing as we add features.

docs/securing_fledge.rst
MarkRiddoch: We don't do this anywhere else when using hanging ident tags like this.

docs/securing_fledge.rst
ashish-jabble: Sure

PR #960, opened 2023-02-03
FOGL-7353 : Deep copy of reading set class

gnandan:


C/common/reading_set.cpp
```diff
@@ -192,6 +192,45 @@ ReadingSet::append(const vector<Reading *>& readings)
...
+			   Reading *in = new Reading(assetName, value);
+			   readings->push_back(in);
+		   }
+	   }
+   }
+   catch(...)
+   {
+	   copyResult = false;
+	   readings->clear();
+	   Logger::getLogger()->error("Copy opeation failed");
```
MarkRiddoch, line 222: It woudl be good to have a little more inforamation. Imagine you are looking at a problem reported by a customer and you see "Copy operation failed" in the log - would it help you find the problem? You do not not what object its s trying to copy? Did it fail because it ran out of memory or was there a problem wit the thing it was copying?

It is always wise to look at you error messages and imagine if it would mean anything to a random engineer 12 months from now.

MarkRiddoch (changes requested):
Looks fine other than I think you need better error reporting

C/common/reading_set.cpp
RVerhoeff: There is a typo in that error message: opeation -> operation.

C/common/reading_set.cpp
```diff
@@ -192,6 +192,51 @@ ReadingSet::append(const vector<Reading *>& readings)
...
+			   Reading *in = new Reading(assetName, value);
+			   readings->push_back(in);
+		   }
+	   }
+   }
+   catch (std::bad_alloc& ex)
+   {
+		copyResult = false;
+		readings->clear();
+		Logger::getLogger()->error("Insufficient memory :: failed while copying :%d: reading from ReadingSet :%s: ",readings->size()+1, ex.what());
```
MarkRiddoch, line 222: You don't need the +1, size() returns the number of items in the container. Also please loose the extra :'s you are using to surround things. Error level messages are meant for users, so should be well formatted statements.

C/common/reading_set.cpp
```diff
@@ -192,6 +192,51 @@ ReadingSet::append(const vector<Reading *>& readings)
...
+   }
+   catch (std::exception& ex)
+   {
+	   copyResult = false;
+	   readings->clear();
+	   Logger::getLogger()->error("Unknown exception :: failed while copying :%d: reading from ReadingSet :%s: ",readings->size()+1, ex.what());
```
MarkRiddoch, line 228: As above, the +1 is not required and formatting of the message for the user. I am currently putting together guidance/rules for errors, I'll send you a link to the draft.

C/common/reading_set.cpp
```diff
@@ -192,6 +192,51 @@ ReadingSet::append(const vector<Reading *>& readings)
 	}
 }
 
+/**
+* Deep copy a set of readings to this reading set.
+*/
+bool
+ReadingSet::copy(const ReadingSet& src)
+{
+   vector<Reading *> *readings = new vector<Reading *>;
```
MarkRiddoch, line 201: This is never freed as is a memory leak, you can simply create this on the stack and then pass it to append

MarkRiddoch (changes requested):

C/common/reading_set.cpp
gnandan: Yes, size() returns the number but +1 will give current items which was being copied when exception occurred and not  collected so far into readings vector.  does it make correct use case to include +1 with size() ?

C/common/reading_set.cpp
```diff
@@ -192,6 +192,51 @@ ReadingSet::append(const vector<Reading *>& readings)
...
+	   for ( auto rs : src.getAllReadings())
+	   {
+		   std::string assetName = rs->getAssetName();
+		   for ( auto dp : rs->getReadingData())
+		   {
+			   std::string dataPointName  = dp->getName();
+			   DatapointValue dv = dp->getData();
+			   Datapoint *value = new Datapoint(dataPointName, dv);
+			   Reading *in = new Reading(assetName, value);
+			   readings->push_back(in);
```
AmandeepArora, line 214: Probably better to use emplace_back instead.

C/common/reading_set.cpp
```diff
@@ -192,6 +192,51 @@ ReadingSet::append(const vector<Reading *>& readings)
...
+   {
+	   for ( auto rs : src.getAllReadings())
+	   {
+		   std::string assetName = rs->getAssetName();
+		   for ( auto dp : rs->getReadingData())
+		   {
+			   std::string dataPointName  = dp->getName();
+			   DatapointValue dv = dp->getData();
+			   Datapoint *value = new Datapoint(dataPointName, dv);
+			   Reading *in = new Reading(assetName, value);
```
AmandeepArora, line 213: Reading object should be created just once, above this DP loop.

AmandeepArora (approved):

AmandeepArora (approved):

C/common/reading_set.cpp
```diff
@@ -192,6 +192,55 @@ ReadingSet::append(const vector<Reading *>& readings)
...
+		   std::vector<Datapoint *> dataPoints;
+		   // Iterate over all the datapoints associated with one reading
+		   for ( auto &dp : reading->getReadingData())
+		   {
+			   std::string dataPointName  = dp->getName();
+			   DatapointValue dv = dp->getData();
+			   dataPoints.emplace_back(new Datapoint(dataPointName, dv));
+			  
+		   }
+		   Reading *in = new Reading(assetName, dataPoints);
```
Innovative-ashwin, line 218: the memory objects created on heap, like "in" and "datapoints members , make sure they are deleted correctly.

C/common/reading_set.cpp
```diff
@@ -192,6 +192,55 @@ ReadingSet::append(const vector<Reading *>& readings)
...
+		   readings.emplace_back(in);
+	   }
+   }
+   catch (std::bad_alloc& ex)
+   {
+		Logger::getLogger()->error("Insufficient memory, failed while copying %d reading from ReadingSet, %s ",readings.size()+1, ex.what());
+		copyResult = false;
```
Innovative-ashwin, line 225: memory is not correctly handled , what if creation of "in" throwss error, datapoints objects do not get deleted properly.

C/common/reading_set.cpp
```diff
@@ -192,6 +192,55 @@ ReadingSet::append(const vector<Reading *>& readings)
...
+		readings.clear();
+   }
+   catch (std::exception& ex)
+   {
+	   Logger::getLogger()->error("Unknown exception, failed while copying %d reading from ReadingSet, %s ",readings.size()+1, ex.what());
+	   copyResult = false;
```
Innovative-ashwin, line 231: same here, memory on heap handling

C/common/reading_set.cpp
gnandan: We are creating new memory to copy the existing datapoints , hence these data points are not supposed to be deleted

C/common/reading_set.cpp
gnandan: Vector of readings is created on stack so it will taken care during stack unwinding, No need to delete separately

C/common/reading_set.cpp
gnandan: Vector of readings is created on stack so it will taken care during stack unwinding, No need to delete separately

RVerhoeff: If readings were defined as _vector<Reading>_, that would be true. This is _vector<Reading *>_. You are normally responsible for deleting the objects pointed to by the vector elements. I don't think calling _emplace_back()_ on a pointer changes that.

tests/unit/C/services/core/reading_set_copy.cpp
```diff
@@ -0,0 +1,173 @@
...
+TEST(READINGSET, DeepCopyCheckReadingFromNestedJSON)
+{
+    ReadingSet *readingSet1 = new ReadingSet(NestedReadingJSON);
+    ReadingSet *readingSet2 = new ReadingSet();
+    readingSet2->copy(*readingSet1);
+
+    auto r1 = readingSet1->getAllReadings();
+    auto dp1 = r1[0]->getReadingData();
+
+    // Fetch neted datapoints
```
RVerhoeff, line 53: typo: Fetch nested datapoints

tests/unit/C/services/core/reading_set_copy.cpp
```diff
@@ -0,0 +1,173 @@
...
+    ASSERT_EQ(dp1[1]->getData().toString(), "{\"floor1\":30, \"floor2\":34, \"floor3\":36}");
+
+    auto r2 = readingSet2->getAllReadings();
+    auto dp2 = r2[0]->getReadingData();
+    ASSERT_EQ(dp2[0]->getName(), "degrees");
+    ASSERT_EQ(dp2[0]->getData().toString(), "[200.65, 34.45, 500.36]");
+    ASSERT_EQ(dp2[1]->getName(), "pressure");
+    ASSERT_EQ(dp2[1]->getData().toString(), "{\"floor1\":30, \"floor2\":34, \"floor3\":36}");
+
+    // Check the address od datapoints
```
RVerhoeff, line 66: typo: Check the address of datapoints

RVerhoeff (changes requested):
Question about cleaning up a vector, plus minor typos.

RVerhoeff (approved):

MarkRiddoch (approved):

PR #963, opened 2023-02-08
FOGL-7398 : Fixed warning message not be shown for blank JSON Docs

gnandan:


C/services/north/north.cpp
```diff
@@ -394,10 +394,17 @@ void NorthService::start(string& coreAddress, unsigned short corePort)
...
-			northPlugin->startData(storedData);
+			if (!m_dryRun)
+			{
+				string key = m_name + m_pluginName;
+				string storedData = m_pluginData->loadStoredData(key);
+				logger->debug("Starting plugin with storedData: %s", storedData.c_str());
+				northPlugin->startData(storedData);
+			}
+			else
+			{
```
Innovative-ashwin, line 405: can put a debug log here

C/plugins/north/OMF/plugin.cpp
```diff
@@ -1486,7 +1486,12 @@ void loadSentDataTypes(CONNECTOR_INFO* connInfo,
 	}
 	else
 	{
-		Logger::getLogger()->warn("Persisted data is not of the correct format, ignoring");
+		// There is no stored data when plugin starts first time 
+		if (JSONData.MemberBegin() != JSONData.MemberEnd())
+		{
+			Logger::getLogger()->warn("Persisted data is not of the correct format, ignoring");
```
Innovative-ashwin, line 1492: what is being ignored, there is no executable statement after this warning

Innovative-ashwin (approved):

MarkRiddoch (dismissed):
Looks good

praveen-garg (commented):
In the scope of given problem, 

*In dry run mode**, We should not unnecessarily load stored data  if (northPlugin->persistData()). We should pass null (not empty) and if storedData is null then just skip the JSON parsing  if isNull & not loadSentDataTypes in plugin_start.

This is not just a problem of warning log in case of data doesn't exist or malformed JSON.

Also check/test for other plugins which can persist data; south plugins dry run load?

gnandan:
> 

Taken care. Now we are not calling start and shutdown method for plugin in case of dryrun is true

praveen-garg (approved):

MarkRiddoch (approved):

PR #977, opened 2023-02-17
Update 91_version_history file for 2.1.1 release

Mohit04tomar:


MarkRiddoch (commented):
I think we have missed a number of items as we switch to using the 2.1.1next fix version when the previous 2.1.1RC was generated.

PR #996, opened 2023-03-02
added wait time in system e2e notification tests

ashish-jabble:


praveen-garg (commented):
What has caused this?

ashish-jabble:
> What has caused this?

I don't think any recent changes in Product affected this, as this test was mostly fails on ubuntu20 with different runs. But now it fails on other platforms well. Also we had already put some extra wait on enabling schedule, So added wait time on disabling schedule as well to remove the flakiness nature of this test.

praveen-garg:
Notification service has got some new stuff to do for Subscription management; Hence it may have got 2-3 seconds of 
extra delay

praveen-garg (approved):

cyberwalk3r (approved):

Mohit04tomar (approved):

PR #998, opened 2023-03-02
FOGL-7431 Python logging level

ashish-jabble:
Logging level across the Fledge core from one place (i.e from Advanced->Logging category --- By default warning). Though for the initial start up logs from python core with INFO level & other areas like service,interest registry, monitoring, microservice calls are also with INFO level.


python/fledge/services/core/server.py
```diff
@@ -278,6 +278,26 @@ class Server:
             'order': '10'
         },
     }
+    _LOGGING_DEFAULT_CONFIG = {
+            'logLevel': {
+                         'description': 'Minimum logging level reported for Core server',
+                         'type': 'enumeration',
+                         'displayName': 'Minimum Log Level',
+                         'options': ['debug', 'info', 'warning', 'error', 'critical'],
```
praveen-garg, line 286: better to keep 
['Debug', 'Info', 'Notice', 'Warning', 'Error', 'Critical']

Titlecase and two more levels. if not all seven.

python/fledge/services/core/server.py
```diff
@@ -278,6 +278,26 @@ class Server:
...
+    _log_level = logging.WARNING
+    """ Numeric logging level for the message """
+
+    # TODO: Python core server should have INFO log level; we need to filtered out the logs for this file.
+    # from fledge.common import logger
+
+    _logger = Logger().get_logger(__name__)
+    # FIXME: If LOGGING category exists then set level from the DB value
+    logging.root.setLevel(_log_level)
+    """ Create Logger singleton class instance and set the root level to WARNING """
```
praveen-garg, line 300: move to a def and use after storage start.

praveen-garg (commented):
initial feedback, before testing

python/fledge/services/core/server.py
```diff
@@ -554,37 +600,36 @@ async def _start_service_monitor(cls):
...
     async def _start_storage(cls, loop):
         if loop is None:
             loop = asyncio.get_event_loop()
+        try:
             # callback with args
-        loop.call_soon(cls.__start_storage, cls._host, cls.core_management_port)
+            cls._logger.info("Start storage, from directory {}".format(_SCRIPTS_DIR))
+            loop.call_soon(cls.__start_storage, cls._host, cls.core_management_port)
+        except Exception as ex:
+            cls._logger.exception(str(ex))
```
praveen-garg, line 632: revert this change please.

python/fledge/services/core/server.py
```diff
@@ -807,6 +852,9 @@ def _start_core(cls, loop=None):
             cls._configuration_manager = ConfigurationManager(cls._storage_client_async)
             cls._interest_registry = InterestRegistry(cls._configuration_manager)
 
+            # Logging category
+            loop.run_until_complete(cls.log_config())
```
praveen-garg, line 856: log_config -> set_core_logger_config

python/fledge/services/core/server.py
```diff
@@ -504,15 +524,41 @@ async def installation_config(cls):
...
+    @classmethod
+    async def log_config(cls):
+        """ Get the logging level configuration """
+        try:
+            config = cls._LOGGING_DEFAULT_CONFIG
+            category = 'LOGGING'
+            description = "Logging Level of Core Server"
+            if cls._configuration_manager is None:
+                cls._logger.error("No configuration manager available.")
+            await cls._configuration_manager.create_category(category, config, description, True,
```
praveen-garg, line 547: if None then create object of that instead of logging error. then we are using that in next statement.

python/fledge/services/core/server.py
```diff
@@ -504,15 +524,41 @@ async def installation_config(cls):
...
+            config = await cls._configuration_manager.get_category_all_items(category)
+            log_level = config['logLevel']['value']
+            if log_level == 'debug':
+                logging_level = logging.DEBUG
+            elif log_level == 'info':
+                logging_level = logging.INFO
+            elif log_level == 'warning':
+                logging_level = logging.WARNING
+            else:
+                logging_level = logging.ERROR
```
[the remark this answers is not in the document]
praveen-garg: To avoid this if elif block repetition move to a common def here or in config manager; and use that.

python/fledge/services/core/server.py
```diff
@@ -504,15 +524,41 @@ async def installation_config(cls):
...
+            cls._log_level = logging_level
+        except Exception as ex:
+            cls._logger.exception(str(ex))
```
praveen-garg, line 561: in server py module, play with old logger module instead of the new one. This exception could be self destroying for that Logger class instance too.

praveen-garg (commented):
Basic testing with scheduler and Config manager looks okay!

MarkRiddoch (commented):
The approach looks pretty good. It's good that it is dynamic. Clearly unit tests are failing as I guess a number of them are expecting log entries are being written and now they are not.

MarkRiddoch (approved):
Thanks for this, looks good to me. It should help considerably with the logging.

MarkRiddoch (approved):

praveen-garg (approved):

PR #1010, opened 2023-03-10
logs for checking change callback payload and parsing at C layer

praveen-garg:


C/common/config_category.cpp
```diff
@@ -138,9 +138,11 @@ ConfigCategory::ConfigCategory(const string& name, const string& json) : m_name(
 {
 	Document doc;
 	doc.Parse(json.c_str());
+	// REMOVE ME
+	Logger::getLogger()->error("ConfigCategory::ConfigCategory %s", json.c_str());
```
praveen-garg, line 142:
```
"DefaultAFLocation" : {
 
  "type" : "string", 
  "value" : "/fledge\s/data_piwebapi/default",
  "default" : "/fledge/data_piwebapi/default", 
```

python/fledge/services/core/interest_registry/change_callback.py
```diff
@@ -43,7 +43,7 @@ async def run(category_name):
     category_value = await cfg_mgr.get_category_all_items(category_name)
     payload = {"category" : category_name, "items" : category_value}
     headers = {'content-type': 'application/json'}
-
+    _LOGGER.error("CHANGE CALLBACK PAYLOAD TO BE SENT %s", json.dumps(payload, sort_keys=True))
```
praveen-garg, line 46:
```

Fledge[210332] ERROR: change_callback: fledge.services.core.interest_registry.change_callback: CHANGE CALLBACK PAYLOAD TO BE SENT {"category": "V 1", "items": {"ADHRegions": {"default": "US-West", "description": "AVEVA Data Hub or OSIsoft Cloud Services region", "displayName": "Cloud Service Region", "group": "Cloud", "options": ["US-West", "EU-West", "Australia"], "order": "2", "type": "enumeration", "validity": "PIServerEndpoint == \"AVEVA Data Hub\" || PIServerEndpoint == \"OSIsoft Cloud Services\"", "value": "US-West"}, "AFMap": {"default": "{ }", "description": "Defines a set of rules to address where assets should be placed in the AF hierarchy.", "displayName": "Asset Framework hierarchy rules", "group": "Asset Framework", "order": "17", "type": "JSON", "validity": "PIServerEndpoint == \"PI Web API\"", "value": "{ }"}, "DefaultAFLocation": {"default": "/fledge/data_piwebapi/default", "description": "Defines the default location in the Asset Framework hierarchy in which the assets will be created, each level is separated by /, PI Web API only.", "displayName": "Default Asset Framework Location", "group": "Asset Framework", "order": "16", 

"type": "string", "validity": "PIServerEndpoint == \"PI Web API\"", 
"value": "/fledge\\s/data_piwebapi/default"},

```

praveen-garg: @pintomax the change callback is good.

C/common/config_category.cpp
praveen-garg: check C/services/common/management_api.cpp which received the change callback payload.

pintomax: "value" : "/fledge\s/data_piwebapi/default"

this is wrong \s is not valid, it should be \\s

You can easily verify it
# echo '{"value" : "/fledge\s/data_piwebapi/default"}' | jq
parse error: Invalid escape at line 1, column 44

The string passed to configchange  must contain \\s


You can easily verify it

echo '{"value" : "/fledge\\s/data_piwebapi/default"}' | jq
{
  "value": "/fledge\\s/data_piwebapi/default"
}

praveen-garg: @pintomax Please see other details.

python/fledge/services/core/interest_registry/change_callback.py
pintomax: In the above JSON I cannot see the value. log is truncated.

Please use PIWebAPIUserId which you should see complete.

Set Debug level of the north service.

Then in debug message you'll see the passed string has \s instead of \\s

praveen-garg: I have already added the value from logs. You might have not seen the comment carefully.

pintomax:
@praveen-garg The issue is in

 C/services/common/management_api.cpp::ManagementApi::configChange
 
 handler->configChange(conf.getName(), conf.itemsToJSON(true));
 
 This call conf.itemsToJSON(true) is removing the \\s leaving the wrong \s

MarkRiddoch (changes requested):
Please do not merge this to develop

praveen-garg:
@MarkRiddoch That is obvious! This is to discuss and see the issue, for clarity and understanding.

pintomax:
@praveen-garg you can check this against branch FOGL-7503

PR #1011, opened 2023-03-13
FOGL-7503: JSON string fix for special chars

pintomax:
FOGL-7503: JSON string fix for special chars

### DIFF
--- a/C/common/config_category.cpp
+++ b/C/common/config_category.cpp
@@ -1152,14 +1152,15 @@ ConfigCategory::CategoryItem::CategoryItem(const string& name,
 	// Item "value" is just a string
 	else if (item.HasMember("value") && item["value"].IsString())
 	{
+		// Get content of script type item as is
+		rapidjson::StringBuffer strbuf;
+		rapidjson::Writer<rapidjson::StringBuffer> writer(strbuf);
+		item["value"].Accept(writer);
+		m_value = strbuf.GetString();
+
 		if (m_itemType == ScriptItem ||
 		    m_itemType == CodeItem)
 		{
-			// Get content of script type item as is
-			rapidjson::StringBuffer strbuf;
-			rapidjson::Writer<rapidjson::StringBuffer> writer(strbuf);
-			item["value"].Accept(writer);
-			m_value = strbuf.GetString();
 			if (m_value.empty())
 			{
 				m_value = "\"\"";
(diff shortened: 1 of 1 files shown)


MarkRiddoch (approved):

praveen-garg:
@pintomax Looks like it is not tested with basic service addition/original issue? 

Add any service, and you see a crash!

For `sin` with sinusoid:

```
Fledge sin[28896]: ERROR: Failed to fetch ACL: ACL with name "" is not found..
Fledge sin[28896]: FATAL: Signal 6 (Aborted) trapped:
Fledge sin[28896]: FATAL: (0) 0       0x55a5b7f9dc59 handler(int) + 73---------
Fledge sin[28896]: FATAL: (1) 1       0x7f80b08b2f10 /lib/x86_64-linux-gnu/libc.so.6(+0x3ef10) [0x7f80b08b2f10]---------
```

C/common/config_category.cpp
```diff
@@ -1152,27 +1152,30 @@ ConfigCategory::CategoryItem::CategoryItem(const string& name,
...
+
 			if (m_options.size() == 0)
 				m_itemType = StringItem;
 			else
 				m_itemType = EnumerationItem;
 		}
+
+		if (m_value.empty())
+		{
+			m_value = "\"\"";
```
praveen-garg, line 1177: Can we add a comment here, what is this for?

C/common/config_category.cpp
```diff
@@ -1255,26 +1258,27 @@ ConfigCategory::CategoryItem::CategoryItem(const string& name,
...
-			m_default = item["default"].GetString();
+			m_default = JSONunescape(strbuf.GetString());
 			if (m_options.size() == 0)
 				m_itemType = StringItem;
 			else
 				m_itemType = EnumerationItem;
 		}
+		if (m_default.empty())
+		{
+			m_default = "\"\"";
```
praveen-garg, line 1280: As above

MarkRiddoch (approved):

praveen-garg:
@pintomax seeing issues with PostgreSQL, for OMF scenario. 

It says, Escape sequence is invalid.

MarkRiddoch:
The pull request addresses the issue that was in the original Jira, which is about the quotes in validity rules. This is a problem that impacts everybody and has no workaround. The other things I am seeing being raised here are the use of the backslash character, or others in asset names, service names etc. This is a different problem and has a workaround, i.e. not using these characters. Please can we put this in a separate Jira as it has a different priority to the real problem addressed by the original Jira.

MarkRiddoch (approved):

praveen-garg (approved):

PR #1017, opened 2023-03-20
FOGL-7567: Modified to remove delay in asset creation

gnandan:
A separate non-blocking thread is used to schedule to check for empty table to remove delay in asset creation.
Purge schedule and purge option from developer tab also uses same class method which is being in thread. So working fine with manual purge from developer tab and purge schedule.

### DIFF
--- a/C/plugins/storage/sqlite/common/include/readings_catalogue.h
+++ b/C/plugins/storage/sqlite/common/include/readings_catalogue.h
@@ -131,11 +131,10 @@ public:
 
 	void          preallocateReadingsTables(int dbId);
 	bool          loadAssetReadingCatalogue();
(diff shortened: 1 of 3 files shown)


MarkRiddoch (commented):
There is no need to run this every two minutes. You are adding load to the storage service that'll does not need to be added. The only way a table can become free is via the purge operation. Therefore there is no need to run anything until a purge is done. In fact the most efficient way would be to look at purge and see if it already knows the number of rows left in a table after a purge operation.

Mohit04tomar (approved):

Innovative-ashwin (approved):

MarkRiddoch (approved):

PR #1018, opened 2023-03-21
FOGL-7579 package name handling in Plugins PUT & DELETE API

ashish-jabble:
~Note: I haven't changed the param in request - name Vs PackageName. So we need to decide for the backward compatability by checking fledge core version `check > 2.1.0`; as this endpoint is only use by FogLAMP Manage.~

~As this endpoint is to update the package. So we should request with package name to make more clarity in the endpoint.~

New endpoints available with 2.1.0 onwards core version. More, details added on JIRA
```
PUT /fledge/plugins/{package_name}
DELETE /fledge/plugins/{package_name}
```

### DIFF
--- a/python/fledge/services/core/api/plugins/remove.py
+++ b/python/fledge/services/core/api/plugins/remove.py
@@ -25,8 +25,8 @@ from fledge.services.core.api.plugins import common
 from fledge.services.core.api.plugins.exceptions import *
 
 
-__author__ = "Rajesh Kumar"
-__copyright__ = "Copyright (c) 2020, Dianomic Systems Inc."
+__author__ = "Rajesh Kumar, Ashish Jabble"
+__copyright__ = "Copyright (c) 2020-2023, Dianomic Systems Inc."
 __license__ = "Apache 2.0"
 __version__ = "${VERSION}"
 
@@ -62,6 +62,9 @@ async def remove_plugin(request: web.Request) -> web.Response:
         plugin_type = str(plugin_type).lower()
         if plugin_type not in valid_plugin_types:
             raise ValueError("Invalid plugin type. Please provide valid type: {}".format(valid_plugin_types))
+        # only OMF is an inbuilt plugin
+        if name.lower() == 'omf':
+            raise ValueError("Cannot delete an inbuilt {} plugin.".format(name.upper()))
         if plugin_type == 'notify':
             installed_dir_name = 'notificationDelivery'
         elif plugin_type == 'rule':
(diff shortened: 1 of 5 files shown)


pintomax:
@ashish-jabble If we keep existing API endpoint for PUT and delete, then FogLAMP manage will work without any instance version check.

Otherwise, a change for PUT and DELETE requires deployment code change in order to call old and new endpoint based on instance version > 2.1.0

This requires some work and tests: if it's worth it then we can do it.

pintomax:
@ashish-jabble in case of API endpoint change I think it's better to use

PUT          /fledge/plugins/fledge-south-modbus/update             (Without plugin type)
DELETE    /fledge/plugins/fledge-south-modbus                          (Without plugin type)

MarkRiddoch:
It makes sense to use the same convention as with the POST call to install a package.

ashish-jabble:
@MarkRiddoch 
> It makes sense to use the same convention as with the POST call to install a package.

Shall I go ahead to change the API URL for PUT & DELETE? ?

```
PUT /fledge/plugins/fledge-south-modbus
DELETE /fledge/plugins/fledge-south-modbus
```
Also, please confirm do we really need to handle backward compatibility of the API URL at Admin API side with the `fledge core version > 2.1.0`?

pintomax:
> @MarkRiddoch
> 
> > It makes sense to use the same convention as with the POST call to install a package.
> 
> Shall I go ahead to change the API URL for PUT & DELETE? ?
> 
> ```
> PUT /fledge/plugins/fledge-south-modbus
> DELETE /fledge/plugins/fledge-south-modbus
> ```
> 
> Also, please confirm do we really need to handle backward compatibility of the API URL at Admin API side with the `fledge core version > 2.1.0`?

The PUT API shouldn't it be PUT /fledge/plugins/fledge-south-modbus/update ?

MarkRiddoch:
> @MarkRiddoch
> 
> > It makes sense to use the same convention as with the POST call to install a package.
> 
> Shall I go ahead to change the API URL for PUT & DELETE? ?
> 
> ```
> PUT /fledge/plugins/fledge-south-modbus
> DELETE /fledge/plugins/fledge-south-modbus
> ```
> 
> Also, please confirm do we really need to handle backward compatibility of the API URL at Admin API side with the `fledge core version > 2.1.0`?

Yes please

pintomax (dismissed):

ashish-jabble:
Hi @MarkRiddoch Shall we merge this into develop branch?

MarkRiddoch:
> Hi @MarkRiddoch Shall we merge this into develop branch?

Yes, I think so

PR #1025, opened 2023-03-24
FOGL-7526: Added EDS version check

RVerhoeff:
OSIsoft has had 3 releases of Edge Data Store so far: EDS 2020 (1.0.0.609) which supported OMF 1.0 and 1.1, and EDS 2023 (1.1.1.46) and EDS 2023 Patch 1 (1.1.3.2) which supported OMF 1.2 in addition. A check for EDS version has been added so OMF North can determine the proper OMF JSON documents to send. This commit has only been tested with EDS 2020 so far.

### DIFF
--- a/C/plugins/north/OMF/plugin.cpp
+++ b/C/plugins/north/OMF/plugin.cpp
@@ -431,7 +431,7 @@ typedef struct
 
 	string		prefixAFAsset;          // Prefix to generate unique asste id
 	string		PIWebAPIProductTitle;
-	string		PIWebAPIVersion;
+	string		RestServerVersion;
 	string		PIWebAPIAuthMethod;     // Authentication method to be used with the PI Web API.
 	string		PIWebAPICredentials;    // Credentials is the base64 encoding of id and password joined by a single colon (:)
 	string 		KerberosKeytab;         // Kerberos authentication keytab file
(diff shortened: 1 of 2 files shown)


C/plugins/north/OMF/plugin.cpp
```diff
@@ -1553,42 +1548,145 @@ int PIWebAPIGetVersion(CONNECTOR_INFO* connInfo, bool logMessage)
...
 	_PIWebAPI->setAuthBasicCredentials(connInfo->PIWebAPICredentials);
 
-	int httpCode = _PIWebAPI->GetVersion(connInfo->hostAndPort, connInfo->PIWebAPIVersion, logMessage);
+	int httpCode = _PIWebAPI->GetVersion(connInfo->hostAndPort, connInfo->RestServerVersion, logMessage);
 	delete _PIWebAPI;
 
 	return httpCode;
 }
 
+static std::string ParseEDSProductInformation(std::string json)
```
MarkRiddoch, line 1557: We should have doxygen comment blocks on new functions.

C/plugins/north/OMF/plugin.cpp
```diff
@@ -1553,42 +1548,145 @@ int PIWebAPIGetVersion(CONNECTOR_INFO* connInfo, bool logMessage)
...
+		catch (...)
+		{
+		}
+	}
+
+	Logger::getLogger()->debug("Edge Data Store Version: %s JSON: %s", version.c_str(), json.c_str());
+	return version;
+}
+
+int EDSGetVersion(CONNECTOR_INFO *connInfo)
```
MarkRiddoch, line 1582: We should have doxygen comment blocks on new functions.

C/plugins/north/OMF/plugin.cpp
RVerhoeff: Done. I also created a new routine called _ParseProductVersion_ so that the version parsing is not done directly in _SetOMFVersion_.

C/plugins/north/OMF/plugin.cpp
RVerhoeff: Done. Also added a new routine _ParseProductVersion_ so that the major/minor version parsing are not done directly in _SetOMFVersion_.

MarkRiddoch (approved):

PR #1031, opened 2023-03-30
FOGL-7556 Add documentation for new GUI features

MarkRiddoch:


ashish-jabble (approved):
Doc style is good.
Also in the new multigraph images there is black border outlined which is not consistent across and may not be required.

monika-Dianomic (approved):
lgtm!

PR #1035, opened 2023-04-06
FOGL-7635: Added support for ingesting readings generated by filters in plugin_shutdown

AmandeepArora:


C/services/south/ingest.cpp
```diff
@@ -676,7 +682,7 @@ void Ingest::processQueue()
 		 */
 		{
 			lock_guard<mutex> guard(m_pipelineMutex);
-			if (m_filterPipeline)
+			if (m_filterPipeline && !m_filterPipeline->isShuttingDown())
```
Innovative-ashwin, line 685: && is left to right associative but which operand will be executed first is compiler dependant, second operand might be evaluated first, in which case , it will be wrong.

Innovative-ashwin (approved):

C/services/south/ingest.cpp
AmandeepArora: I don't think so. Can you please provide some reference?

MarkRiddoch (approved):

PR #1037, opened 2023-04-11
FOGL-7643 Errors in reset of test_e2e_fledge_pair.py

Mohit04tomar:


tests/system/python/pair/test_e2e_fledge_pair.py
```diff
@@ -81,7 +81,13 @@ def reset_and_start_fledge_remote(self, storage_plugin, remote_user, remote_ip,
                                                                                                       remote_ip,
                                                                                                       remote_fledge_path)], shell=True, check=True)
         storage_plugin_val = "postgres" if storage_plugin == 'postgres' else "sqlite"
-        subprocess.run(["ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i {} {}@{} echo $(jq -c --arg STORAGE_PLUGIN_VAL {} '.plugin.value=$STORAGE_PLUGIN_VAL' {}/data/etc/storage.json) > {}/data/etc/storage.json".format(key_path, remote_user, remote_ip, storage_plugin_val, remote_fledge_path, remote_fledge_path)], shell=True, check=True)
+        # Check whether storage.json file exist on remote machine or not, if it doesn't exist then raise assertion otherwise update its storage plugin value.
```
ashish-jabble, line 84: We should avoid the storage.json existence check as reset fixture should be complete reset.
Before running the test machine should be in clean state.

YashTatkondawar (approved):

tests/system/python/pair/test_e2e_fledge_pair.py
praveen-garg: @ashish-jabble I think at this stage of test, instance must have a storage JSON file.

praveen-garg (approved):

tests/system/python/pair/test_e2e_fledge_pair.py
ashish-jabble: Correct. As storage.json will be created on fledge start and even on fledge kill/stop commands.

PR #1039, opened 2023-04-13
FOGL-7432 - multiline traceback utility method added in python logger

ashish-jabble:
Below is one of the example; On audit entry failure

Python traceback with multilines
```
Apr 13 13:35:51 aj-ThinkPad-E450 Fledge[16823] ERROR: audit: fledge.services.core.api.audit: Failed to fetch log audit entry.
Apr 13 13:35:51 aj-ThinkPad-E450 Fledge[16823] ERROR: audit: fledge.services.core.api.audit: Traceback (most recent call last):
Apr 13 13:35:51 aj-ThinkPad-E450 Fledge[16823] ERROR: audit: fledge.services.core.api.audit:   File "/home/aj/fledge/python/fledge/services/core/api/audit.py", line 252, in get_audit_entries#012    rows = results['rows1']
Apr 13 13:35:51 aj-ThinkPad-E450 Fledge[16823] ERROR: audit: fledge.services.core.api.audit: KeyError: 'rows1'
```

### DIFF
--- a/python/fledge/common/logger.py
+++ b/python/fledge/common/logger.py
@@ -7,6 +7,7 @@
 """ Fledge Logger """
 import os
 import subprocess
+import traceback
 import logging
 from logging.handlers import SysLogHandler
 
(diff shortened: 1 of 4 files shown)


MarkRiddoch (commented):
Looks good

praveen-garg:
I would say - avoid this approach to have for loop in all exception logs individually. We should find a way to do it using a wrapper function or context manager in logging class itself

ashish-jabble:
> I would say - avoid this approach to have for loop in all exception logs individually. We should find a way to do it using a wrapper function or context manager in logging class itself

In pythonic way there is actually no way to achieve the traceback as same as C language. I have tried so many ways in past but couldn't get more closer results. Though python exception messages are still understandable without this change given fact that logs come in a single line.

Anyway, I tried with Custom handler we may try more or put some efforts into it. I have pushed the [branch](https://github.com/fledge-iot/fledge/compare/FOGL-7432.custom_handler?expand=1) and check.

ashish-jabble:
Closed in favor of https://github.com/fledge-iot/fledge/pull/1044

PR #1040, opened 2023-04-13
FOGL-7616: Passed mgmt_client_url_base in config

gnandan:
FOGL-7667 is created to have access to management client into north plugins


MarkRiddoch (changes requested):
Why are you doing this? What is thew requirement?

gnandan:
> 

Hi Mark, 
It's required to fetch Foglamp instance name, which needs to be populated into the table at Redshift

MarkRiddoch:
I don't think we should do this here. Having seen your other review with the redshift plugin ode let's simply pass a name fr now and have a better thought out approach to this.

PR #1044, opened 2023-04-17
FOGL-7432 override logging error in python logger class

ashish-jabble:
Note - We need to revisit logger error/exception messages to get traceback in multiline via another JIRA. First, we will track down Python core and then Admin API.

### DIFF
--- a/python/fledge/common/logger.py
+++ b/python/fledge/common/logger.py
@@ -8,7 +8,11 @@
 import os
 import subprocess
 import logging
+import inspect
+import traceback
 from logging.handlers import SysLogHandler
+from functools import wraps
+
 
 __author__ = "Praveen Garg, Ashish Jabble"
 __copyright__ = "Copyright (c) 2017-2023 OSIsoft, LLC"
@@ -23,10 +27,10 @@ r"""Send log entries to /var/log/syslog
 """
 CONSOLE = 1
 """Send log entries to STDERR"""
-
-
-FLEDGE_LOGS_DESTINATION = 'FLEDGE_LOGS_DESTINATION'  # env variable
-default_destination = SYSLOG    # default for fledge
+FLEDGE_LOGS_DESTINATION = 'FLEDGE_LOGS_DESTINATION'
+"""Log destination environment variable"""
+default_destination = SYSLOG
+"""Default destination of logger"""
 
 
 def set_default_destination(destination: int):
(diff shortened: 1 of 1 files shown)


praveen-garg:
Remove those loop based statements in exception handler blocks? To test and clean up the base branch code

MarkRiddoch (dismissed):
Looks good to me

python/fledge/common/logger.py
```diff
@@ -8,7 +8,11 @@
 import os
 import subprocess
 import logging
+import inspect
```
praveen-garg, line 11: remove?

python/fledge/common/logger.py
```diff
@@ -181,10 +202,37 @@ def get_logger(self, logger_name: str):
...
         self.add_handlers(_logger, [syslog_handler, console_handler])
         _logger.propagate = False
+
+        @wraps(_logger.error)
+        def error(msg, *args, **kwargs):
+            """Case: When we pass exception in error and having different args
+            For example:
+            a) _logger.error(ex)
+            b) _logger.error(ex, "Failed to add data.")
+            """
```
praveen-garg, line 220: """ Override error logger to print multi-line traceback and error string with newline"

case/example part can be moved to if block, as given in the else part.

python/fledge/common/logger.py
```diff
@@ -181,10 +202,37 @@ def get_logger(self, logger_name: str):
...
+            if isinstance(msg, Exception):
+                trace_msg = traceback.format_exception(msg.__class__, msg, msg.__traceback__)
+                if args:
+                    trace_msg[:0] = ["{}\n".format(args[0])]
+                [__logging_error(line.strip('\n')) for line in trace_msg]
+            else:
+                """Case: When we pass string in error
+                For example:  
+                a) _logger.error(str(ex))
+                b) _logger.error("Failed to add data for key: {} along with error:{}".format("Test", str(ex)))
```
praveen-garg, line 230: a) _logger.error("Failed to log audit trail entry")
           b) _logger.error("Failed to log audit trail entry '{}' \n{}".format(code, str(ex))) # \n will result the str(ex) to be printed in next line

praveen-garg (dismissed):
Tested okay with injecting a simulated exception.

Let's  do a manual review of syslogs with some make based system test run & also backup/restore/north python special logs

PR #1045, opened 2023-04-17
FOGL-7057 Fix resources leaks and add overflow table

MarkRiddoch:


pintomax (approved):

C/services/storage/storage_api.cpp
```diff
@@ -396,6 +396,19 @@ StorageApi::StorageApi(const unsigned short port, const unsigned int threads) :
...
+/**
+ * Destructor for the storage API class. There is only ever one StorageApi class
+ * in existance and it lives for the entire duration of the storage service, so this
+ * is really for completerness rather than any pracitical use.
+ */
+StorageApi::~StorageApi()
+{
+	if (m_server)
+	{
+		delete m_server;
```
Innovative-ashwin, line 408: it is a good practice to set ptr to nullptr after deletion

PR #1048, opened 2023-04-20
FOGL-7612 : Added SetValue function for string type

gnandan:


C/common/include/datapoint.h
```diff
@@ -140,7 +140,21 @@ class DatapointValue {
...
 
-		
+		/**
+                 * Set the value of a datapoint, this may
+                 * also cause the type to be changed.
+                 * @param value An string value to set
+                 */
+                void setValue(std::string value)
+                {
+                        if(m_value.str)
```
AmandeepArora, line 150: In C++, delete is safe on null pointers, so this check is not required.

PR #1050, opened 2023-04-21
FOGL-7644: Test statistics history notification rule system tests

YashTatkondawar:


tests/system/python/packages/test_statistics_history_notification_rule.py
```diff
@@ -0,0 +1,293 @@
...
+        "rule_config": {
+            "source": "Statistics History",
+            "asset": "READINGS",
+            "trigger_value": "10.0",
+        },
+        "channel": "asset",
+        "delivery_config": {"enable": "true"},
+        "notification_type": "retriggered",
+        "retrigger_time": "30",
+        "enabled": True,
```
praveen-garg, line 102: Here we are using bool? I think in fledge we handle every type as string instead of their actual type.

tests/system/python/packages/test_statistics_history_notification_rule.py
```diff
@@ -0,0 +1,293 @@
...
+        
+        # When rule is triggered, there should be audit entries for NTFSN
+        get_url = "/fledge/audit?source=NTFSN"
+        resp1 = utils.get_request(fledge_url, get_url)
+        assert len(resp1['audit'])
+        assert "test #1" in [s["details"]["name"] for s in resp1["audit"]]
+        for audit_detail in resp1['audit']:
+            if "test #1" == audit_detail['details']['name']:
+                assert "NTFSN" == audit_detail['source']
+        # Waiting for 60 sec to get more NTFSN entries
```
praveen-garg, line 183: more? Test should tell us exact expected

tests/system/python/packages/test_statistics_history_notification_rule.py
```diff
@@ -0,0 +1,295 @@
...
+import os
+import ssl
+import subprocess
+import time
+import urllib.parse
+from pathlib import Path
+
+import pytest
+import utils
+from pytest import PKG_MGR
```
Mohit04tomar, line 25: This import is not required if we are calling pytest.PKG_MGR in line L#76

tests/system/python/packages/test_statistics_history_notification_rule.py
```diff
@@ -0,0 +1,295 @@
...
+        resp1 = utils.get_request(fledge_url, get_url)
+        assert len(resp1['audit'])
+        assert "test #1" in [s["details"]["name"] for s in resp1["audit"]]
+        for audit_detail in resp1['audit']:
+            if "test #1" == audit_detail['details']['name']:
+                assert "NTFSN" == audit_detail['source']
+        # Waiting for 60 sec to get more NTFSN entries
+        time.sleep(60)
+        resp2 = utils.get_request(fledge_url, get_url)
+        assert len(resp2['audit']) - len(resp1['audit']) == 2, "ERROR: NTFSN not triggered properly"
```
Mohit04tomar, line 250: I have tried executing this test multiple time in my local machine but this test failed in every runof this pytest at this step.
Please share the link of any Jenkins job that is successful.

Mohit04tomar:
Please add a way in Jenkins job to skip execution of tests of class (TestStatisticsHistoryBasedNotificationRuleOnIngress) and only execute tests of class (TestStatisticsHistoryBasedNotificationRuleOnEgress) only if RUN_NORTH_EGRESS_TEST is True.

As of now, when RUN_NORTH_EGRESS_TEST is True then tests of class (TestStatisticsHistoryBasedNotificationRuleOnIngress) execute and failed at verify_ping because they try to check data sent to north or not where north is not configured.
Only test of class (TestStatisticsHistoryBasedNotificationRuleOnEgress) executes because OMF is configured in that test only.

Mohit04tomar (approved):

PR #1051, opened 2023-04-24
FOGL-7685 args fixes in logger and other fixes

ashish-jabble:
This primarily has the fix of below cases:

a) string format when having with %s as it prints as is
```
For example:
_logger.error('Failed to log audit trail entry for code: %s', "CONCH")
O/P
Failed to log audit trail entry for code: %s
```
b) non-Exception cases 
For example:
```
color_list = ['Orange', 'White', 'Green']
i) _logger.error(color_list)
color_dict = {'colors': {'Orange', 'White', 'Green'}}
ii) _logger.error(color_dict)
iii) _logger.error(type(request))

O/p
 i) 'list' object has no attribute 'splitlines'
ii) 'dict' object has no attribute 'splitlines'
iii) type object 'Request' has no attribute 'splitlines'
```

This needs to get merged ASAP to block others

### DIFF
--- a/python/fledge/common/logger.py
+++ b/python/fledge/common/logger.py
@@ -108,9 +108,6 @@ def setup(logger_name: str = None,
     else:
         raise ValueError("Invalid destination {}".format(destination))
 
-    # save the old logging.error function
-    __logging_error = logger.error
-
     # TODO: Consider using %r with message when using syslog .. \n looks better than #
     fmt = '{}[%(process)d] %(levelname)s: %(module)s: %(name)s: %(message)s'.format(get_process_name())
     formatter = logging.Formatter(fmt=fmt)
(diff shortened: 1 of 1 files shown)


python/fledge/common/logger.py
```diff
@@ -120,20 +117,51 @@ def setup(logger_name: str = None,
...
+                """For example:
+                    a) _logger.error(str(ex))
+                    b) _logger.error("Failed to log audit trail entry")
+                    c) _logger.error('Failed to log audit trail entry for code: %s', "CONCH")
+                    d) _logger.error('Failed to log audit trail entry for code: {log_code}'.format(log_code="CONAD"))
+                    e) _logger.error('Failed to log audit trail entry for code: {0}'.format("CONAD"))
+                    f) _logger.error("Failed to log audit trail entry for code '{}' \n{}".format("CONCH", "Next line"))
+                """
+                if args:
+                    msg = msg % args
```
praveen-garg, line 158: Add inline comment here in next PRs for log statements fixes

praveen-garg (approved):
lgtm!


This is great, that we avoided def repetition here.

python/fledge/common/logger.py
ashish-jabble: Sure

PR #1053, opened 2023-04-25
FOGL-5409 error handling added in create user

ashish-jabble:
Return early when user sent bad request.

This PR only to handle the storage exception that was occurred and DB errors in syslog on ADD user when attempt with deactivated user. 

Also new audit log codes step and handling to override the values in separate PR as new enhancement.


MarkRiddoch (commented):
Looks good to me

MarkRiddoch (approved):

PR #1057, opened 2023-04-26
FOGL-7645: Add system test for rule data availability

YashTatkondawar:


tests/system/python/packages/test_rule_data_availability.py
```diff
@@ -0,0 +1,278 @@
...
+import json
+from pathlib import Path
+import http
+from datetime import datetime
+import pytest
+import utils
+from pytest import PKG_MGR
+
+# This  gives the path of directory where fledge is cloned. test_file < packages < python < system < tests < ROOT
+PROJECT_ROOT = Path(__file__).parent.parent.parent.parent.parent
```
Mohit04tomar, line 27: We can also use`PROJECT_ROOT = subprocess.getoutput("git rev-parse --show-toplevel")`

PR #1064, opened 2023-05-04
pipeline is not allowed on the same Service Or Script type for both source and destination

ashish-jabble:
As per matrix provided in FOGL-7735 comments.


monika-Dianomic (approved):

MarkRiddoch (approved):
Looks right to me

PR #1071, opened 2023-05-22
FOGL-7687 Add fixture to install & add external services in conftest.py

Mohit04tomar:


ashish-jabble (changes requested):
Nope this is not the right fix. Change the sed expression with match pattern to find and comment out the line.

Mohit04tomar:
Closed due to fix of https://github.com/dianomic/FogLAMP-release/pull/133

PR #1084, opened 2023-06-12
FOGL-7860 Fix Caching endpoint address scheme of core management API used by Package management API

praveen-garg:
Similar issue was fixed with https://github.com/fledge-iot/fledge/pull/994


pintomax (dismissed):

ashish-jabble (dismissed):
If the idea is to fix only for removal of plugin scenario then its okay. As we need to fix the same thing at other places probably with new JIRA. Like update plugin and service and I believe these areas will not used either by GUI client or FogLAMP Manage.

PR #1090, opened 2023-06-14
FOGL-7861: Memory growth on north service for DICT and LIST readings.

cyberwalk3r:
For north Python plugins, when converting T_DP_DICT/T_DP_LIST readings to PyObject, new DICT/LIST is being created for each datapoint.

### DIFF
--- a/C/common/pythonreading.cpp
+++ b/C/common/pythonreading.cpp
@@ -577,9 +577,9 @@ PyObject *PythonReading::convertDatapoint(Datapoint *dp, bool bytesString)
 	else if (dataType == DatapointValue::dataTagType::T_DP_DICT)
 	{
 		vector<Datapoint *>* children = dp->getData().getDpVec();;
+		value = PyDict_New();
 		for (auto child = children->begin(); child != children->end(); ++child)
 		{
-			value = PyDict_New();
 			PyObject *childValue = convertDatapoint(*child);
 			// Add Datapoint: key and value
 			PyObject *key = PyUnicode_FromString((*child)->getName().c_str());
@@ -593,9 +593,9 @@ PyObject *PythonReading::convertDatapoint(Datapoint *dp, bool bytesString)
 	{
 		vector<Datapoint *>* children = dp->getData().getDpVec();
 		int i = 0;
+		value = PyList_New(children->size());
 		for (auto child = children->begin(); child != children->end(); ++child)
 		{
-			value = PyList_New(children->size());
 			PyObject *childValue = convertDatapoint(*child);
 			// TODO complete
 			// Add Datapoint: key and value


AmandeepArora:
> For north Python plugins, when converting T_DP_DICT/T_DP_LIST readings to PyObject, new DICT/LIST is being created for each datapoint.

This would be true for south service also?

cyberwalk3r:
> > For north Python plugins, when converting T_DP_DICT/T_DP_LIST readings to PyObject, new DICT/LIST is being created for each datapoint.
> 
> This would be true for south service also?

No this would not be applicable to south service, in case of south service PyObject is converted to DatapointValue. (func getDatapointValue)

AmandeepArora:
> > > For north Python plugins, when converting T_DP_DICT/T_DP_LIST readings to PyObject, new DICT/LIST is being created for each datapoint.
> > 
> > 
> > This would be true for south service also?
> 
> No this would not be applicable to south service, in case of south service PyObject is converted to DatapointValue. (func getDatapointValue)

On south side, this would be required when data has to be passed thru' a python filter next.

AmandeepArora:
And with the code before the fix, it is not just a matter of memory leak, it would also have been incorrect behavior. i.e. convertDatapoint() would be returning only the last PyDict/PyList in case of multiple list/dict items.

AmandeepArora (approved):
Looks good.

praveen-garg:
> > > > For north Python plugins, when converting T_DP_DICT/T_DP_LIST readings to PyObject, new DICT/LIST is being created for each datapoint.
> 
> > > 
> 
> > > 
> 
> > > This would be true for south service also?
> 
> > 
> 
> > No this would not be applicable to south service, in case of south service PyObject is converted to DatapointValue. (func getDatapointValue)
> 
> 
> 
> On south side, this would be required when data has to be passed thru' a python filter next.

@AmandeepArora Should a Jira be logged for this?

AmandeepArora:
> > > > > For north Python plugins, when converting T_DP_DICT/T_DP_LIST readings to PyObject, new DICT/LIST is being created for each datapoint.
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > This would be true for south service also?
> > 
> > 
> > > 
> > 
> > 
> > > No this would not be applicable to south service, in case of south service PyObject is converted to DatapointValue. (func getDatapointValue)
> > 
> > 
> > On south side, this would be required when data has to be passed thru' a python filter next.
> 
> @AmandeepArora Should a Jira be logged for this?

It seems similar problem would have occurred in south service + python filter case also, just need to verify it. There probably won't be any further code changes required.

MarkRiddoch (approved):

PR #1091, opened 2023-06-14
FOGL-7862: Rapid memory growth in north service when plugin raises error

cyberwalk3r:


gnandan (approved):

C/services/north/data_send.cpp
```diff
@@ -84,6 +84,9 @@ void DataSender::sendThread()
 
 				// Set readings removal
 				removeReadings = vec->size() == 0;
+			} else {
+				// If no reading sent, set readings removal
```
AmandeepArora, line 88: If no readings are sent out in this attempt, shouldn't there be subsequent attempt(s) to do so?

cyberwalk3r: Yes, there can be subsequent attempts. Currently, readings are fetched again based on the last sent id. For optimization, We'll have to figure out a way to use the same reading set and keep the storage service in sync accordingly.

praveen-garg: We better log a Jira and add TODO: FOGL-xxx ... 

I think we have similar issue logged already, where it just jump over unsent block. 

And here inline comment should be updated why cleanup is required when gave up trying/ failed to send

MarkRiddoch (approved):

C/services/north/data_send.cpp
AmandeepArora: Yes, this should be fixed under a separate JIRA. Reading again and again from DB for retransmission doesn't sound good.

AmandeepArora (approved):

PR #1094, opened 2023-06-16
FOGL-7886: pospone write and operation calls for async plugins

pintomax:
FOGL-7886: pospone write and operation calls for async plugins


MarkRiddoch (approved):
Looks good to me

PR #1097, opened 2023-06-22
FOGL-7855:Reconfig and shutdown  does not work with north Kafka-C plugin

Innovative-ashwin:


MarkRiddoch (approved):

gnandan:
After the root cause analysis, the real problem identified was different which is fixed by https://github.com/fledge-iot/fledge-north-kafka/pull/34

PR #1100, opened 2023-06-26
FOGL-7881 service and plugin API cleanup fixes

ashish-jabble:


python/fledge/services/core/api/service.py
```diff
@@ -665,10 +663,10 @@ async def update_service(request: web.Request) -> web.Response:
         return web.json_response(result_payload)
 
 
-async def _put_schedule(protocol: str, host: str, port: int, sch_id: uuid, is_enabled: bool) -> None:
-    management_api_url = '{}://{}:{}/fledge/schedule/{}/enable'.format(protocol, host, port, sch_id)
+async def _put_schedule(host: str, port: int, sch_id: uuid, is_enabled: bool) -> None:
+    management_api_url = 'http://{}:{}/fledge/schedule/{}/enable'.format(host, port, sch_id)
     headers = {'content-type': 'application/json'}
-    verify_ssl = False if protocol == 'HTTP' else True
+    verify_ssl = False
```
praveen-garg, line 669: Yes

python/fledge/services/core/api/plugins/remove.py
```diff
@@ -360,8 +360,9 @@ async def _check_plugin_usage_in_notification_instances(plugin_name: str) -> lis
     return notification_instances
 
 
-async def _put_refresh_cache(protocol: str, host: int, port: int) -> None:
-    management_api_url = '{}://{}:{}/fledge/cache'.format(protocol, host, port)
+async def _put_refresh_cache(host: int, port: int) -> None:
```
praveen-garg, line 363: I would suggest to keep these def as is, and from caller pass http only. We may really want to run all this on TLS in near future.

python/fledge/services/core/api/plugins/remove.py
```diff
@@ -169,7 +169,7 @@ def _uninstall(pkg_name: str, version: str, uid: uuid, storage: connect) -> tupl
         loop.run_until_complete(storage.update_tbl("packages", payload))
         if code == 0:
             # Clear internal cache
-            loop.run_until_complete(_put_refresh_cache("http", Server._host, Server.core_management_port))
+            loop.run_until_complete(_put_refresh_cache(Server._host, Server.core_management_port))
```
praveen-garg, line 172: keep it as is.

python/fledge/services/core/api/plugins/remove.py
```diff
@@ -417,7 +418,7 @@ def purge_plugin(plugin_type: str, plugin_name: str, pkg_name: str, version: str
 
         if code == 0:
             # Clear internal cache
-            loop.run_until_complete(_put_refresh_cache("http", Server._host, Server.core_management_port))
+            loop.run_until_complete(_put_refresh_cache(Server._host, Server.core_management_port))
```
praveen-garg, line 421: Keep as is

python/fledge/services/core/api/plugins/update.py
```diff
@@ -148,12 +148,9 @@ async def update_package(request: web.Request) -> web.Response:
             if response:
                 pn = "{}-{}".format(action, package_name)
                 uid = response[0]['id']
-                p = multiprocessing.Process(name=pn,
-                                            target=do_update,
-                                            args=(server.Server.is_rest_server_http_enabled,
-                                                  server.Server._host, server.Server.core_management_port,
-                                                  storage_client, plugin_type, plugin_name, package_name, uid,
-                                                  schedules, notifications))
+                p = multiprocessing.Process(name=pn, target=do_update, args=(
```
praveen-garg, line 151: just replace server.Server.is_rest_server_http_enabled with "http"

python/fledge/services/core/api/plugins/update.py
```diff
@@ -342,10 +339,11 @@ async def _get_sch_id_and_enabled_by_name(name: str) -> list:
     return result['rows']
 
 
-async def _put_schedule(protocol: str, host: str, port: int, sch_id: uuid, is_enabled: bool) -> None:
-    management_api_url = '{}://{}:{}/fledge/schedule/{}/enable'.format(protocol, host, port, sch_id)
+async def _put_schedule(host: str, port: int, sch_id: uuid, is_enabled: bool) -> None:
```
praveen-garg, line 342: revert to old signature

python/fledge/services/core/api/plugins/update.py
```diff
@@ -342,10 +339,11 @@ async def _get_sch_id_and_enabled_by_name(name: str) -> list:
...
+    # Scheme is always http:// on core_management_port
+    management_api_url = 'http://{}:{}/fledge/schedule/{}/enable'.format(host, port, sch_id)
     headers = {'content-type': 'application/json'}
-    verify_ssl = False if protocol == 'HTTP' else True
+    verify_ssl = False
```
praveen-garg, line 346: Yes, this check has no impact; and actually was wrong

python/fledge/services/core/api/plugins/update.py
```diff
@@ -377,12 +375,9 @@ def _update_repo_sources_and_plugin(pkg_name: str) -> tuple:
     return ret_code, link
 
 
-def do_update(http_enabled: bool, host: str, port: int, storage: connect, _type: str, plugin_name: str,
+def do_update(host: str, port: int, storage: connect, _type: str, plugin_name: str,
```
praveen-garg, line 378: keep http_enabled as protocol/scheme in signature

python/fledge/services/core/api/plugins/update.py
```diff
@@ -377,12 +375,9 @@ def _update_repo_sources_and_plugin(pkg_name: str) -> tuple:
...
               pkg_name: str, uid: str, schedules: list, notifications: list) -> None:
     _logger.info("{} package update started...".format(pkg_name))
-    
-    # Protocol is always http:// on core_management_port
-    protocol = "HTTP"
```
praveen-garg, line 385: Yes, let's remove it and make sure caller pass http only

python/fledge/services/core/api/service.py
```diff
@@ -644,10 +644,8 @@ async def update_service(request: web.Request) -> web.Response:
         if result['response'] == "inserted" and result['rows_affected'] == 1:
             pn = "{}-{}".format(action, name)
             # Scheme is always http:// on core_management_port
-            p = multiprocessing.Process(name=pn, target=do_update, args=(server.Server.is_rest_server_http_enabled,
-                                                                         server.Server._host,
-                                                                         server.Server.core_management_port,
-                                                                         storage_client, package_name, uid, sch_list))
+            p = multiprocessing.Process(name=pn, target=do_update, args=(
```
praveen-garg, line 647: As above

python/fledge/services/core/api/service.py
```diff
@@ -665,10 +663,10 @@ async def update_service(request: web.Request) -> web.Response:
         return web.json_response(result_payload)
 
 
-async def _put_schedule(protocol: str, host: str, port: int, sch_id: uuid, is_enabled: bool) -> None:
-    management_api_url = '{}://{}:{}/fledge/schedule/{}/enable'.format(protocol, host, port, sch_id)
+async def _put_schedule(host: str, port: int, sch_id: uuid, is_enabled: bool) -> None:
```
praveen-garg, line 666: As above

python/fledge/services/core/api/service.py
```diff
@@ -682,16 +680,11 @@ async def _put_schedule(protocol: str, host: str, port: int, sch_id: uuid, is_en
             _logger.debug("PUT Schedule response: %s", response)
 
 
-def do_update(http_enabled: bool, host: str, port: int, storage: connect, pkg_name: str, uid: str,
-              schedules: list) -> None:
+def do_update(host: str, port: int, storage: connect, pkg_name: str, uid: str, schedules: list) -> None:
```
praveen-garg, line 683: keep http_enabled as protocol/scheme in signature

Are these duplicate def, across API modules?

python/fledge/services/core/api/plugins/update.py
ashish-jabble: Though, verify SSL matters only if HTTPS

python/fledge/services/core/api/service.py
ashish-jabble: Plugins and Service Update are dealing differently that's why no common def. Only to avoid over complicated things by passing flag.

python/fledge/services/core/api/plugins/update.py
praveen-garg: Yes, As we use self signed certificate, so if that comes into force via protocol then we want to set it to False. with HTTP its setting is useless.

praveen-garg (approved):

PR #1106, opened 2023-07-10
FOGL-7839 Improvements in memory_leakage test script

Mohit04tomar:


pintomax:
This is ok.

We now need to prepare match templates, excluding know issues like Python in south-sinusoid, and raise error if new Valgrind output diverges from expected results

pintomax (approved):

pintomax:
This looks good now.

Are you adding output match in another PR?

PR #1116, opened 2023-07-19
FOGL-7696 Addition of UTC timezone to the python payload builder & also Sqlite plugins readings has with UTC instead of localtime

ashish-jabble:
UTC always with Sqlite Plugins on readings!

As per CI Jobs - unit & system test health is good on both Sqlite & PostgreSql

### DIFF
--- a/C/plugins/storage/sqlite/common/readings.cpp
+++ b/C/plugins/storage/sqlite/common/readings.cpp
@@ -1659,13 +1659,11 @@ vector<string>  asset_codes;
 					sql.append(' ');
 				}
 
-				sql.append("id, asset_code, reading, strftime(')" F_DATEH24_SEC);
-				sql.append("(', user_ts, '");
+				sql.append("id, asset_code, reading, strftime('" F_DATEH24_SEC "', user_ts, '");
 				sql.append(timezone);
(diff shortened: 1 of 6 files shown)


pintomax (approved):

hemantgarg923 (dismissed):

C/plugins/storage/sqlite/common/readings.cpp
```diff
@@ -1644,16 +1659,13 @@ vector<string>  asset_codes;
...
-                    FROM  )";
-
-				sql.append(sql_cmd);
+				sql.append("id, asset_code, reading, strftime(')" F_DATEH24_SEC);
+				sql.append("(', user_ts, '");
+				sql.append(timezone);
+				sql.append("')  || substr(user_ts, instr(user_ts, '.'), 7) AS user_ts, strftime(')" F_DATEH24_MS);
+			       	sql.append("(', ts, '");
+				sql.append(timezone);
+				sql.append("') AS ts FROM  )");
```
ashish-jabble, line 1668: Syntax error here; No need to close the bracket after FROM. Fixed in https://github.com/fledge-iot/fledge/pull/1118

pintomax (approved):

hemantgarg923 (approved):

PR #1124, opened 2023-07-25
FOGL-7953 New control user added

ashish-jabble:


ashish-jabble:
@MarkRiddoch Please have a look on the new addition and let me know for below cases?

a) Shall we need to restrict all operations of Control script and pipelines for normal role user and view role user? At the moment GET allowed
b) Control user can create/update ACL? or only Admin user can do?
c) Any other case?

MarkRiddoch (approved):
Look good to me. We need to be careful about checkin as FOGL-7980 has schema version 62 reserved so we need to get these run in the right order.

To answer your questions
a) Get for other users is fine.
b) Both control and admin. I think as a general rule it makes sense for admin to be able to do everything
c) There is more to come in the next sprint

ashish-jabble:
> Look good to me. We need to be careful about checkin as FOGL-7980 has schema version 62 reserved so we need to get these run in the right order.

That's why I have added empty SQL files for 62 number. While rebasing we can simply do. Assuming that both schema versions will go into develop before 2.2 cut off.

MarkRiddoch (dismissed):
I've merged my branch that had the previous schema version, so once the conflict is resolved we can go ahead and merge this. Thanks.

ashish-jabble:
Conflicts resolved as per latest changes!

MarkRiddoch (approved):

PR #1127, opened 2023-07-27
FOGL-7743: clear the readings vector after append operation

gnandan:


C/services/filter-plugin-interfaces/python/python_plugin_interface.cpp
```diff
@@ -298,7 +298,7 @@ void filter_plugin_ingest_fn(PLUGIN_HANDLE handle, READINGSET *data)
 				data->removeAll();
 
 				// Append filtered readings
-				data->append(filteredReadingSet->getAllReadings());
+				data->append(filteredReadingSet);
```
AmandeepArora, line 301: Maybe update comment also that the append call empties the passed ReadingSet also.

AmandeepArora:
@gnandan : The changeset in this PR seems to contain extra commits, please check.

AmandeepArora (approved):

MarkRiddoch (approved):

PR #1133, opened 2023-08-06
FOGL-8014 Prevent crash in resultset constructor with integer overflow

MarkRiddoch:


RVerhoeff (approved):
This change solves the problem of the fault log messages every 15 seconds in the Menlo Park test system.

PR #1134, opened 2023-08-07
FOGL-7982 monitors and statistics content added into support bundle

ashish-jabble:
Below are the tables added to support bundle

- [x] monitors 
- [x] statistics - added this was missed before


ashish-jabble:
@MarkRiddoch Do we want to limit the records for monitors table; as like statistics history (TOP 1000 rows)?

MarkRiddoch (approved):

PR #1135, opened 2023-08-07
FOGL-7983 Store statistic history in UTC and fix the newer/older

MarkRiddoch:
condition

### DIFF
--- a/C/plugins/storage/sqlite/common/connection.cpp
+++ b/C/plugins/storage/sqlite/common/connection.cpp
@@ -1085,7 +1085,7 @@ vector<string>  asset_codes;
 			 
 				if (document.HasMember("join"))
 				{
-					if (!jsonWhereClause(document["where"], sql, asset_codes, true, "t0."))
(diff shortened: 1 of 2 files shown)


hemantgarg923 (approved):

hemantgarg923:
If we upgrade from latest release package (which already provide UTC time) to this branch package then it is working fine i.e. it provides UTC time. But if we upgrade from nightly package (which provide local time) to this branch package then it still provides local timezone, **it does not change to UTC**. Restarting the FogLAMP does not help, only after resetting the FogLAMP it provides UTC time.

MarkRiddoch:
> If we upgrade from latest release package (which already provide UTC time) to this branch package then it is working fine i.e. it provides UTC time. But if we upgrade from nightly package (which provide local time) to this branch package then it still provides local timezone, **it does not change to UTC**. Restarting the FogLAMP does not help, only after resetting the FogLAMP it provides UTC time.

Correct, it does not change the existing data in the database, but new data is added as UTC

PR #1150, opened 2023-08-21
Use sqlite3 in /usr/local rather than /tmp

MarkRiddoch:


requirements.sh
```diff
@@ -207,6 +207,7 @@ if [[ $YUM_PLATFORM = true ]]; then
 		set -e
 		make
 		# TODO: Use make install to install sqlite3 as a command
```
pintomax, line 209: Remove TODO if useless right now

pintomax (dismissed):

pintomax (approved):

PR #1156, opened 2023-08-24
FOGL-8060 Limit size of insert statements sent to Postgres

MarkRiddoch:


ashwinscale (changes requested):
Unable to run make install.

ashish-jabble:
@ashwinscale  I see this PR has not updated with latest develop and now fully updated. 
`make install` should be fixed now. You may check the on-demand packages also if required and available at `fixes/FOGL-8060/ubuntu2004/x86_64/` for both Fledge/FogLAMP.

gnandan:
Purge process is purging data even without any ingest.  Also in case of ingest statistics shows that purged data is more then ingested data. It seems the purge schedule is increasing statistics count after every run. 
Added steps to reproduce on JIRA

pintomax (approved):

ashwinscale (approved):

PR #1157, opened 2023-08-25
sqlite3 make install fixes for YUM based as per FOGL-8039 changes

ashish-jabble:
The problem is as with previous centos versions we had a condition for sqlite3 pkg `make` failure on very first attempt. So had a check to deal such situation to re-attempting `make` again.

And now with latest sqlite3 we always use `make install` which we forgot to update and call `make install` in  re-attempt condition.

Note: I haven't removed this re-attempt code as it will only call when first try of make fail. So no harmful.


MarkRiddoch (approved):

ashish-jabble:
System tests health are also good.

PR #1158, opened 2023-09-04
FOGL-7839 clean-up memory deallocation

MarkRiddoch:
Also remove copy of ingested readings when using Python south plugins

Fixed leak in north service when no readings are read from the database.

Although not all issues reported are fixed, there are a couple of significant things that make it worth getting this into develop and continuing. 

The vast majority are one-off that would be cleaned up only on service exit anyway, so resolving that no practical impact.


cyberwalk3r (commented):
Getting segmentation fault on Ubuntu 20 when testing via packages. (http://archives.dianomic.com/foglamp/fixes/FOGL-7839/ubuntu2004/x86_64/)
Steps:
1. Add a sinusoid service.
2. Add scale filter.
Following dump is seen in logs after adding filter to sinusoid service.
```
FogLAMP s1[103295]: FATAL: Signal 11 (Segmentation fault) trapped:
FogLAMP s1[103295]: FATAL: (0) 0       0x561be7931d22 handler(int) + 82---------
FogLAMP s1[103295]: FATAL: (1) 1       0x7f298c331090 /lib/x86_64-linux-gnu/libc.so.6(+0x43090) [0x7f298c331090]---------
FogLAMP s1[103295]: FATAL: (2) 2       0x7f2988187746 plugin_ingest + 1142---------
FogLAMP s1[103295]: FATAL: (3) 3       0x561be7928f5a Ingest::processQueue() + 6138---------
FogLAMP s1[103295]: FATAL: (4) 4       0x561be792ad40 ./foglamp.services.south(+0x53d40) [0x561be792ad40]---------
FogLAMP s1[103295]: FATAL: (5) 5       0x7f298c5d3df4 /lib/x86_64-linux-gnu/libstdc++.so.6(+0xd6df4) [0x7f298c5d3df4]---------
FogLAMP s1[103295]: FATAL: (6) 6       0x7f298cb4c609 /lib/x86_64-linux-gnu/libpthread.so.0(+0x8609) [0x7f298cb4c609]---------
FogLAMP s1[103295]: FATAL: (7) 7       0x7f298c40d133 clone + 67---------
```

MarkRiddoch:
The issue with Ubuntu 20 is related to how the job builds the packages. If there is no explicit build of a package requested then it is copied from nightly. However in this case the table had changed, so nightly packages are not compatible and result in issues because the size of the Reading and ReadingSet objects has changed. If a new scale filter is built as part of the fledge_wl_on-demand job and installed, then it all works fine.

cyberwalk3r (approved):
LGTM!

PR #1165, opened 2023-09-15
FOGL-7885: Memory leak on south service with python filter for T_DP_DICT readings

AmandeepArora:
Extensive testing done - with filters on south and north services and also filters on disabled north service in conditional forwarding setup. Tried with C++ and python filters and checked with different order of filters also. Also checked for memory leaks. QA also helped out in this test coverage.


C/services/south/ingest.cpp
```diff
@@ -1002,11 +1008,18 @@ void Ingest::useFilteredData(OUTPUT_HANDLE *outHandle,
...
 {
 	Ingest* ingest = (Ingest *)outHandle;
+
 	if (ingest->m_data != readingSet->getAllReadingsPtr())
 	{
-		if (ingest->m_data)
+		if (ingest->m_data && ingest->m_data->size())
 		{
-			ingest->m_data->clear();// Remove any pointers still in the vector
+			// Remove the readings in the vector
```
MarkRiddoch, line 1016: I'm not sure this will work in all cases. I suspect we might be deleting the readings twice in C++ filters

AmandeepArora: @MarkRiddoch : I tested again with sinusoid south service + asset filter (python) + metadata (C++) and scale (C++) filters, under valgrind, kept changing the filter order, I don't see any double free errors.

MarkRiddoch: That's good, thanks

MarkRiddoch (approved):

PR #1168, opened 2023-09-20
FOGL-8111 execute control Flow API endpoint updated as per new specs

ashish-jabble:


python/fledge/services/core/api/control_service/entrypoint.py
```diff
@@ -435,36 +441,52 @@ async def update_request(request: web.Request) -> web.Response:
     """
     name = request.match_info.get('name', None)
     try:
+        # check the dispatcher installation
```
MarkRiddoch, line 444: It would be better to check by looking up the service registry as this is reliant on the dispatcher running on the same machine/container as the core. Although this will be the case currently the plan is to allow services to be run on different machines/containers at some point in the future.

Likewise, this would also mean you do not need to check the schedules as you can do a single check to simply see if it is running.

ashish-jabble: Sure, service registry check is actually good for Dispatcher service as on Fledge restart we always try to enable Dispatcher service If in disable state.

MarkRiddoch (approved):

PR #1191, opened 2023-10-09
purge tests fixes and updated as per FOGL-8166

ashish-jabble:
As purge tests were too old and some assert statements are NOT good. So fixed that and cover more coverage when we have both Age And Size with purge and retain flag in the configuration set.


MarkRiddoch (approved):
Thanks

Issue #1069, opened 2023-05-15
Erroneous documentation for most plugin functions involving PLUGIN_HANDLE


FlorentP42:
Most of the documentation involving [PLUGIN_HANDLE](https://fledge-iot.readthedocs.io/en/latest/search.html?q=PLUGIN_HANDLE&check_keywords=yes&area=default#
) provides erroneous usage examples for the C++ implementation of a plugin.

The `plugin_init()` function is correct and returns a `PLUGIN_HANDLE` type (`void*`) containing a pointer to the main plugin class (in this example a `DHT11*`):
![image](https://github.com/fledge-iot/fledge/assets/45787476/9808d2ce-80ec-4100-b85c-1c2f61bc5370)

But then most of the other functions in the documentation take a `PLUGIN_HANDLE*` as parameter (`void**`) when it should in fact use a `PLUGIN_HANDLE` only. It is clearly visible that the input type is actually the same type as the one returned by `plugin_init()` when the input type is casted into `DHT11*` here:
![image](https://github.com/fledge-iot/fledge/assets/45787476/aae0d07b-62f7-4e91-a767-90ac6656a086)

All of the documentation involving this type needs to be reviewed: some functions may actually need `PLUGIN_HANDLE*` as an input, but most of them doesn't and must be fixed.

Sadly most of the existing plugins are based on those examples and do contain this erroneous implementation which is not raising any error because it is using the "jocker" `void*` type.

As an example in one of the plugins I reviewed this led to really messed up codes likes this one where you end up casting a `PLUGIN_HANDLE` into `PLUGIN_HANDLE*` which makes zero sense in term of pure C++:
```cpp
TEST(HNZ, PluginInit) {
  ConfigCategory *config = new ConfigCategory("Test_Config", default_config);
  config->setItemsValueFromDefault();
  PLUGIN_HANDLE handle = nullptr;

  ASSERT_NO_THROW(handle = plugin_init(config));

  if (handle != nullptr) plugin_shutdown((PLUGIN_HANDLE *)handle);

  delete config;
}
```

FlorentP42:
Looks like part of this issue was already reported in #843, sorry for missing it.

Issue #843, opened 2022-10-03
Invalid Filter plugin interface


jchabod:
I'm using Fledge v2.0.0

There seem to be an inconsistency in FILTER plugin API:
- https://fledge-iot.readthedocs.io/en/develop/plugin_developers_guide/06_filter_plugins.html states:
```
 * @param handle     The plugin handle returned from plugin_init
...
void plugin_ingest(PLUGIN_HANDLE *handle,
                   READINGSET *readingSet)
{
        LogFilter *log = (LogFilter *) handle;
        log->ingest(readingSet);
}
```

But : `typedef void * PLUGIN_HANDLE;` is defined in `plugin_api.h` 
Thus `handle` is a `void**` and C-style casted to `LogFilter*`, which in no way can be a valid cast. (The comment `The plugin handle returned from plugin_init` is not helping here)

We may infer that either:
- The correct API is `void plugin_ingest(PLUGIN_HANDLE handle, READINGSET *readingSet`
- Or, the CONTENT of handle holds the pointer value : ` LogFilter *log = (LogFilter *) (*handle);`

jchabod:
Note: in `filter_plugin.cpp`, it seems that the correct API is `void plugin_ingest(PLUGIN_HANDLE handle, READINGSET *readingSet`

MarkRiddoch:
Thanks, I suspect this might be a documentation issue. We will need to look into it.

mbourhis:
It seems that issue concerns the *code* of at least one 'filter' plugin too.

These 2 'filter' plugins don't have the same API:

https://github.com/fledge-iot/fledge-filter-change/blob/develop/plugin.cpp
https://github.com/fledge-iot/fledge-filter-log/blob/develop/plugin.cpp

In one case, we can see
(https://github.com/fledge-iot/fledge-filter-change/blob/c82d07a4e520e9b8b4da68fcd5c83221ca718d28/plugin.cpp#L204)
```
void plugin_reconfigure(PLUGIN_HANDLE handle, const string& newConfig)
```
and
(https://github.com/fledge-iot/fledge-filter-change/blob/c82d07a4e520e9b8b4da68fcd5c83221ca718d28/plugin.cpp#L213)
```
void plugin_shutdown(PLUGIN_HANDLE handle)
```

and in the second case : 
(https://github.com/fledge-iot/fledge-filter-log/blob/6fe471d099137f1548f2cca56f43110ed7b767bf/plugin.cpp#L110)
```
[...]

PR #1195, opened 2023-10-13
Documentation: fix C++ South plugin signatures involving PLUGIN_HANDLE

mbourhis:
Fix the issue #1069

In the 'plugin_developers_guide/03_south_C_plugins.html' documentation page, the signature of some plugin functions involving PLUGIN_HANDLE is erroneous.

According to the definitions of the function pointers about the South plugins ($FLEDGE_ROOT/C/services/south/include/south_plugin.h), fix these signatures.

And use the appropriate C++ 'cast' operator,
more restrictive and accurate 'cast' operator than the C one, for the 'PLUGIN_HANDLE' object.


mbourhis:
The code source of the examples have also been updated; see:
 * https://github.com/fledge-iot/fledge-south-random/pull/28
 * DHT plugin (https://github.com/fledge-iot/fledge-south-dht): already done in January 2023

mbourhis:
This PR concerns the 'South plugin' but other documentation pages are also concerned by this 'inconsistency':
* the 'filter' plugins (an issue is already open: https://github.com/fledge-iot/fledge/issues/843)
* the 'notification delivery' plugins (https://fledge-iot.readthedocs.io/en/latest/plugin_developers_guide/08_notify_plugins.html)

mbourhis:
Hi @pintomax, @MarkRiddoch

this PR is about the issue #1069,
where some API plugin signatures involving PLUGIN_HANDLE are erroneous.

So, I am looking for a reviewer for this PR.

Thank you.

mbourhis:
Hi @MarkRiddoch ,

I would like to know if you are ok or not with this commit
and this improvement in the documentation,
because new users who read the docs (readthedocs.io) -like me- may encountered
this subtlety with the "PLUGIN_HANDLE' as "void *". 

Thank you.

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -300,8 +300,8 @@ In C/C++ the *plugin_reconfigure* method is very similar, note however that the
    */
   void plugin_reconfigure(PLUGIN_HANDLE *handle, string& newConfig)
```
FlorentP42, line 301: Can we use a simple PLUGIN_HANDLE instead of PLUGIN_HANDLE* here too? (and remove the * from *handle to go with it?) Or is it mandatory to keep it like that due to the actual API behavior?

FlorentP42 (approved):

docs/plugin_developers_guide/02_writing_plugins.rst
mbourhis: No, we can not use a simple 'PLUGIN_HANDLE' for 'plugin_reconfigure()',
because the function pointers declared in
 * C/services/south/include/south_plugin.h:58 (for the south plugins)
 * C/services/north/include/north_plugin.h:53 (for the north plugins)

require a 'PLUGIN_HANDLE*'

(but for the filter plugins, it seems to be different ( C/common/include/filter_plugin.h:52)

This PR focuses on the North and the South plugins,

but we have to check this mix (between 'PLUGIN_HANDLE' and 'PLUGIN_HANDLE*') for the other types of plugins (as I mentioned here: https://github.com/fledge-iot/fledge/pull/1195#issuecomment-1761189195)

FlorentP42 (approved):

ashish-jabble:
@MarkRiddoch Do we need to merge this into the develop branch?

PR #1193, opened 2023-10-11
FOGL-8175

Mohit04tomar:

PR #1196, opened 2023-10-13
FOGL-8190 Temporarily revert to statistics collection per asset and

MarkRiddoch:
service as the default for south services

This will be reverted after the next release and a GUI change done for the display of south services

### DIFF
--- a/docs/tuning_fledge.rst
+++ b/docs/tuning_fledge.rst
@@ -75,12 +75,15 @@ The south services within Fledge each have a set of advanced configuration optio
     | |stats_options| |
     +-----------------+
 
-    The *per asset & per service* setting will collect one statistic per asset ingested and an overall statistic for the entire service. The *per service* option just collects the overall service ingest statistics and the *per asset* option just collects the statistics for each asset and not for the entire service. The default is to collect statistics on a per service basis, use of the per asset or the per asset and service options should be limited to south service that collect a relatively small number of distinct assets. Collecting large number of statistics, for 1000 or more distinct assets will have a significant performance overhead and may overwhelm less well provisioned Fledge instances.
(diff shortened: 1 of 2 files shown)


docs/tuning_fledge.rst
```diff
@@ -75,12 +75,15 @@ The south services within Fledge each have a set of advanced configuration optio
...
+    The *per asset & per service* setting will collect one statistic per asset ingested and an overall statistic for the entire service. The *per service* option just collects the overall service ingest statistics and the *per asset* option just collects the statistics for each asset and not for the entire service. The default is to collect statistics on a per asset & service basis, this is not the best setting if large numbers of distinct assets are ingested by a single south service. Use of the per asset or the per asset and service options should be limited to south service that collect a relatively small number of distinct assets. Collecting large number of statistics, for 1000 or more distinct assets will have a significant performance overhead and may overwhelm less well provisioned Fledge instances. When a large number of assets are ingested by a single south service this value should be set to *per service*.
```
ashish-jabble, line 78: option is: `per asset & service`

ashish-jabble (approved):

ashish-jabble:
@Mohit04tomar fyi, We need to do updates on https://github.com/fledge-iot/fledge/pull/1193

Smoke tests should be okay with this PR and should be removed from your PR. And for the rest of the tests, we need to add `default=True` in stats collection fixture when True nothing to be changed in Config item else update with payload you need to test with.

MarkRiddoch:
> @Mohit04tomar fyi, We need to do updates on #1193
> 
> Smoke tests should be okay with this PR and should be removed from your PR. And for the rest of the tests, we need to add `default=True` in stats collection fixture when True nothing to be changed in Config item else update with payload you need to test with.

I would suggest we write all tests so that if they need a particular setting, even if it is the default, then they explicitly  set that value. This would make them more robust to defaults being changed.

Mohit04tomar (approved):

PR #1208, opened 2023-10-20
refs #1207 Added log wrapper functions to be able to see logs in unit tests output.

### LINKED ISSUE #1207: Display all Fledge logs in stdout during unit tests
**Is your feature request related to a problem? Please describe.**
Right now the Logger class only logs into specific Fledge output files and not into the standard output, meaning they will not appear when running unit tests.
Having a compile time flag to make all Fledge logs appear in stdout while running unit tests would be quite useful in plugin debugging phase.

**Describe the solution you'd like**
A solution would be to provide log function wrappers that would call both the Logger and the printf() function when a compile time flag is on so that the log is visible in both places.

**Describe alternatives you've considered**
A more advanced approach would be to further upgrade the Logger class into handling multiple types of appenders so that the user could define its own appenders when configuring Fledge (see examples of implementation in https://log4cpp.sourceforge.net/).
This was not retained so far as it would have more impact on existing code and may result in interface evolutions.

**Additional context**
N/A

FlorentP42:
Closes #1207


FlorentP42:
@MarkRiddoch: Please let me know when this can be reviewed.
PS: Any idea why unit test pipelines are not running on this PR?

ashish-jabble:
@MarkRiddoch Do we really need to merge this into the develop branch?

PR #1211, opened 2023-10-24
refs #1210 Forward result of NorthService::sendToDispatcher() to the …

### LINKED ISSUE #1210: Return value of operations not forwarded propelry to north plugins
**Describe the bug**
When calling an operation from a north plugin, `NorthService::operation()` calls `NorthService::sendToDispatcher()` but ignores its return value and always returns -1, making it impossible to determine if an operation was a success from the perspective of the north plugin.

**To Reproduce**
1. Write a south plugin with an operation to call.
2. Write a north plugin calling said operation and print the return value of the call to `m_oper()`.
3. Notice that this return value is always -1.

**Expected behavior**
 `NorthService::operation()` returns -1 in case of internal error (as it already does), 0 in case `NorthService::sendToDispatcher()` failed to properly send the operation to the dispatcher and 1 in case `NorthService::sendToDispatcher()` sent the operation to the dispatcher successfully.

FlorentP42:
…output of NorthService::operation().

Closes #1210


FlorentP42:
@MarkRiddoch: Please let me know when this can be reviewed.
PS: Any idea why unit test pipelines are not running on this PR?

MarkRiddoch (approved):

MarkRiddoch:
Sorry, for some reason I did not get a notification that this was waiting. I've just checked and the unit tests seem to be running now, I did an "Update Branch" to force a rerun of the tests.

FlorentP42:
Can you see any unit test result then? For me it looks like this: 
image

It does not appear like thay are even trying to run...

If you wtill want to mere like this it's fine by me... But I would like to have confirmation through the unit tests that this change did not have any unplanned impacts somewhere else that the unit tests could reveal.

MarkRiddoch (approved):

PR #1217, opened 2023-11-01
FOGL-8222 Improve error and warning reporting when updates and inserts

MarkRiddoch:
require retries


pintomax (approved):

ashwinscale (changes requested):
Details in JIRA.

MarkRiddoch:
> Details in JIRA.

The details in the Jira are about something different. I suggest we merge this change and address the issue added to the Jira in a new Jira or the current FOGL-8011 Jira

PR #1220, opened 2023-11-17
System test fixes

Mohit04tomar:
Below the jobs that are fixed after changes of this PR:

1. http://jenkins.dianomic.com:8080/view/PI%20stuff/job/fledge_packages_system_test_north_pi_webapi_nw_throttle_piwebapi2021sp2/296/
2. http://jenkins.dianomic.com:8080/job/fledge_packages_system_test_data_availability_notification_rule/80/

### DIFF
--- a/tests/system/python/packages/test_north_pi_webapi_nw_throttle.py
+++ b/tests/system/python/packages/test_north_pi_webapi_nw_throttle.py
@@ -329,7 +329,7 @@ class TestPackagesSinusoid_PI_WebAPI:
             raise Exception("None of packet delay or rate limit given, "
                             "cannot apply network impairment.")
         # Insert some readings before turning off compression.
-        time.sleep(2)
+        time.sleep(3)
         # Turn off south service
         disable_schedule(fledge_url, SOUTH_SERVICE_NAME)
         time.sleep(5)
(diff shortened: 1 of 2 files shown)


praveen-garg (approved):

tests/system/python/packages/test_north_pi_webapi_nw_throttle.py
```diff
@@ -329,7 +329,7 @@ def test_omf_in_impaired_network(self, clean_setup_fledge_packages, reset_fledge
             raise Exception("None of packet delay or rate limit given, "
                             "cannot apply network impairment.")
         # Insert some readings before turning off compression.
-        time.sleep(2)
+        time.sleep(3)
```
praveen-garg, line 332: May be better safe to set 5s

ashish-jabble: Anything change in Core or Plugin itself? Seems like it is required for system test only for ingestion of readings?
We have a fixture named as **wait_time** and having default `5s` value.

tests/system/python/packages/test_rule_data_availability.py
```diff
@@ -263,6 +263,7 @@ def test_data_availability_north(self, check_eds_installed, reset_fledge, start_
         get_url = "/fledge/audit?source=NTFSN"
         resp1 = utils.get_request(fledge_url, get_url)
         
+        time.sleep(wait_time)
```
ashish-jabble, line 266: We should look for a viable solution NOT a hardcoded wait time interim solution. 
We may add poll mechanism to receive the notification and retry based check in expected time

PR #1222, opened 2023-11-17
FOGL-8167 Performance Counters API addition

ashish-jabble:
DOC Pending queries resolved with FOGL-8275


MarkRiddoch (approved):
Looks good to me. Let's get this checked in without the todo items as it gives the first level of access to the counters. We will need to think about how we want to visualise these later.

PR #1228, opened 2023-11-23
FOGL-8282: now() is the only supported function in postgres plugin

pintomax:
FOGL-8282: now() is the only supported function in postgres plugin


pintomax:
@Mohit04tomar would you mind setting up all the system tests in order to verify there are no other hidden issues?

Mohit04tomar (approved):

PR #1229, opened 2023-11-28
FOGL-8290: Update configuration manage cache for new config item.

cyberwalk3r:


praveen-garg (approved):

python/fledge/services/core/api/configuration.py
```diff
@@ -426,6 +426,10 @@ async def add_configuration_item(request):
         result = await storage_client.update_tbl("configuration", payload)
         response = result['response']
 
+        # update cache with new config item
```
pintomax, line 429: I have tested similar change with an extra check before the update

if new_config_item not in cf_mgr._cacheManager.cache[category_name]['value']:

pintomax (changes requested):

python/fledge/services/core/api/configuration.py
cyberwalk3r: With update function this extra check is not required, as it will update the existing value or create one if it doesn't exist.

MarkRiddoch (approved):

pintomax (approved):

PR #1232, opened 2023-12-04
FOGL-8249: C++ Configuration Item support for model selection

AmandeepArora:
Tested with modified benchmark plugin and sinusoid plugin.

### DIFF
--- a/C/services/common/service_security.cpp
+++ b/C/services/common/service_security.cpp
@@ -608,7 +608,7 @@ void ServiceAuthHandler::refreshBearerToken()
 			// Shutdown service
 			if (m_refreshRunning)
 			{
-				Logger::getLogger()->warn("Service is being shut down " \
+				Logger::getLogger()->warn("Service is being restarted " \
 						"due to bearer token refresh error");
 				this->restart();
 				break;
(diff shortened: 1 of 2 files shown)


C/common/config_category.cpp
```diff
@@ -1587,6 +1621,25 @@ ostringstream convert;
 	return convert.str();
 }
 
+vector<pair<string,string>>* ConfigCategory::parseBucketItemValue(const string & json)
```
MarkRiddoch, line 1624: A doxygen header here would be useful

MarkRiddoch (dismissed):

C/common/config_category.cpp
AmandeepArora: Added

C/common/config_category.cpp
```diff
@@ -1095,7 +1118,7 @@ ConfigCategory::CategoryItem::CategoryItem(const string& name,
 		}
 	}
 
-	std:string m_typeUpperCase = m_type;
+	std::string m_typeUpperCase = m_type;
```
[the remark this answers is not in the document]
AmandeepArora: In "std:string" case, std was treated as a goto label. 

https://stackoverflow.com/questions/52049468/why-can-stdset-with-a-single-colon-compile

MarkRiddoch (approved):

PR #1264, opened 2024-01-10
FOGL-4102: fix for aggregate array with timestamp

pintomax:
FOGL-4102: fix for aggregate array with timestamp


MarkRiddoch (approved):

praveen-garg:
Don't we have unit tests around this?


What are the possible regression areas?

MarkRiddoch:
@pintomax This PR will soon be two years old! Are you going for a record here? 

Please look and determine if this is actually needed, is obsolete or has been covered since by another Jira.

Let's not allow this to go into a third year!

PR #1271, opened 2024-01-16
FOGL-8414: Pass management client URL/port in filter config

AmandeepArora:
Made this change to enable python filters to connect to bucket service and get model files.


MarkRiddoch (approved):
Not thrilled with this as a mechanism as exposes what should be private, internal information into the public API and GUI. Let's use this in the short term, but think about how to avoid this for a final solution. Let's not document its use either.

AmandeepArora:
> Not thrilled with this as a mechanism as exposes what should be private, internal information into the public API and GUI. Let's use this in the short term, but think about how to avoid this for a final solution. Let's not document its use either.

Sure, I think a better way would be to get the singleton BucketClient class instance and make calls on that from python plugin/filter code. Will look into that in future.

PR #1273, opened 2024-01-22
FOGL-8439 Alerts feature

ashish-jabble:
- [x] GET Alerts admin API
- [x] DELETE Alerts admin API
- [x] Ping response updated
- [x] SQLite schema changes for alerts table 
- [x] existing tests updated
- [x] upgrade and downgrade scripts
- [x] Sqlitelb 
- [x] PostgreSQL
- [x] FOGL-8440 Python management client entrypoints for alert

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,2 +1,2 @@
 fledge_version=2.3.0
-fledge_schema=67
+fledge_schema=68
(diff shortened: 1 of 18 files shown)


MarkRiddoch (commented):
Looks good top me. I did update the spec following your comments about use of DELETE and also dropped the /acknowledge from the URL as it didn't;t seem that it was really needed if we use the DLETE operation.

praveen-garg:
I assume /acknowledge is more for read/unread. There could be possible options - hide, delete or read, so that a user who read does not see as unread on next reload/ via other client (state)

MarkRiddoch (approved):

PR #1282, opened 2024-02-02
FOGL-8441 : method to getalert and raise alert added to management client

gnandan:
getAlertByKey and raiseAlert method has been introduced into management client to get alert by key and raise new alert.


C/common/management_client.cpp
```diff
@@ -2059,3 +2059,67 @@ int ManagementClient::validateDatapoints(std::string dp1, std::string dp2)
...
+	try
+	{
+		std::string url = "/fledge/alert/" + urlEncode(key) ;
+		auto res = this->getHttpClient()->request("GET", url.c_str());
+		Document doc;
+		response = res->content.string();
+		doc.Parse(response.c_str());
+		if (doc.HasParseError())
+		{
+			bool httpError = (isdigit(response[0]) && isdigit(response[1]) && isdigit(response[2]) && response[3]==':');
```
MarkRiddoch, line 2081: You don't need to do this. It is best to check res->status_code before trying to parse the GET response.

C/common/management_client.cpp
```diff
@@ -2059,3 +2059,67 @@ int ManagementClient::validateDatapoints(std::string dp1, std::string dp2)
...
+bool ManagementClient::raiseAlert(const std::string& payload)
+{
+	try
+	{
+		std::string url = "/fledge/alert" ;
+		auto res = this->getHttpClient()->request("POST", url.c_str(), payload);
+		Document doc;
+		std::string response = res->content.string();
+		doc.Parse(response.c_str());
+		if (doc.HasParseError())
```
MarkRiddoch, line 2109: As above use res->status_code

C/common/management_client.cpp
```diff
@@ -2059,3 +2059,67 @@ int ManagementClient::validateDatapoints(std::string dp1, std::string dp2)
...
+}
+
+
+/**
+ * Raise an alert
+ *
+ * @param    payload        Alert payload
+ * @return   whether operation was successful
+ */
+bool ManagementClient::raiseAlert(const std::string& payload)
```
MarkRiddoch, line 2100: Don't make the caller create the JSON, better to take key, urgency and message as arguments. The client is then the only place the exact payload format is required.

C/common/include/management_client.h
```diff
@@ -117,6 +117,8 @@ class ManagementClient {
 								const std::string& event);
 		int 			validateDatapoints(std::string dp1, std::string dp2);
 		AssetTrackingTable	*getDeprecatedAssetTrackingTuples();
+		std::string		getAlertByKey(const std::string& key);
+		bool			raiseAlert(const std::string& payload);
```
MarkRiddoch, line 121: Better to pass key, message and urgency, maybe even with a default value for urgency.

MarkRiddoch (changes requested):

MarkRiddoch (approved):

PR #1292, opened 2024-02-21
FOGL-8461 Interim checkin to make safe if pipeline branches are created

MarkRiddoch:
This is an interim checkin that is here to protect against users creating non-linear pipelines. It is only here to allow the API to be checked in without risk of crashes in Fledge if the API is used before the backend is completed.

A couple of other classes are included in this checkin but are not yet used.


ashish-jabble:
@MarkRiddoch Do you mind to update the base branch?

ashish-jabble (approved):
Got the warning as expected when we have multiple filter branches associated with service.

```
Fledge Sine[32280]: WARNING: This version of Fledge does not support branching of pipelines. The branch will be ignored.
```
Linear filter pipelines case also works.

praveen-garg:
@hemantgarg923 can you please verify this with https://github.com/fledge-iot/fledge-gui/pull/380

hemantgarg923 (approved):

PR #1294, opened 2024-02-22
FOGL-8389: Problem changing data type formats

RVerhoeff:
The original bug report ([FOGL-8389](https://scaledb.atlassian.net/browse/FOGL-8389)) noted that disabling an _integer_ OMFHint generated errors and caused data to stop flowing to the PI Server. It turns out that OMF causes the PI Server to create AF Attributes and PI Points based on the original data type information. OMF cannot change these data types after the PI entities are created. The only change in this Pull Request is documentation to this effect. A procedure for detecting the problem and repairing the PI Server is included.

### DIFF
--- a/docs/OMF.rst
+++ b/docs/OMF.rst
@@ -235,6 +235,7 @@ Formats & Types
 ~~~~~~~~~~~~~~~
 
 The *Formats & Types* tab provides a means to specify the detail types that will be used and the way complex assets are mapped to OMF types to also be configured.
+See the section *Numeric Data Types* below for more information on configuring data types.
 
 +--------------+
 | |OMF_Format| |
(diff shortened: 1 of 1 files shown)


docs/OMF.rst
```diff
@@ -235,6 +235,7 @@ Formats & Types
 ~~~~~~~~~~~~~~~
 
 The *Formats & Types* tab provides a means to specify the detail types that will be used and the way complex assets are mapped to OMF types to also be configured.
+See the section *Numeric Data Types* below for more information on configuring data types.
```
MarkRiddoch, line 238: I think we should really make these links, see the LinkedTypes for an example.

MarkRiddoch (changes requested):

MarkRiddoch (approved):

docs/OMF.rst
ashish-jabble: reference link issue fixed in PR-https://github.com/fledge-iot/fledge/pull/1302

PR #1300, opened 2024-03-06
FOGL-8533 : Handle "null" string

gnandan:


MarkRiddoch:
How is the null without quotes getting put in the JSON payload in the first place? I think the real issue is what puts this in and not that we change the payload to add the quotes.

gnandan:
> How is the null without quotes getting put in the JSON payload in the first place? I think the real issue is what puts this in and not that we change the payload to add the quotes.

RapidJSON is converting it to null without quotes.

MarkRiddoch:
This is very surprising functionality for RapidJSON to implement. It would violate the integrity of the JSON encoding.  It would also imply that there are several other places we would need this fix. Modifying one of the example they have shows that it does not stopped quotes from strings when the string value is "null".

`// JSON simple example
// This example does not handle errors.

#include "rapidjson/document.h"
#include "rapidjson/writer.h"
#include "rapidjson/stringbuffer.h"
#include <iostream>

using namespace rapidjson;

int main() {
    // 1. Parse a JSON string into DOM.
    const char* json = "{\"project\":\"rapidjson\",\"stars\":10, \"test\" : \"null\"}";
    Document d;
    d.Parse(json);

    // 2. Modify it by DOM.
    Value& s = d["stars"];
    s.SetInt(s.GetInt() + 1);

    // 3. Stringify the DOM
    StringBuffer buffer;
    Writer<StringBuffer> writer(buffer);
    d.Accept(writer);

    // Output {"project":"rapidjson","stars":11}
    std::cout << buffer.GetString() << std::endl;

    if (d["test"].IsNull())
    {
	    std::cout << "Quotes have been stripped" << std::endl;
    }
    else if (d["test"].IsString())
    {
	    std::cout << "Quotes have not been stripped " << d["test"].GetString() << std::endl;
    }
    return 0;
}`

Can you produce an example that illustrates the issue in RapidJSON please?

MarkRiddoch (changes requested):
Please add example of misbehaviour in RapidJSON.

gnandan:
Following is the sample program using RapidJSON with output to simulate situation in fledge storage plugin code 
`#include "rapidjson/document.h"
#include "rapidjson/writer.h"
#include "rapidjson/stringbuffer.h"
#include <iostream>
#include <cstring>

using namespace rapidjson;

int main() {
// 1. Parse a JSON string into DOM.
char nullString[5];
char buff[1024];

// Following statement char* str = (char *)sqlite3_column_text(pStmt, i); gives four characters null without quotes from sqliteDB

strcpy(nullString,"null"); // This statement simulate same situation as char* str = (char *)sqlite3_column_text(pStmt, i); statement

//If we don't add have null as \"null\" then it will get converted to null character instead of a string "null"

strcpy(buff, "{\"project\":\"rapidjson\",\"stars\":10, \"test\" :");
const char* json = strcat(strcat(buff,nullString), "}");

Document d;
d.Parse(json);

// 2. Modify it by DOM.
Value& s = d["stars"];
s.SetInt(s.GetInt() + 1);

// 3. Stringify the DOM
StringBuffer buffer;
Writer<StringBuffer> writer(buffer);
d.Accept(writer);

// Output {"project":"rapidjson","stars":11}
std::cout << buffer.GetString() << std::endl;

if (d["test"].IsNull())
{
    std::cout << "Quotes have been stripped" << std::endl;
}
else if (d["test"].IsString())
{
    std::cout << "Quotes have not been stripped " << d["test"].GetString() << std::endl;
}
return 0;
}

output 
=====

./sample
{"project":"rapidjson","stars":11,"test":null}
Quotes have been stripped
`

MarkRiddoch:
> Following is the sample program using RapidJSON with output to simulate situation in fledge storage plugin code
> 
> ```
> #include "rapidjson/document.h"
> #include "rapidjson/writer.h"
> #include "rapidjson/stringbuffer.h"
> #include <iostream>
> #include <cstring>
> 
> using namespace rapidjson;
> 
> int main() {
> // 1. Parse a JSON string into DOM.
> char nullString[5];
> char buff[1024];
> 
> // Following statement char* str = (char *)sqlite3_column_text(pStmt, i); 
> //gives four characters null without quotes from sqliteDB
> 
> // Below statement simulate same situation as char* str = (char *)sqlite3_column_text(pStmt, i); statement
> strcpy(nullString,"null"); 
> 
> //If we don't add quotes explicitly to null as \"null\" 
> //then it will get converted to null character instead of a string "null"
> 
> strcpy(buff, "{\"project\":\"rapidjson\",\"stars\":10, \"test\" :");
> const char* json = strcat(strcat(buff,nullString), "}");
> 
> Document d;
> d.Parse(json);
> 
> // 2. Modify it by DOM.
> Value& s = d["stars"];
> s.SetInt(s.GetInt() + 1);
> 
> // 3. Stringify the DOM
> StringBuffer buffer;
> Writer<StringBuffer> writer(buffer);
> d.Accept(writer);
> 
> // Output {"project":"rapidjson","stars":11}
> std::cout << buffer.GetString() << std::endl;
> 
> if (d["test"].IsNull())
> {
>     std::cout << "Quotes have been stripped" << std::endl;
> }
> else if (d["test"].IsString())
> {
>     std::cout << "Quotes have not been stripped " << d["test"].GetString() << std::endl;
> }
> return 0;
> }
> 
> output  
> 
> ./sample
> {"project":"rapidjson","stars":11,"test":null}
> Quotes have been stripped
> ```

This example is not right, it creates a string that looks like
`{"project":"rapidjson","stars":10, "test" : null}`
and passed that to RapidJSON, so of course it treats test as a null.  There are no quotes to be stripped. This is not what we are doing when we give something the name "null". Your example is just showing me that RapidJSON is doing the right thing.

gnandan:
@MarkRiddoch For exact reference I used the code of mapResultSet function from sqlite storage plugin. Following code shows when null is passed as parameter than output is without quote for any other value except null output is with quotes.  It shows Rapid JSON treat string "null" differently. 

```
#include "rapidjson/document.h"
#include "rapidjson/writer.h"
#include "rapidjson/stringbuffer.h"
#include <iostream>
#include <string>

using namespace rapidjson;

void mapResultSet(std::string& resultSet,  unsigned long *rowsCount, std::string val1)
{
// JSON generic document
Document doc;

// Number of returned rows, number of columns
unsigned long nRows = 0, nCols = 0;

	// Create the JSON document
	doc.SetObject();
	// Get document allocator
	Document::AllocatorType& allocator = doc.GetAllocator();
	// Create the array for returned rows
	Value rows(kArrayType);
	// Rows counter, set it to 0 now
	Value count;
	count.SetInt(0);

    // Create the 'row' object
    Value row(kObjectType);

    // JSON document for the current row
    Document d;
    // Set object name as the column name
    Value name("Category", allocator);
    // Get the "TEXT" value of the column value
    char* str = (char *)val1.c_str();

    Value value;
    if (!d.Parse(str).HasParseError())
    {
        if (d.IsNumber())
        {
            // Set string
            value = Value(str, allocator);
        }
        else
        {
            // JSON parsing ok, use the document
            value = Value(d, allocator);
        }
    }
    else
    {
        // Use (char *) value
        value = Value(str, allocator);
    }
    // Add name & value to the current row
    row.AddMember(name, value, allocator);

    

    // All fields added: increase row counter
    nRows++;

    // Add the current row to the all rows object
    rows.PushBack(row, allocator);
	

	if (rowsCount != nullptr)
	{
		*rowsCount = nRows;
	}

	// All rows added: update rows count
	count.SetInt(nRows);

	// Add 'rows' and 'count' to the final JSON document
	doc.AddMember("count", count, allocator);
	doc.AddMember("rows", rows, allocator);

	/* Write out the JSON document we created */
	StringBuffer buffer;
	Writer<StringBuffer> writer(buffer);
	doc.Accept(writer);

	// Set the result as a CPP string 
	resultSet = buffer.GetString();

}

int main(int argc, char* argv[]) {

    std::string resultSet;
    unsigned long rowsCount;
    mapResultSet(resultSet, &rowsCount, argv[1] );
    std::cout << resultSet << std::endl;
    return 0;
}

Output:

NOTE: input value sinusoid is with quote in output
./sample sinusoid
{"count":1,"rows":[{"Category":"sinusoid"}]} 

NOTE : input value null is without quote
./sample null
{"count":1,"rows":[{"Category":null}]} 

```

MarkRiddoch:
This is not fixing the route cause of the problem

PR #1307, opened 2024-03-11
FOGL-8521 Add configuration item type information

MarkRiddoch:


docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -358,8 +358,12 @@ We have used the properties *type* and *default* to define properties of the con
      - A description of the configuration item used in the user interface to give more details of the item. Commonly used as a mouse over help prompt.
    * - displayName
      - The string to use in the user interface when presenting the configuration item. Generally a more user friendly form of the item name. Item names are referenced within the code.
+   * - items
+     - The type of the items in a list or kvlist configuration item.
```
ashish-jabble, line 362: Explain properties For Object List type and type bucket?

For example:
 ```
type:list/kvlist
items: object
properties:{...}

OR
type: bucket
properties: {...}
```
And also explain the contract for properties...

File: docs/plugin_developers_guide/02_writing_plugins.rst [hunk 2]
```diff
@@ -385,6 +389,52 @@ We have used the properties *type* and *default* to define properties of the con
 
 Of the above properties of a configuration item *type*, *default* and *description* are mandatory, all other may be omitted.
```
ashish-jabble, line 390: We may add a note directive here!

And for text `all other may be omitted.` Better to say all other are optional

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -385,6 +389,52 @@ We have used the properties *type* and *default* to define properties of the con
...
+   * - boolean
+     - A boolean value that can be assigned the values *true* or *false*.
+   * - IPv4
+     - An IP version 4 address.
+   * - IPv6
+     - An IP version 6 address.
+   * - X509 certificate
+     - An X509 certificate
+   * - password
+     - A string that is used as a password. There is not difference between this or a string type other than user interfaces do not show this in plain text.
```
ashish-jabble, line 417: not difference? Or no difference

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -385,6 +389,52 @@ We have used the properties *type* and *default* to define properties of the con
...
+   * - JSON
+     - A JSON document. The value is checked to ensure it is a valid JSON document.
+   * - URL
+     - A universal resource locator string. The API Will check for correct URL formatting of the value.
```
ashish-jabble, line 421: lowercase `will`

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -385,6 +389,52 @@ We have used the properties *type* and *default* to define properties of the con
...
+   * - enumeration
+     - The item can be assigned one of a fixed set of values. These values are defined in the *options* property of the item.
+   * - script
+     - A block of text that is executed as a script. In this case the script is not stored on the database, but as an external file.
+   * - northTask
+     - The name of a north task. The API will check that the value matches the name of an existing north task.
+   * - ACL
+     - An access control list. The value is the string name of an access control list that has been created within Fledge.
+   * - list
+     - A list of items, the items can be of type *string*, *integer* or *float*. The type of the items within the list must all be the same, and this is defined via the *items* property of the list. A limit on the maximum number of entries allowed in the list can be enforced by use of the *listSize* property.
```
ashish-jabble, line 431: can be of type - + *enumeration* as well

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -385,6 +389,52 @@ We have used the properties *type* and *default* to define properties of the con
...
+   * - kvlist
+     - A key value pair list. The key is a string value always but the value of the item in the list may be of type *string*, *enumeration*, *float* or *integer*. The type of the values in the kvlist is defined by the *items* property of the configuration item. A limit on the maximum number of entries allowed in the list can be enforced by use of the *listSize* property.
+
```
ashish-jabble, line 434: Add other types like
- code
- bucket

docs/plugin_developers_guide/02_writing_plugins.rst
MarkRiddoch: Buckets are not in Fledge, hence they are missing here. I had assumed we did not have support for objects yet, but if we do I will add them.

docs/plugin_developers_guide/02_writing_plugins.rst
MarkRiddoch: Bucket is FogLAMP only really, so will not be covered. Not sure about code, I've never used it and was not aware it was there.

docs/plugin_developers_guide/02_writing_plugins.rst
MarkRiddoch: My assumption on object was based on the fact it is not included in the definition of _valid_type_strings in configuration_manager.py. However I see it is elsewhere in that file.

MarkRiddoch: Added section on object lists

ashish-jabble: > Buckets are not in Fledge, hence they are missing here.

Even that be tricky to be handled in white-labelled one's

docs/plugin_developers_guide/02_writing_plugins.rst
ashish-jabble: > code

yeah it has been added for simple Python code to be passed with

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -343,7 +343,114 @@ The configuration items within a category are JSON object, the object key is the
...
+An example use of an object list might allow for a map structure to be built for accessing a device like a PLC. The following shows the defintions of a key/value pair list where the value is an object.
+
+.. code-block:: JSON
+
+  "map": {
+        "description": "A list of datapoints to read and PLC register definitions",
+        "type": "kvlist",
+        "items" : "object",
+        "default": "{}",
```
ashish-jabble, line 408: cannot be empty here; as its properties exist `register, width, type`

ashish-jabble (commented):
Other typos and some fixes required

docs/plugin_developers_guide/02_writing_plugins.rst
MarkRiddoch: Yes, but what I really do not want to do is start putting bucket stuff in Fledge. We will need to do something, but maybe we limit it Ito a section of the documentation about storing different data types, i.e. models, metrics and MATLAB compiled scripts

ashish-jabble: yeah, separate section can be achievable with white-labelled

PR #1308, opened 2024-03-11
FOGL-8540 Add documentation for the update checker

MarkRiddoch:


docs/quick_start/update.rst
```diff
@@ -0,0 +1,41 @@
...
+Package Updates
+===============
+
+Fledge will periodically check for updates to the various packages that are installed. If updates are available then this will be indicated by a status indicating on the bar at the top of the Fledge GUI.
+
++---------+
+| |alert| |
++---------+
+
+Clicking on the *bell* icon will display the current system alerts, including the details of the packages available to be updated.
```
ashish-jabble, line 13: Have you noticed periodic task upgrade alert is generating per each package?

For development purpose (i.e nightly or develop) that would makes sense But for **Production** it won't make any sense to get alerts per package. Because whenever we release new version of fledge, we released other packages as well with condition `>=<fledge.version>` ! 

**For example:**

Let say, User have installed fledge version 2.3.1 (_in which alert periodic feature is there_) along with 
10 packages.

Now, on the periodic task run we found the new upgrades (let say 2.3.2 version is available). And periodic task will generate 10 alerts for the same stuff.
```
{
"alerts": [
{
"key": “fledge-gui",
"message": "A newer version of fledge-gui is available for upgrade",
"urgency": "Normal",
"timestamp": "2024-03-12 07:30:55.232"
},
{
"key": “fledge”,
"message": "A newer version of fledge is available for upgrade",
"urgency": "Normal",
"timestamp": "2024-03-12 07:30:55.238"
},
{
"key": “fledge-south-sinusoid”,
"message": "A newer version of fledge-south-sinusoid is available for upgrade",
"urgency": "Normal",
"timestamp": "2024-03-12 07:30:55.238"
},

…
…
{
"key": “fledge-service-notification”,
"message": "A newer version of fledge-service-notification is available for upgrade",
"urgency": "Normal",
"timestamp": "2024-03-12 07:30:55.238"
},
…
…

]
}
```
I believe for Production it does not make any sense to generate and show 10 alerts to users. Because when user will try to upgrade fledge either via `curl -X PUT http://localhost:8081/fledge/update` or on command line `apt install fledge` the fledge will automatically upgrade along with the dependent packages.

I believe, periodic task should only generate a single alert for version upgrade:
```
{
"key": “upgrades”,
"message": "A newer version 2.3.2 of fledge is available for upgrade",
"urgency": "Normal",
"timestamp": "2024-03-12 07:30:55.238"
}
```
OR 
`message: "A newer version of fledge, fledge-gui, fledge-south-sinusoid, fledge-service-notification, ....  is available for upgrade",`

docs/quick_start/update.rst
```diff
@@ -0,0 +1,41 @@
...
+------------------
+
+Updates must either be installed manually from the command line or via the Fledge API. To update via the API a call to the */fledge/update* should be made using the PUT method.
+
+.. code-block:: console
+
+   curl -X PUT http://localhost:8081/fledge/update
+
+If the Fledge instance has been configured to require authentication then a valid authentication token must be passed in the request header and that authentication token must by for a user with administration rights on the instance.
+
```
ashish-jabble, line 25:
```
Better to provide the example

.. code-block:: console

    curl -H "authorization: <token>" -X PUT http://localhost:8081/fledge/update


```

docs/quick_start/update.rst
MarkRiddoch: Yes, I think the update checker needs to be changed as it is not what was intended.

docs/quick_start/update.rst
```diff
@@ -0,0 +1,41 @@
...
+.. code-block:: console
+
+   apt upgrade
+
+Or for the *yum* package manager
+
+.. code-block:: console
+
+   yum upgrade
```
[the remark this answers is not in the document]
MarkRiddoch: I'm hoping we will have Centos 9 soon and hence we will need this, We mention yum in other places, so I would say let's leave it here and do the CentOS ASAP

docs/quick_start/update.rst
MarkRiddoch: Added

docs/quick_start/update.rst
```diff
@@ -0,0 +1,45 @@
...
+
+.. code-block:: console
+
+    curl -H "authorization: <token>" -X PUT http://localhost:8081/fledge/update
+
+Manual updates can be down from the command line using the appropriate package manager for your Linux host. If using the *apt* package manager then the command would be
+
+.. code-block:: console
+
+   apt upgrade --only-upgrade 'fledge*'
```
ashish-jabble, line 34: `apt install --only-upgrade fledge*`

docs/quick_start/update.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Or for the *yum* package manager
+
+.. code-block:: console
+
+   yum upgrade
```
ashish-jabble, line 40: Most likely for RPM based ; `yum upgrade fledge*`

docs/quick_start/update.rst
ashish-jabble: Please create a JIRA as fit!

ashish-jabble: I see FOGL-8579, Thanks!

ashish-jabble (approved):

PR #1316, opened 2024-03-20
FOGL-8596: avoid core server locks with POST proxy API

pintomax:
FOGL-8596: avoid core server locks with POST proxy API

### DIFF
--- a/python/fledge/services/core/proxy.py
+++ b/python/fledge/services/core/proxy.py
@@ -180,30 +180,14 @@ async def _call_microservice_service_api(
                         _logger.error("GET Request Error: Http status code: {}, reason: {}, response: {}".format(
                             resp.status, resp.reason, message))
         elif request.method == 'POST':
-            payload = await request.post()
-            if 'multipart/form-data' in request.headers['Content-Type']:
-                import requests
-                from requests_toolbelt.multipart.encoder import MultipartEncoder
-                from aiohttp.web_request import FileField
-                multipart_payload = {}
-                for k, v in payload.items():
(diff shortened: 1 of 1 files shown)


ashish-jabble (changes requested):
Nope!

pintomax:
Tested against

https://github.com/dianomic/foglamp-service-bucketstorage/pull/38

pintomax:
Note: this suggested fix shows the core is no longer locked during autit log operation from bucket service.
Current code works only for for text bucket files

python/fledge/services/core/proxy.py
```diff
@@ -180,30 +180,14 @@ async def _call_microservice_service_api(
...
-            else:
-                payload = await request.json()
-                async with aiohttp.ClientSession() as session:
-                    async with session.post(url, data=json.dumps(payload), headers=headers) as resp:
-                        message = await resp.text()
-                        response = (resp.status, message)
-                        if resp.status not in range(200, 209):
-                            _logger.error("POST Request Error: Http status code: {}, reason: {}, response: {}".format(
-                                resp.status, resp.reason, message))
+            payload = request.content
```
ashish-jabble, line 183: We should do streaming processing using multipart to avoid unwanted footprint which may be significant if you’re dealing with large files, and heavy I/O operation.

ashish-jabble:
Covered all of the scenarios in PR- https://github.com/fledge-iot/fledge/pull/1317

Also we have another solution at[ C-side only](https://github.com/dianomic/foglamp-service-bucketstorage/pull/39) to call audit trail entry in a separate thread.

ashish-jabble:
Closed in favor of https://github.com/fledge-iot/fledge/pull/1317

PR #1320, opened 2024-03-27
FOGL-8573 filtered out manage based packages list in GET update API

ashish-jabble:


MarkRiddoch (commented):
Slightly odd to see fledge-manage, but I get why

ashish-jabble:
> Slightly odd to see fledge-manage, but I get why

This happens only when manage and white labelled product installation via packages on the same machine. So this change is to be filtered out records having `-manage` prefix related packages. Otherwise user may confuse with the upgradable list of packages in GET API as it is related only to white labelled one NOT with the manage one's.

MarkRiddoch (approved):

PR #1321, opened 2024-03-27
Fix JSON unescaping performance issue

geofjamg:
When escaping a large json string, we get a performance issue. 
I tested with a 30000 assets data exchange json and unescaping last around 20s while with new code only 20ms (1000 times faster).


MarkRiddoch (approved):
Thank you for this, the change looks good to me.

PR #1325, opened 2024-03-29
FOGL-8579: Changed logic to raise single alert for all the updates

gnandan:


C/tasks/check_updates/check_updates.cpp
```diff
@@ -75,21 +75,13 @@ void CheckUpdates::raiseAlerts()
 	m_logger->debug("raiseAlerts running");
 	try
 	{
-		for (auto item: getUpgradablePackageList())
+		int availableUpdates = getUpgradablePackageList().size();
+		std::string key = "package_updates";
```
praveen-garg, line 79: We must standardised the keys with enum. 

We must have a details or metadata associated with each alert for meaningful context. 

/cc @MonikaSharma06

MarkRiddoch: Neither of these are things we are doing as part of this Jira.

Keys are very deliberately not enumerated values and will stay that way.

MarkRiddoch (approved):

MarkRiddoch (approved):

PR #1326, opened 2024-04-01
FOGL-8625: Fix for properly incrementing stream ID in north service when a filter creates new reading(s)

AmandeepArora:
Fix for properly incrementing stream ID when a filter on north service adds new readings which have 0 id and that, previously, caused last send ID logic to fail.


MarkRiddoch (approved):
Looks good to me

PR #1328, opened 2024-04-02
FOGL-8609: Support for 64-bit reading ID

AmandeepArora:


MarkRiddoch (dismissed):
A long is also 64bits on Intel, so must of this will not change anything. Only the ARM platform will benefit

MarkRiddoch (approved):

PR #1334, opened 2024-04-05
FOGL-8581 - Plugin cross referencing as per shared keywords

ashish-jabble:


docs/scripts/plugin_and_service_documentation
```diff
@@ -90,6 +92,132 @@ INBUILT_PLUGINS="fledge-north-OMF fledge-rule-Threshold fledge-rule-DataAvailabi
 REPOSITORIES=$(echo ${REPOS} ${INBUILT_PLUGINS} | xargs -n1 | sort -f | xargs)
 echo "REPOSITORIES LIST: "${REPOSITORIES}
 
+# Keywords
```
MarkRiddoch, line 95: I don;t think we need to do it this way. As it is we need to update this script every time we want to introduce a new keyword. We can build the keyword list by looking at al the keyword files and then do the cross referencing once we have the list. Or even build the list as we go along. That way it becomes much easier to maintain.

ashish-jabble: correct, at the moment it's hard to know and prepare list as on fly. Next stage will be with improvements.

ashish-jabble: Fixed

MarkRiddoch (approved):
Looks good to me

PR #1348, opened 2024-05-02
FOGL-8671 Initial implementation for Pipeline type service support at python Core side

ashish-jabble:
As we move further we will support and fix more areas

Note:

Integration testing only once FOGL-8461 is ready!


MarkRiddoch (approved):
Looks good. We will need to make sure we do not do any schema updates om develop until develop3.0 gets merged.

PR #1351, opened 2024-05-06
FOGL-8653: Add API to remove the HttpClient associated with current thread

AmandeepArora:


AmandeepArora:
@MarkRiddoch , @pintomax: I am still waiting for this review.

C/common/storage_client.cpp
```diff
@@ -83,6 +83,27 @@ StorageClient::~StorageClient()
...
+	std::thread::id thread_id = std::this_thread::get_id();
+
+	if(m_client_map.find(thread_id) == m_client_map.end())
+		return false;
+
+	ostringstream ss;
+	ss << thread_id;
+	Logger::getLogger()->debug("Storage client deleting HttpClient object @ %p for thread %s", m_client_map[thread_id], ss.str().c_str());
+	
+	delete m_client_map[thread_id];
```
pintomax, line 101: Maybe worth adding

sto_mtx_client_map.lock();

delete m_client_map[thread_id];

sto_mtx_client_map.unlock();

pintomax (approved):

C/common/storage_client.cpp
AmandeepArora: Good point, thanks.

pintomax (approved):

MarkRiddoch (approved):
Other than Massimiliano's comment I do have anything extra. What is the driver for doing this? It might be worth adding in the doxygen function header why you might use this.

AmandeepArora:
> Other than Massimiliano's comment I do have anything extra. What is the driver for doing this? It might be worth adding in the doxygen function header why you might use this.

Actually notify-north plugin has a singleton StorageClient reference/pointer. In conditional forwarding setting, every time postTrigger readings have to be pulled from storage, a new worker thread is created by notify-north, which then calls StorageClient::readingQueryToReadings(), which further calls getHttpClient and the StorageClient object in question has m_client_map which maps from current thread ID to HttpClient object to allow reuse for same thread. Notify-north always creates a new worker thread and hence HttpClient objects keep accumulating these threads without any possibility of reuse. These are eventually cleaned up in StorageClient destructor (notification service shutdown) but before that, notification service process runs out of FDs and hence specgram images file creation, csv files writing and network connection to storage, UDP connection to syslogd, all start failing.

AmandeepArora:
> > Other than Massimiliano's comment I do have anything extra. What is the driver for doing this? It might be worth adding in the doxygen function header why you might use this.
> 
> @MarkRiddoch : Actually notify-north plugin has a singleton StorageClient reference/pointer. In conditional forwarding setting, every time postTrigger readings have to be pulled from storage, a new worker thread is created by notify-north, which then calls StorageClient::readingQueryToReadings(), which further calls getHttpClient and the StorageClient object in question has m_client_map which maps from current thread ID to HttpClient object to allow reuse for same thread. Notify-north always creates a new worker thread and hence HttpClient objects keep accumulating these threads without any possibility of reuse. These are eventually cleaned up in StorageClient destructor (notification service shutdown) but before that, notification service process runs out of FDs and hence specgram images file creation, csv files writing and network connection to storage, UDP connection to syslogd, all start failing.

So this is a specific scenario, I am not sure what more to add to doxygen comments for deleteHttpClient function.

PR #1355, opened 2024-05-09
FOGL-8533: fix for "null" string value in mapResultSet

pintomax:
FOGL-8533: fix for "null" string value in mapResultSet


praveen-garg:
Will this work with SQLiteLB?

This seems similar to #1300  @gnandan

MarkRiddoch:
> Will this work with SQLiteLB?
> 
> This seems similar to #1300 @gnandan

PR #1300 was not addressing the problem and was completely incorrect. That PR needs to be closed.

PR #1360, opened 2024-05-16
FOGL-5277 Don't regenerate version.h if VERSION has not changed

MarkRiddoch:


mkversion
```diff
@@ -1,5 +1,12 @@
 #!/bin/sh
 output=$1/cmake_build/version.h
+. $1/VERSION
+if [ -f $output ]; then
+	grep -s "VERSION[ 	]*\"${fledge_version}\"" $output > /dev/null
```
ashish-jabble, line 5: we may add `[[:space:]] or [[:blank:]]` instead of `[ 	]*`

ashish-jabble (approved):

PR #1366, opened 2024-05-21
FOGL-8678 Add more example for list and kvlist documentation

MarkRiddoch:


docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -392,6 +392,44 @@ The configuration items within a configuration category can each be defined as o
...
+                "default" : "[ \"speed\", \"temperature\", \"voltage\" ]",
+                "order" : "4",
+                "displayName" : "Labels"
+           }
+
+The returned value for such a configuration item is merely an array of strings.
+
+.. code-block:: JSON
+
+    [ "speed", "temperature", "voltage", "current", "status" ]
```
ashish-jabble, line 415: `[ \"speed\", \"temperature\", \"voltage\" ]`

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -392,6 +392,44 @@ The configuration items within a configuration category can each be defined as o
...
+
+In the above example the user is able to supply any number of unconstrained strings, however if we wanted a user to choose from a set of fixed options they may use an enumeration as the type of the list rather than a string.
```
ashish-jabble, line 417: We do not see much point in a simple list that can only take a subset of values with enumeration. Makes more sense with kvlist type. Though API have a support for a list with enumeration. We may exclude it from doc.

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -410,12 +448,83 @@ A key/value list is a way of storing tagged item pairs within a list. For exampl
...
+A returned list with 4 entries would have a value as shown below.
+
+.. code-block:: JSON
+
+   {
+        "idle" : "\"speed == 0\",
+        "operational" : \"speed == 5000\",
+        "transitional" : \"speed > 0 && speed < 5000\",
+        "overspeed" : \"speed > 5000\"
+   }
```
ashish-jabble, line 460: Not a valid JSON, so a highlighting issue

ashish-jabble: Also for type kvlist and having items string define in above example would return only entry
```
{"idle": "\"speed == 0\""}
```
Either update the example or its result

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -410,12 +448,83 @@ A key/value list is a way of storing tagged item pairs within a list. For exampl
...
+    },
+    {
+        "datapoint" : "current",
+        "register" : "14",
+        "width" : "4",
+        "type" : "float"
+    }
+  ]
+
+An alternative might be to use a key/value pair list where the key is the name of the item to be read, the datapoint in the previous example and the value is an object that describes how to read the data from the PLC.
```
ashish-jabble, line 527: Instead of `key/value pair list` we should refer the actual term `kvlist`

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -392,6 +392,44 @@ The configuration items within a configuration category can each be defined as o
...
+                "type" : "list",
+                "items" : "enumeration",
+                "options" : [ "speed", "current", "temperature", "voltage", "status", "runtime" ],
+                "default" : "{ \"speed\", \"temperature\", \"voltage\" }",
+                "order" : "4",
+                "displayName" : "Labels"
+             }
+
+This differs from an enumeration in that the user is allowed to select more than one option from the set of options.
+
```
ashish-jabble, line 432: We may add others example items like `float and integer` or atleast a note...

ashish-jabble (changes requested):
Minor comments but require them to fix before merge

docs/plugin_developers_guide/02_writing_plugins.rst
MarkRiddoch: This is the returned value, not shown as a string, so no need for the escaping

docs/plugin_developers_guide/02_writing_plugins.rst
MarkRiddoch: Yes, there is a number of times when you want to ave something that is effectively choose n thing from m choices.

docs/plugin_developers_guide/02_writing_plugins.rst
MarkRiddoch: The returned value is when the user has defined 4 entries in the kit, it is not the default.

docs/plugin_developers_guide/02_writing_plugins.rst
ashish-jabble: But `"current", "status"` are extra values as per given JSON example. I believe we should match up the JSON examples with result otherwise user may confuse.

ashish-jabble (approved):
I believe we should match up the results of JSON examples, otherwise user may confuse.

PR #1367, opened 2024-05-21
FOGL-8065: mostrecent parameter fix for sqlite storage engines

pintomax:
FOGL-8065: mostrecent parameter fix for sqlite storage engines

### DIFF
--- a/python/fledge/services/core/api/browser.py
+++ b/python/fledge/services/core/api/browser.py
@@ -218,7 +218,9 @@ async def asset(request):
             _logger.debug("DTS: {} most_recent_ts: {}".format(date_times, most_recent_ts))
             window = int(request.query['seconds'])
             to_ts = most_recent_ts - datetime.timedelta(seconds=window)
-            most_recent_str = most_recent_ts.strftime(dt_format)
+            # Add "0001" to upper limit date string: this allows right query execution
+            # for Sqlite engines, Postgres is fine too with that.
+            most_recent_str = most_recent_ts.strftime(dt_format) + "0001"
             to_str = to_ts.strftime(dt_format)
             _logger.debug("user_ts <={} TO user_ts>{}".format(most_recent_str, to_str))
             _and_where = PayloadBuilder(_where).AND_WHERE(['user_ts', '<=', most_recent_str]).AND_WHERE(


ashish-jabble (changes requested):
We should address/fix the real problem, this is pure hack made on API side.
This is the one of use case for the lost reading given in JIRA, we have more scenarios coming; there is another limitation on storage side to not having support of Between operator which means we have another workaround implemented at API side for most recent functionality to pass datetime UTC string in payload as newer/older KV pair work with datetime.

pintomax:
@ashish-jabble please provide examples of the new payload being passed to storage.

The fix here is to pass some extra data to date string  so SQLite query fetches the upper timestamp.

It there are other scenarios similar to this, then we can then try a different solution for SQLite only

MarkRiddoch:
> We should address/fix the real problem, this is pure hack made on API side. 
Sorry, in what way is this not addressing the real problem?

> This is the one of use case for the lost reading given in JIRA, we have more scenarios coming; there is another limitation on storage side to not having support of Between operator which means we have another workaround implemented at API side for most recent functionality to pass datetime UTC string in payload as newer/older KV pair work with datetime.
Why do we need a between operator to fix this particular problem, there is not set of timestamps in the API to require a between.

ashish-jabble:
> Sorry, in what way is this not addressing the real problem?

In JIRA, I have given the payloads which actually hit the storage readings query endpoint.

Sending user_ts in datetime UTC string with micorseconds, e.g. 2024-05-17 10:01:32.702005

This works with PostgreSQL but not the SQLite. 

Do We know why?

My expectation was to see a fix at database layer for SQLite plugins, not extending the datetime string. Do We store differently or comparison algorithm is broken with conversion mechanism?

Coming to given fix, by adding "0001" to upper limit date string; What is the explanation of this?



> Why do we need a between operator to fix this particular problem, there is not set of timestamps in the API to require a between.

I didnot comment for this problem, But the fact that we had similar situations for between operator and we ended up adding a hack on API side as a workaround, instead of adding proper supports at Storage layer. 


Imagine end users writing their own APIs.

MarkRiddoch:
I am pulling this. This is a very simple, minor issue that we have spent far too long over. Closing the pull request and looking at abandoning the functionality ion the product.

PR #1378, opened 2024-05-29
FOGL-8749 (B) c++ configuration support for listName optional 

ashish-jabble:


ashish-jabble:
@MarkRiddoch I have tried to support at C++ side, seems like an easy task. If you think there is something left then feel free to amend the changes in same branch.

praveen-garg:
We should add docs too for this.

PR #1396, opened 2024-06-10
FOGL-8755 Support bundle and backup restricted with Admin users

ashish-jabble:


MarkRiddoch (approved):
Looks good to me

praveen-garg (commented):
Should We restrict these on GUI as well? Any Jira?

cc @mshariq-nerd

PR #1401, opened 2024-06-19
FOGL-8736 Allow Cache size tuning in core configuration manager

ashish-jabble:
Note: 

If cache disabling is really required then it can take up with another ticket.


python/fledge/common/configuration_manager.py
```diff
@@ -742,6 +739,8 @@ async def _update_value_val(self, category_name, item_name, new_value_val):
                 .FORMAT("return", ("ts", "YYYY-MM-DD HH24:MI:SS.MS")) \
                 .WHERE(["key", "=", category_name]).payload()
             await self._storage.update_tbl("configuration", payload)
+            cat_value = {item_name: {"value": new_value_val}}
+            self._update_special_config_items(category_name, cat_value)
```
praveen-garg, line 743: _set_cache_size() may be a better name or _handle_config_change

python/fledge/common/configuration_manager.py
```diff
@@ -1995,3 +1995,9 @@ def _validate_min_max(_type, val):
                         if not isinstance(eval_s, type_check):
                             raise ValueError(type_mismatched_message)
 
+    def _update_special_config_items(self, cat_name: str, cat_value: dict) -> None:
+        """ Update value in config items for a category which will be required without restarting fledge """
```
praveen-garg, line 1999: which will not/don't require a restart of Fledge
or
which are required without restart of Fledge

Ideally, We don't want a restart anyway.

python/fledge/services/core/server.py
```diff
@@ -295,6 +295,18 @@ class Server:
         }
     }
 
+    _CONFIGURATION_DEFAULT_CONFIG = {
+        'cacheSize': {
+            'description': 'To control the caching size of Core Configuration Manager',
+            'type': 'integer',
+            'displayName': 'Size of Configuration Cache',
```
praveen-garg, line 302: Configuration Manager Cache Size

python/fledge/services/core/server.py
```diff
@@ -604,6 +616,24 @@ async def core_logger_setup(cls):
...
+    @classmethod
+    async def configuration_tune_setup(cls):
+        """ Configuration manager category """
+        try:
+            if cls._configuration_manager is None:
+                cls._configuration_manager = ConfigurationManager(cls._storage_client_async)
+
+            config = cls._CONFIGURATION_DEFAULT_CONFIG
+            category = 'CONFIGURATION'
+            description = "Configuration Manager of Core Server"
```
praveen-garg, line 628: Core Configuration Manager

python/fledge/services/core/server.py
```diff
@@ -604,6 +616,24 @@ async def core_logger_setup(cls):
             _logger.exception(ex)
             raise
 
+    @classmethod
+    async def configuration_tune_setup(cls):
```
praveen-garg, line 620: setup_config_manager_category

python/fledge/services/core/server.py
```diff
@@ -899,6 +930,9 @@ def _start_core(cls, loop=None):
             cls._configuration_manager = ConfigurationManager(cls._storage_client_async)
             cls._interest_registry = InterestRegistry(cls._configuration_manager)
 
+            # Configuration Manager tune setup
```
praveen-garg, line 933: this is not tuning setup

tests/unit/python/fledge/common/test_configuration_cache.py
```diff
@@ -16,7 +16,7 @@ class TestConfigurationCache:
     def test_init(self):
```
praveen-garg, line 16: Add test for trying setup 0, 1,  1000, 1001; can be unit tests too.

praveen-garg (commented):
LGTM! 

Some text fixes suggestion.

Let's run some system test  RPi, push/poll centralised management and upgrade.

MarkRiddoch (dismissed):

python/fledge/common/configuration_manager.py
praveen-garg: Using such name "special" is highly discouraged. What is special?

tests/unit/python/fledge/common/test_configuration_cache.py
```diff
@@ -13,10 +13,13 @@
 @pytest.allure.story("common", "configuration_manager", "configuration_cache")
 class TestConfigurationCache:
 
-    def test_init(self):
-        cached_manager = ConfigurationCache()
+    @pytest.mark.parametrize("size", [
+        1, 10, 20, 1000
```
praveen-garg, line 17: Add unit test for 0, 1001 to ensure not only min/max API prevents them but functions are safe to that

praveen-garg:
Setting to 1 is almost equivalent to no/disabled cache. That part is okay.

tests/unit/python/fledge/common/test_configuration_cache.py
ashish-jabble: These tests are only meant for Configuration Cache and no access of min and max attributes that applies to configuration manager calls

python/fledge/common/configuration_manager.py
ashish-jabble: Special - Unlike value updates it has more things to do; that's why we have another def to tackle such things separately but NOT in the actual Update calls.

praveen-garg: Avoid using special like word in def name

tests/unit/python/fledge/common/test_configuration_cache.py
praveen-garg: then at-least 0 test should be there;

And 0,1,1000,1001 should be around API

python/fledge/services/core/server.py
```diff
@@ -604,6 +616,31 @@ async def core_logger_setup(cls):
...
+            category = 'CONFIGURATION'
+            description = "Core Configuration Manager"
+            await cls._configuration_manager.create_category(category, config, description, True,
+                                                             display_name='Configuration Manager')
+            config = await cls._configuration_manager.get_category_all_items(category)
+            cache_size = int(config['cacheSize']['value'])
+            # Internal handling of cache size when with 0 and setting to default cache size of 30;
+            if cache_size == 0:
+                default_cache_size = cls._configuration_manager._cacheManager.max_cache_size
+                _logger.warn("Configuration Manager Cache Size is setting to default size of {}".format(
```
praveen-garg, line 636: Configuration Manager Cache Size is being set to the default size {}.
Setting Configuration Manager Cache Size to to the default size {}

praveen-garg (approved):

PR #1403, opened 2024-06-20
FOGL-8842 API is protected from Shell injection calls

ashish-jabble:


MarkRiddoch (commented):
A don't see how this addresses the original issue of inserting commands into the Python package install command?

ashish-jabble:
> A don't see how this addresses the original issue of inserting commands into the Python package install command?

`curl -X POST http://localhost:8081/fledge/python/package -d '{"package":"aj", "version": "2.2.5; touch /tmp/aj.txt; ls"}'`

Previously, it was created a file and list of directory contents. But with the new change  I don't see it executes the shell command anymore.

MarkRiddoch (approved):

praveen-garg (approved):

PR #1411, opened 2024-07-03
FOGL-8760 permissions optional support added in configuration Manager along with restrictions on configuration update API

ashish-jabble:


python/fledge/common/configuration_manager.py
```diff
@@ -266,7 +266,7 @@ async def _validate_category_val(self, category_name, category_val, set_value_va
 
             optional_item_entries = {'readonly': 0, 'order': 0, 'length': 0, 'maximum': 0, 'minimum': 0,
                                      'deprecated': 0, 'displayName': 0, 'rule': 0, 'validity': 0, 'mandatory': 0,
-                                     'group': 0, 'listSize': 0, 'listName': 0}
+                                     'group': 0, 'listSize': 0, 'listName': 0, 'permission': 0}
```
praveen-garg, line 269: Why 0 not []. 

Key better as permissions?

python/fledge/common/configuration_manager.py
```diff
@@ -301,6 +301,19 @@ def get_entry_val(k):
                         else:
                             d = {entry_name: entry_val}
                             expected_item_entries.update(d)
+                    elif entry_name == "permission":
+                        if not isinstance(entry_val, list):
+                            raise ValueError(
+                                'For {} category, {} entry value must be in list for item name {}; got {}.'
```
praveen-garg, line 307: in list -> a list of user roles

python/fledge/common/configuration_manager.py
```diff
@@ -301,6 +301,19 @@ def get_entry_val(k):
...
+                                ''.format(category_name, entry_name, item_name, type(entry_val)))
+                        if not entry_val:
+                            raise ValueError(
+                                'For {} category, {} entry value must not be an empty for item name '
```
praveen-garg, line 311: an?

python/fledge/common/configuration_manager.py
```diff
@@ -348,6 +361,20 @@ def get_entry_val(k):
                             raise ValueError('For {} category, listName cannot be empty for item name '
                                              '{}'.format(category_name, item_name))
                         item_val['listName'] = list_name
+                    elif "permission" in item_val:
+                        permission = item_val['permission']
+                        if not isinstance(permission, list):
+                            raise ValueError(
+                                'For {} category, permission entry value must be in list for item name {}; got {}.'
```
praveen-garg, line 368: as above

python/fledge/common/configuration_manager.py
```diff
@@ -348,6 +361,20 @@ def get_entry_val(k):
...
+                                ''.format(category_name, item_name, type(permission)))
+                        if not permission:
+                            raise ValueError(
+                                'For {} category, permission entry value must not be an empty for item name {}.'.format(
+                                    category_name, item_name))
+                        else:
+                            if not all(isinstance(ev, str) and ev != '' for ev in permission):
+                                raise ValueError('For {} category, permission entry values must be in string and '
```
praveen-garg, line 376: For {} category, permissions value should be a list of user roles

in string is not clear

python/fledge/common/configuration_manager.py
```diff
@@ -487,6 +514,19 @@ def get_entry_val(k):
                     if entry_name in ('properties', 'options'):
                         d = {entry_name: entry_val}
                         expected_item_entries.update(d)
+                elif entry_name == "permission":
+                    if not isinstance(entry_val, list):
+                        raise ValueError(
+                            'For {} category, {} entry value must be in list for item name {}; got {}.'
```
praveen-garg, line 520: a list 
a list of user roles

python/fledge/common/configuration_manager.py
```diff
@@ -510,6 +550,19 @@ def get_entry_val(k):
                             self._validate_type_value('float', entry_val)) is False:
                             raise ValueError('For {} category, entry value must be an integer or float for item name '
                                              '{}; got {}'.format(category_name, entry_name, type(entry_val)))
+                    elif entry_name == "permission":
+                        if not isinstance(entry_val, list):
+                            raise ValueError(
+                                'For {} category, {} entry value must be in list for item name {}; got {}.'
```
praveen-garg, line 556: As above

python/fledge/common/configuration_manager.py
```diff
@@ -510,6 +550,19 @@ def get_entry_val(k):
...
+                                ''.format(category_name, entry_name, item_name, type(entry_val)))
+                        if not entry_val:
+                            raise ValueError(
+                                'For {} category, {} entry value must not be an empty for item name '
+                                '{}.'.format(category_name, entry_name, item_name))
```
praveen-garg, line 561: as above

python/fledge/common/configuration_manager.py
```diff
@@ -510,6 +550,19 @@ def get_entry_val(k):
...
+                        else:
+                            if not all(isinstance(ev, str) and ev != '' for ev in entry_val):
+                                raise ValueError('For {} category, {} entry values must be in string and non-empty '
```
praveen-garg, line 564: as above

python/fledge/common/configuration_manager.py
```diff
@@ -751,26 +804,44 @@ async def _update_value_val(self, category_name, item_name, new_value_val):
...
             None
         """
-
         try:
             payload = {"updates": []}
             audit_details = {'category': category_name, 'items': {}}
             cat_info = await self.get_category_all_items(category_name)
             if cat_info is None:
                 raise NameError("No such Category found for {}".format(category_name))
+            """ Note: Update reject to the properties with permission property when the logged in user type is not
```
praveen-garg, line 824: user type  -> (requesting) user role

python/fledge/common/configuration_manager.py
```diff
@@ -751,26 +804,44 @@ async def _update_value_val(self, category_name, item_name, new_value_val):
...
+            if request is not None:
+                if hasattr(request, "user_is_admin"):
+                    if not request.user_is_admin:
+                        from fledge.services.core.user_model import User
+                        roles = await User.Objects.get_roles()
+                        user_role_name = [r['name'] for r in roles if request.user['role_id'] == r['id']]
+                        if user_role_name:
+                            user_role_name = user_role_name[0]
+                        else:
+                            raise ValueError("No role found for logged in user.")
```
praveen-garg, line 835: Should be better text?

Also, It is not possible to log in for any such user.

 To handle any worst case, Requesting user's role is not matched with any existing roles.

python/fledge/common/configuration_manager.py
```diff
@@ -751,26 +804,44 @@ async def _update_value_val(self, category_name, item_name, new_value_val):
...
                 raise NameError("No such Category found for {}".format(category_name))
+            """ Note: Update reject to the properties with permission property when the logged in user type is not
+             given in the list of permission. """
+            if request is not None:
+                if hasattr(request, "user_is_admin"):
+                    if not request.user_is_admin:
+                        from fledge.services.core.user_model import User
+                        roles = await User.Objects.get_roles()
+                        user_role_name = [r['name'] for r in roles if request.user['role_id'] == r['id']]
+                        if user_role_name:
```
praveen-garg, line 832: Avoid assigning array back to string; May be have a temp variable. Type hints will be better.

python/fledge/common/configuration_manager.py
```diff
@@ -1075,6 +1149,21 @@ async def set_category_item_value_entry(self, category_name, item_name, new_valu
...
+                            user_role_name = user_role_name[0]
+                            if 'permission' in storage_value_entry:
+                                if not (user_role_name in storage_value_entry['permission']):
+                                    raise Exception('Forbidden')
+                        else:
+                            raise ValueError("No role found for logged in user.")
```
praveen-garg, line 1166: As above

python/fledge/services/core/api/configuration.py
```diff
@@ -293,14 +278,20 @@ async def set_configuration_item(request):
...
             await cf_mgr.set_optional_value_entry(category_name, config_item, list(found_optional.keys())[0], list(found_optional.values())[0])
     except ValueError as ex:
         raise web.HTTPNotFound(reason=ex) if not found_optional else web.HTTPBadRequest(reason=ex)
     except (TypeError, KeyError) as ex:
         raise web.HTTPBadRequest(reason=ex)
-
+    except Exception as ex:
+        msg = str(ex)
+        if 'Forbidden' in msg:
+            msg = "Insufficient access privileges to change the value for category."
```
praveen-garg, line 292: This is for category level only or individual config items as well?

praveen-garg: Also mention which category/ configuration item <?>.

python/fledge/common/configuration_manager.py
ashish-jabble: Possibility of this case will never execute

python/fledge/common/configuration_manager.py
ashish-jabble: Here we are checking instance type only

python/fledge/common/configuration_manager.py
ashish-jabble: user type and role is same thing

python/fledge/services/core/api/configuration.py
ashish-jabble: only on config item

python/fledge/common/configuration_manager.py
praveen-garg: in is not correct, `a list` even better `a list of string`

python/fledge/common/configuration_manager.py
praveen-garg: Move this to check_permissions def and also add & call User.Objects.get_role_name_by_id with LIMIT 1.

Then you don't need it here but in next loop you can simply have

```
 if 'permission' in cat_info[item_name]:
      if not check_permissions:
           raise <custom exception>
```

This will also add reusability in bulk update method.

python/fledge/common/configuration_manager.py
```diff
@@ -751,26 +804,44 @@ async def _update_value_val(self, category_name, item_name, new_value_val):
...
             for item_name, new_val in config_item_list.items():
                 if item_name not in cat_info:
                     raise KeyError('{} config item not found'.format(item_name))
+                if request is not None:
+                    if hasattr(request, "user_is_admin"):
+                        if not request.user_is_admin:
+                            if 'permission' in cat_info[item_name]:
+                                if not (user_role_name in cat_info[item_name]['permission']):
+                                    raise Exception('Forbidden')
```
praveen-garg, line 844: Better to add a know exception custom class, AccessDenied or PermissionDenied and handle that instead of string match

ashish-jabble: yeah it was planned but this exception actually to be defined in user model but Not in configuration manager; due to avoid of circular dependency issue, I have kept it under general exception having message Forbidden to get the actual HTTP status code.

PR #1413, opened 2024-07-08
FOGL-8840 Add C++ support for permissions property of configuration items

MarkRiddoch:


pintomax (approved):

C/common/config_category.cpp
```diff
@@ -872,6 +872,50 @@ vector<string> ConfigCategory::getOptions(const string& name) const
...
+		}
+	}
+	throw new ConfigItemNotFound();
+}
+
+/**
+ * Return true if the user has permission to update the named item
+ *
+ * @param name	The name of the configuration item to return
+ * @param username	The name of the user to test
```
ashish-jabble, line 898: Param should be `rolename` - The name of role user

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -636,6 +636,8 @@ Properties
      - The current value of the configuration item. This is not included when defining a set of default configuration in, for example, a plugin.
    * - properties
      - A set of items that are used in list and kvlist type items to create a list of groups of configuration items.
+   * - permissions
+     - An array of user roles that are allowed to update this configuration item. If not given then the configuration item can be updated by any user.
```
ashish-jabble, line 640: and array cannot be empty

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -724,3 +726,47 @@ There is also a convenience function that can be used if you not want to define
...
+.. code-block:: JSON
+
+        'logLevel': {
+            'description': 'Minimum logging level reported for Core server',
+            'type': 'enumeration',
+            'displayName': 'Minimum Log Level',
+            'options': ['debug', 'info', 'warning', 'error', 'critical'],
+            'default': 'warning',
+            'order': '1',
+            'permissions': ['admin', 'edit']
```
ashish-jabble, line 765: There is no 'edit' role name exists as such. You may add 'user' here. As we have only inbuilt roles so with 'edit' may confuse here to user.

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -724,3 +726,47 @@ There is also a convenience function that can be used if you not want to define
...
+        }
+
+In this case users with the role admin or edit are allowed to alter the configuration item.
```
ashish-jabble, line 768: same as above

C/common/config_category.cpp
```diff
@@ -872,6 +872,50 @@ vector<string> ConfigCategory::getOptions(const string& name) const
...
+ * @return bool	True if the named user can update the configuration item
+ * @throws exception if the item does not exist in the category
+ */
+bool ConfigCategory::hasPermission(const std::string& name, const std::string& username) const
```
ashish-jabble, line 902: For future use?

C/common/config_category.cpp
```diff
@@ -872,6 +872,50 @@ vector<string> ConfigCategory::getOptions(const string& name) const
...
+	for (unsigned int i = 0; i < m_items.size(); i++)
+	{
+		if (name.compare(m_items[i]->m_name) == 0)
+		{
+			return m_items[i]->m_permissions;
+		}
+	}
+	throw new ConfigItemNotFound();
+}
+
```
ashish-jabble, line 893: As you have added getter for this, Don't we require setter for the same item?

docs/plugin_developers_guide/02_writing_plugins.rst
```diff
@@ -724,3 +726,47 @@ There is also a convenience function that can be used if you not want to define
...
+      "admin"
+    ],
+    "value": "optional"
+  }
+
+The use of the permissions property with the single role of admin means that only admin users can change this setting. If we wished to allow more than admin users we can add another role. For example
+
+.. code-block:: JSON
+
+        'logLevel': {
```
ashish-jabble, line 758: Use Double quotes in dict for the JSON highlighting.

ashish-jabble (changes requested):
With the Python server changes, configuration API integration tests are failing.
We should also think of permissions KV Pair for a category, as of now only for config item. For example: For REST API category, permissions KV pair added for each config item.

C/common/config_category.cpp
MarkRiddoch: Yes, nothing is using this currently

C/common/config_category.cpp
MarkRiddoch: Nothing is using the currently, setting is an issue with knowing how has permissions. It is a construction time only thing to set it currently. We can tackle that if and when we need it.

pintomax (approved):

C/common/config_category.cpp
ashish-jabble: Okay, I thought it was required in `setItemAttribute`. Seems like only required for an optional attribute in categoryItem constructor and its JSON representation function i.e `toJSON` and `defaultToJSON`

ashish-jabble (approved):

PR #1415, opened 2024-07-09
FOGL-8538A: New file and directory utilities

RVerhoeff:
New file and directory routines are added in new files:

- createDirectory : creates a single directory (not a directory tree)
- removeDirectory : removes a directory (with all files and subdirectories)
- copyFile : copies a file from a source to a destination


MarkRiddoch (approved):

ashish-jabble:
fyi, We have almost same utility which exists at https://github.com/fledge-iot/fledge/blob/develop/C/plugins/utils/cmdutil.cpp and  to run the supported commands for fledge as root using setuid.
As of now, only consumer is python.
So not sure if we can merge into one utility or keep individual utilities for different works.

Is this required for any specific plugin?

RVerhoeff:
@ashish-jabble , thanks for your feedback. I looked at the code for _cmdutil_ and tried a few things. I have some comments:

- Environment variable _FLEDGE_ROOT_ must be defined. If not, it will not assume the default of _/usr/local/fledge_ .
- _cmdutil_ will not run unless Fledge is running in _/usr/local/fledge_. If running from a development directory with _FLEDGE_ROOT_ defined, subroutine '_checkfile_' fails to find any subdirectories.
- For _cp_, the destination file can only be in the _$FLEDGE_ROOT_ directory. In my case, that's OK.
- For _cp_, destination file is owned by root not the user. The account running Fledge may not have root permissions. Your comment says that _setuid_ can be turned on to run commands as root. How do you turn it off?
- As I understand it, calling _cmdutil_ from C++ would require _system()_ or _execvp()_ to run which requires spawning a process. _Cmdutil_ calls _execvp()_ itself so this would be a spawned process spawning another process.
 
For C++ programs, I think we would benefit from utilities created with the C run-time library so that creating command lines and spawning processes is not necessary. This Pull Request came out of my work on updating the S2OPCUA South plugin. Creating directories and copying files happens many times during startup but not while the plugin is actively processing data.

By the way, one of the error messages has a typo: insallation -> installation.

ashish-jabble:
Thanks for the detailed info.

> Environment variable FLEDGE_ROOT must be defined.

yeah this utility is only meant and worked inside FLEDGE_ROOT; as per initially requirements.

I just only want to share the existing stuff and this may only fit/worked with the specific scenarios moreover inside FLEDGE_ROOT. 

If that does not meet your requirement, you can go ahead with the merge.

MarkRiddoch (approved):

PR #1422, opened 2024-07-16
FOGL-8913: Readings count double on restart.

cyberwalk3r:


C/services/north/data_load.cpp
```diff
@@ -652,6 +646,7 @@ void DataLoad::configChange(const string& category, const string& newConfig)
 		if (config.itemExists("source"))
 		{
 			setDataSource(config.getValue("source"));
+			m_lastFetched = 0; // Reset last fetch counter on source change
```
MarkRiddoch, line 649: Isn't this still going to reset the last fetched id when we have a configuration change? Surely we only wan to do this when the actual value of the source item changes?

cyberwalk3r: Yes you are correct, I have updated code to reset last fetched id only if source changes.

MarkRiddoch (approved):

ashish-jabble:
@cyberwalk3r rebase with develop branch, so that CI checker will pass

cyberwalk3r:
Closing this branch in favor of https://github.com/fledge-iot/fledge/pull/1430.

PR #1431, opened 2024-07-26
FOGL-8947: Last send Id not updated for single send cycle.

cyberwalk3r:
- Updated last sent id if no readings to send.

### DIFF
--- a/C/services/north/data_load.cpp
+++ b/C/services/north/data_load.cpp
@@ -176,6 +176,7 @@ void DataLoad::readBlock(unsigned int blockSize)
 {
 	int n_waits = 0;
 	int n_update_streamId = 0;
+	int max_wait_count = 5;	// Maximum wait counter to update streams table
 	unsigned int waitPeriod = INITIAL_BLOCK_WAIT;
 	do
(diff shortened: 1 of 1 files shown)


MarkRiddoch (approved):

praveen-garg:
@gnandan Could you review by test, Please?

C/services/north/data_load.cpp
```diff
@@ -226,6 +229,13 @@ void DataLoad::readBlock(unsigned int blockSize)
 		{
 			// Delete the empty readings set
 			delete readings;
+			n_update_streamId++;
+			if (n_update_streamId > 5) {
```
gnandan, line 233: Why if condition is checking for 5, is it some random number?

cyberwalk3r: This is maximum wait counter when no reading are sent to update streams table. Removed constant.

gnandan (approved):
Reviewed by test. Stream table is being updated and data is being displayed on GUI as expected.

MarkRiddoch (approved):

PR #1432, opened 2024-07-29
FOGL-8507 : Taken care of quotes and slash in asset name

gnandan:


MarkRiddoch (commented):
What tests have been run against this branch?

gnandan:
> What tests have been run against this branch?

I executed system tests, lab scripts using CI Jobs and did some manual testing. Please refer following sheet for detailed report https://docs.google.com/spreadsheets/d/1brHU_U5kCGqKkUxm5igvvu_SPYjmtuDHRDpIQxy3ZE8/edit?usp=sharing

C/common/string_utils.cpp
```diff
@@ -462,3 +462,71 @@ std::string StringAround(const std::string& str, unsigned int pos,
...
+
+/**
+ * Replicate a character/substring within an string
+ *
+ * @param out StringToManage    string in which apply the search and replicate
+ * @param     StringToSearch    string to search and replicate
+ * @param     StringReplicate  substitution string
+ *
+ */
+void replicate(std::string& StringToManage,
```
MarkRiddoch, line 474: Replicate is a bad name for this. Probably replaceAll us better

MarkRiddoch (commented):
We have a proliferation of escape routines, it woudl be good to consolidate them or at least make use of one already written.

C/common/string_utils.cpp
gnandan: Here function is replicating a single character to double character. Also **StringReplaceAll** function is already there but will not work in this scenario. That's why I created new function. I will rename the function as **StringReplaceAllEx** to make it more intuitive.

MarkRiddoch (approved):

PR #1440, opened 2024-08-07
FOGL-8285: Added two new columns in steams table to track last sent stats and audit

gnandan:


gnandan:
Alternate approach is used at https://github.com/fledge-iot/fledge/pull/1455

PR #1446, opened 2024-08-15
FOGL-9027 Resolve issue with Python filters sending data during

MarkRiddoch:
shutdown.

### DIFF
--- a/C/common/filter_pipeline.cpp
+++ b/C/common/filter_pipeline.cpp
@@ -312,8 +312,16 @@ bool FilterPipeline::setupFiltersPipeline(void *passToOnwardFilter, void *useFil
 void FilterPipeline::cleanupFilters(const string& categoryName)
 {
 
-	// Shutdown filters - do this down the pipeline to allow
-	// filters to pass data on to the next ask they shutdown
+	// Shutdown filters - do this down the pipeline, starting
+	// from the first filter in the pipeline This allows a filter
+	// to asynchnrously send data in the shutdown call to the
+	// next element in the pipeline sine that next element has
+	// not yet been asked to shutdown.
+	//
+	// This is not behaviour that is encouraged or designed, but a
+	// small number of Python filters have implemented sending data
+	// during shutdown, hence the need to ensure that data has
+	// somewhere to go.
 	for (auto it = m_filters.begin(); it != m_filters.end(); ++it)
 	{
 		PipelineElement *element = *it;
--- a/C/common/filter_pipeline.cpp
+++ b/C/common/filter_pipeline.cpp
@@ -311,13 +311,19 @@ bool FilterPipeline::setupFiltersPipeline(void *passToOnwardFilter, void *useFil
  */
 void FilterPipeline::cleanupFilters(const string& categoryName)
 {
-	// Cleanup filters, in reverse order
-	for (auto it = m_filters.rbegin(); it != m_filters.rend(); ++it)
+
+	// Shutdown filters - do this down the pipeline to allow
+	// filters to pass data on to the next ask they shutdown
+	for (auto it = m_filters.begin(); it != m_filters.end(); ++it)
 	{
 		PipelineElement *element = *it;
 		ConfigHandler *configHandler = ConfigHandler::getInstance(mgtClient);
 		element->shutdown(m_serviceHandler, configHandler);
-
+	}
+	// Delete filters, in reverse order
+	for (auto it = m_filters.rbegin(); it != m_filters.rend(); ++it)
+	{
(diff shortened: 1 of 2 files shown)


C/services/south/ingest.cpp
```diff
@@ -1056,6 +1056,12 @@ void Ingest::useFilteredData(OUTPUT_HANDLE *outHandle,
 	lock_guard<mutex> guard(ingest->m_useDataMutex);
 	
 	vector<Reading *> *newData = readingSet->getAllReadingsPtr();
+	if (!ingest->m_data)
+	{
+		// If we are called during shutdown there will be no m_data in place
+		// and we create a new one to handle this special case.
+		ingest->m_data = new vector<Reading *>;
```
RVerhoeff, line 1063: If a new vector is allocated here, when is it deleted? I can see that this might be tricky because _m_data_ is sometimes allocated directly and at other times it's a copy of _m_queue_. If the _m_data_ allocated here is not deleted, it's a minor error because the plugin is shutting down anyway.

C/common/filter_pipeline.cpp
```diff
@@ -311,13 +311,19 @@ bool FilterPipeline::setupFiltersPipeline(void *passToOnwardFilter, void *useFil
  */
 void FilterPipeline::cleanupFilters(const string& categoryName)
 {
-	// Cleanup filters, in reverse order
-	for (auto it = m_filters.rbegin(); it != m_filters.rend(); ++it)
+
+	// Shutdown filters - do this down the pipeline to allow
+	// filters to pass data on to the next ask they shutdown
```
RVerhoeff, line 316: Could you clarify this comment?

RVerhoeff (changes requested):
Comment about deleting the vector allocated for _m_data_.

C/services/south/ingest.cpp
MarkRiddoch: It isn't, but this only occurs during a shutdown, so the process is about to exit and the memory will be cleaned then

C/common/filter_pipeline.cpp
```diff
@@ -312,8 +312,16 @@ bool FilterPipeline::setupFiltersPipeline(void *passToOnwardFilter, void *useFil
 void FilterPipeline::cleanupFilters(const string& categoryName)
 {
 
-	// Shutdown filters - do this down the pipeline to allow
-	// filters to pass data on to the next ask they shutdown
+	// Shutdown filters - do this down the pipeline, starting
+	// from the first filter in the pipeline This allows a filter
```
RVerhoeff, line 316: first filter in the pipeline This allows... -> first filter in the pipeline. This allows... (add a period after "pipeline").

C/common/filter_pipeline.cpp
```diff
@@ -312,8 +312,16 @@ bool FilterPipeline::setupFiltersPipeline(void *passToOnwardFilter, void *useFil
...
+	// to asynchnrously send data in the shutdown call to the
```
RVerhoeff, line 317: Typo: asynchnrously -> asynchronously

RVerhoeff (changes requested):
Typos. Otherwise excellent.

RVerhoeff (approved):

C/services/south/ingest.cpp
```diff
@@ -1056,6 +1056,15 @@ void Ingest::useFilteredData(OUTPUT_HANDLE *outHandle,
 	lock_guard<mutex> guard(ingest->m_useDataMutex);
 	
 	vector<Reading *> *newData = readingSet->getAllReadingsPtr();
+	if (!ingest->m_data)
+	{
+		// If we are called during shutdown there will be no m_data in place
+		// and we create a new one to handle this special case. In this case
+		// the m_data will not be explicitly deleted. However as we are shutting
+		// down this will note cause a problem as all memory is recovered at process
```
AmandeepArora, line 1064: "note" -> "not"

Is it possible that pipeline shutdown happens without the service going down? E.g. removal of all filters. It is still an corner case, and we may leave it unhandled if it is not straightforward to handle this.

AmandeepArora (approved):

C/services/south/ingest.cpp
MarkRiddoch: Yes, it is possible. I will need to think about that one. My inclination is to checkin what we have, as it is not going to result in a recurring memory leak, and then resolve this afterwards. That way we can at least progress with this and the FOGL-8944 that is probably a bigger issue than this.

AmandeepArora: Thanks. I agree.

PR #1447, opened 2024-08-16
FOGL-9004 Python metadata fixes in setup file

ashish-jabble:


python/setup.py
```diff
@@ -2,11 +2,11 @@
 
 setup(
     name='Fledge',
-    python_requires='~=3.5',
-    version='0.1',
-    description='Fledge',
-    url='http://github.com/fledge/Fledge',
-    author='OSIsoft, LLC',
+    python_requires='>=3.6.9',
+    version='2.5.0',
```
praveen-garg, line 6: Another case of version bump?

python/setup.py
```diff
@@ -2,11 +2,11 @@
...
+    description='Fledge, the open source platform for the Internet of Things',
+    url='https://github.com/fledge-iot/fledge',
+    author='Dianomic Systems Inc.',
```
praveen-garg, line 9: I'm not really sure, if We should replace or append?

Dianomic Systems, Inc.

python/setup.py
ashish-jabble: yupe for consistency across

python/setup.py
ashish-jabble: @MarkRiddoch Please suggest.

MarkRiddoch: We should probably append

ashish-jabble: Also please confirm which one we should use for the following:
a) Dianomic Systems
b) Dianomic Systems Inc.
c) Dianomic Systems, Inc.

praveen-garg (approved):

MarkRiddoch (approved):

PR #1449, opened 2024-08-16
FOGL-9032 Add basic disk usage monitoring for the disks holding storage

MarkRiddoch:
data

Currently data is only written to the log. Once we work out a way to talk to the management client from within the storage plugin we should add at least some of these warnings to the alert mechanism.


C/plugins/storage/common/disk_monitor.cpp
```diff
@@ -0,0 +1,159 @@
...
+ */
+#include <disk_monitor.h>
+#include <logger.h>
+#include <sys/vfs.h>
+#include <string.h>
+
+using namespace std;
+
+/**
+ * Construt a disk space monitor class
```
AmandeepArora, line 18: Construct

C/plugins/storage/common/disk_monitor.cpp
```diff
@@ -0,0 +1,159 @@
+/*
+ * Fledge storage service.
```
AmandeepArora, line 2: Need to update this?

C/plugins/storage/common/disk_monitor.cpp
```diff
@@ -0,0 +1,159 @@
...
+/**
+ * Construt a disk space monitor class
+ *
+ * It monitors the free space on two path, since the storage service may use
+ * different locations for reading and configuration storage. If they are both
+ * the same file system then the monitor is only done once.
+ *
+ * If the free space falls below 5% a fatal error is written to the error log
+ * If it falls below 10% a warning is written advising that disk space should be released
+ * It attempts to predict when storage will bebeome exhausted. If it is less then 14 days
```
AmandeepArora, line 26: "become"?

C/plugins/storage/common/disk_monitor.cpp
```diff
@@ -0,0 +1,159 @@
...
+using namespace std;
+
+/**
+ * Construt a disk space monitor class
+ *
+ * It monitors the free space on two path, since the storage service may use
+ * different locations for reading and configuration storage. If they are both
+ * the same file system then the monitor is only done once.
+ *
+ * If the free space falls below 5% a fatal error is written to the error log
```
AmandeepArora, line 24: 5% and 10% may be too late for the cases where disk is getting filled too fast e.g. with vibration data? Also some reasonable disk space may need to be kept aside for system related files to avoid corrupting the system files.

C/plugins/storage/common/disk_monitor.cpp
```diff
@@ -0,0 +1,159 @@
...
+ * If it falls below 10% a warning is written advising that disk space should be released
+ * It attempts to predict when storage will bebeome exhausted. If it is less then 14 days
+ * it will report this to the error log once a day.
+ * If it is less than 72 hours it will report this once per hour.
```
AmandeepArora, line 28: These are configurable?

C/plugins/storage/common/disk_monitor.cpp
```diff
@@ -0,0 +1,159 @@
...
+		{
+			m_sameDevice = true;
+		}
+		m_started = true;
+	}
+	m_lastCheck += interval;
+
+	if (m_lastCheck < CHECK_THRESHOLD)
+	{
+		// Do not check too frerquenctly
```
AmandeepArora, line 70: "frequently"

C/plugins/storage/common/disk_monitor.cpp
```diff
@@ -0,0 +1,159 @@
...
+		{
+			m_logger->fatal("Disk space is critically low. Urgent action required, continuing may result in data corruption");
+		}
+		else if (perc < 10.0)
+		{
+			m_logger->error("Available disk space is becoming low, please consider releasing more disk space");
+		}
+		if (m_lastPerc1 > 0.0)
+		{
+			double diff = m_lastPerc1 - perc;
```
AmandeepArora, line 93: diff over a very small duration like 15 seconds may not be representative of speed of disk filling. Wait a bit longer?

C/plugins/storage/common/disk_monitor.cpp
```diff
@@ -0,0 +1,159 @@
...
+				double prediction = (perc * CHECK_THRESHOLD)/ (3600.0 * diff);
+				if (prediction <= 72.0 && m_lastPrediction1 - prediction > 1.0)
+				{
+					m_logger->error("At current rates disk space will be exhausted in %.0f hours", prediction);
+					m_lastPrediction1 = prediction;
+				}
+				else if (prediction / 24.0 <= 14 && (m_lastPrediction1 == 0.0 || m_lastPrediction1 - prediction > 24.0))
+				{
+					m_lastPrediction1 = prediction;
+					m_logger->warn("At current rates disk space will be exhausted in %.1f days", prediction / 24);
```
AmandeepArora, line 105: floor(prediction/24) or rounding will be better?

C/plugins/storage/common/disk_monitor.cpp
```diff
@@ -0,0 +1,159 @@
...
+		}
+		if (m_lastPerc2 > 0.0)
+		{
+			double diff = m_lastPerc2 - perc;
+			if (diff > 0.0)
+			{
+				double prediction = (perc * CHECK_THRESHOLD)/ (3600.0 * diff);
+				if (prediction <= 72.0 && (m_lastPrediction2 == 0.0 || m_lastPrediction2 - prediction > 1.0))
+				{
+					m_logger->error("At current rates disk space will be exhausted in %.0f hours", prediction);
```
AmandeepArora, line 143: "rate"?

AmandeepArora (dismissed):
Added some minor comments, looks good otherwise.

C/plugins/storage/common/disk_monitor.cpp
MarkRiddoch: Hence the prediction stuff. Although I would say if you are ingesting that sort of data then you should have a biggish storage system anyway. In which case 10% is a lot of storage.

C/plugins/storage/common/disk_monitor.cpp
MarkRiddoch: Not possible unfortunately

C/plugins/storage/common/disk_monitor.cpp
MarkRiddoch: It would either be "At the current rate" or "At current rates". Since it could be two storage plugins consuming disk space, and maybe other files such as buckets, using rates rather than rate is probably more correct.

C/plugins/storage/common/disk_monitor.cpp
```diff
@@ -15,19 +15,23 @@
...
- * It attempts to predict when storage will bebeome exhausted. If it is less then 14 days
+ * It attempts to predict when storage will become exhausted. If it is less then 14 days
  * it will report this to the error log once a day.
  * If it is less than 72 hours it will report this once per hour.
  *
  * All reporting is to the system log
+ *
+ * NB In an ideal world we would make the thresholds and reporting interval configurable,
+ * however we are running within the limited environment of a storage plugin and do not
+ * have access to the manaagement client or configuration subsystem.
```
AmandeepArora, line 34: "manaagement"

C/plugins/storage/common/disk_monitor.cpp
```diff
@@ -67,7 +71,7 @@ void DiskSpaceMonitor::periodic(int interval)
 
 	if (m_lastCheck < CHECK_THRESHOLD)
 	{
-		// Do not check too frerquenctly
+		// Do not check too frerquently
```
AmandeepArora, line 74: "frerquently"

PR #1452, opened 2024-08-22
FOGL-8951 Response header fixes in backup API and also fixed empty timestamp issue for SQLite 

ashish-jabble:


mshariq-nerd:
Download is working fine on this PR for a uploaded backup.
For timestamps we already have couple of issues reported. Like FOGL-7683 & FOGL-7693. So not sure if those are also fixed by the timestamp change? Also Verified with Postgres?

mshariq-nerd (approved):

PR #1463, opened 2024-09-12
FOGL-9061: modified asset to have html encoded value

gnandan:


python/fledge/common/storage_client/storage_client.py
```diff
@@ -22,6 +22,7 @@
 from fledge.common.service_record import ServiceRecord
 from fledge.common.storage_client.exceptions import *
 from fledge.common.storage_client.utils import Utils
+import urllib.parse
```
ashish-jabble, line 25: Wrong import order. Standard import should be placed before aiohttp

python/fledge/common/storage_client/storage_client.py
```diff
@@ -581,7 +582,7 @@ async def purge(self, age=None, sent_id=0, size=None, flag=None, asset=None):
         if flag:
             put_url += "&flags={}".format(flag.lower())
         if asset is not None:
-            put_url = '/storage/reading/purge?asset={}'.format(asset)
+            put_url = '/storage/reading/purge?asset={}'.format(urllib.parse.quote(asset))
```
ashish-jabble, line 585: We better to add unit tests for the asset case

ashish-jabble (approved):

MarkRiddoch (approved):

PR #1469, opened 2024-09-30
FOGL-9163 API statistics schedule wait time fixes in tests

ashish-jabble:
The test scenario has been failing intermittently on platforms since the FOGL-8788 merge. Instead of adding an extra wait after the stats history request call to check the READINGS stats count, implemented a consistent 15-second wait for the schedule to ensure it runs before the stats history call.

ran custom Ci Jobs:

```
ub18 + postgres - Build 850 - **PASSED**
ub20 + sqlite - Build 71 -  **PASSED**
centos stream9 + sqlite - Build 729 - **PASSED**
```

### DIFF
--- a/tests/system/python/api/test_statistics.py
+++ b/tests/system/python/api/test_statistics.py
@@ -150,7 +150,7 @@ class TestStatistics:
         assert 1 == stats['READINGS']
 
         # Allow stats collector schedule to run i.e. by default 15s
-        time.sleep(wait_time * 2 + 1)
+        time.sleep(wait_time * 3)
 
         # check stats history
         conn.request("GET", '/fledge/statistics/history')


tests/system/python/api/test_statistics.py
```diff
@@ -150,7 +150,7 @@ def test_statistics_history_with_service_enabled(self, start_south_coap, fledge_
         assert 1 == stats['READINGS']
 
         # Allow stats collector schedule to run i.e. by default 15s
-        time.sleep(wait_time * 2 + 1)
+        time.sleep(wait_time * 3)
```
praveen-garg, line 153: In Postgres specific CI job, what is the wait-time arg value?

ashish-jabble: > what is the wait-time arg value

CI used default i.e 5 seconds for python based tests except the `E2E`. And this issue is NOT only with Postgres we see with SQLite as well intermittently on Ubuntu Platform, Rpi bullseye but constantly fails on Rpi buster, CentOS Stream9

Mohit04tomar (approved):

PR #1474, opened 2024-10-21
FOGL-9206 Correct image names

MarkRiddoch:


ashish-jabble (changes requested):
images loading are fine but need Title section in [introduction](https://github.com/fledge-iot/fledge/blob/develop/docs/monitoring/introduction.rst), otherwise RST will complain about it.

ashish-jabble (approved):

PR #1495, opened 2024-11-13
FOGL-9252: Updated tests to send data to PI Server instead of connector relay

Mohit04tomar:
I ran following jobs on latest commit(`aeeaa2e`) and all tests which were failing due to issue in sending data to Connector Relay are passed:
1. http://jenkins.dianomic.com:8080/job/fledge_on_demand_code_coverage_ub2004/88/
2. http://jenkins.dianomic.com:8080/job/fledge_build_rpi0_ephat/1350/
 

Now, test will send data to PI Web API 2021 SP3 instead of Connector Relay.


tests/system/python/conftest.py
```diff
@@ -716,6 +716,108 @@ def _read_data_from_pi_web_api(host, admin, password, pi_database, af_hierarchy_
     return _read_data_from_pi_web_api
 
 
+@pytest.fixture
+def read_data_from_pi_asset_server():
```
praveen-garg, line 720: We did not already have a fixture to read data using PI Web API?!

praveen-garg (commented):
Are we using 2023 or older? If there a flexibility to choose from dropdown? 


We may need the same for supported EDS Versions.

tests/system/python/conftest.py
Mohit04tomar: Yes, we already have a fixture to read data from the PI Web API, but it lacks the flexibility to read data from different AF hierarchies. It only accesses a specific AF structure and cannot identify specific assets or data points if multiple FogLAMPs are sending data to the same AF location.

Mohit04tomar:
> Are we using 2023 or older? If there a flexibility to choose from dropdown?
> 
> We may need the same for supported EDS Versions.

We are using PI Web API 2021 SP3, and currently, the CI job lacks flexibility to choose from a dropdown. However, if needed, I can add functionality to select the PI Web API for data egress.

praveen-garg (dismissed):

praveen-garg (approved):

tests/system/python/e2e/test_e2e_modbus_c_pi.py
```diff
@@ -93,13 +93,13 @@ def start_south_north(self, reset_and_start_fledge, add_south, remove_directorie
         add_south(SOUTH_PLUGIN, south_branch, fledge_url, service_name=SVC_NAME, config=cfg,
                   plugin_lang="C", start_service=False, plugin_discovery_name=PLUGIN_NAME)
 
-        start_north_pi_server_c(fledge_url, pi_host, pi_port, pi_token)
+        start_north_pi_server_c_web_api(fledge_url, pi_host, pi_port, pi_db=pi_db, pi_user=pi_admin, pi_pwd=pi_passwd)
```
ashish-jabble, line 96: This change alone is not sufficient, as the tests rely on dependent values further downstream, which may still cause issues.

And related Job run says FAILED in console though XML reports is missing from the build run

tests/system/python/e2e/test_e2e_modbus_c_rtu_pi.py
```diff
@@ -84,13 +84,13 @@ def start_south_north(self, reset_and_start_fledge, add_south, skip_verify_north
                   plugin_lang="C", start_service=False, plugin_discovery_name=PLUGIN_NAME)
 
         if not skip_verify_north_interface:
-            start_north_pi_server_c(fledge_url, pi_host, pi_port, pi_token)
+            start_north_pi_server_c_web_api(fledge_url, pi_host, pi_port, pi_db=pi_db, pi_user=pi_admin, pi_pwd=pi_passwd)
```
ashish-jabble, line 87: This change alone is not sufficient, as the tests rely on dependent values further downstream, which may still cause issues.

This tests SKIPPED from nightly runs as "Test requires a connected serial port!!"

PR #1504, opened 2024-11-27
FOGL-9266 Enhance OMF Plugin Logging Functionality

ashwini-k-pandey:
This pull request introduces significant enhancements to the OMF plugin's logging functionality based on user feedback. The current method of creating the omf.log file by simply touching the file is not suitable for all customers. This update provides a more flexible configuration option and improves the file management process.


docs/OMF.rst
```diff
@@ -165,6 +165,9 @@ The *Basic* tab contains the most commonly modified items
     When using the PI Buffer Subsystem, set the action code to *create* which will allow new data to be compressed normally.
     One disadvantage of the *create* action code is that multiple values with the same timestamp will all be stored. 
 
+  - **Enable OMF Logging**: The Enable OMF Logging flag allows users to toggle the logging functionality on or off.
```
MarkRiddoch, line 168: This should probale not be described as logging as this will cause confusion with the logging we do to the syste log. We should probably call this detailed tracing of OMF messages. We shoudl also say the fiel created will be collected by the support bundle or may be viewed in the filesystem under the data directory.

MarkRiddoch (dismissed):
Looks fine to me. I would say please try tio following existing style for indentation etc, it makes the code neater. In general don't change things that don't need to be changed and don't reformat code as it makes the review harder.

Only one small documentation change that need doing.

C/plugins/common/http_sender.cpp
```diff
@@ -25,3 +30,37 @@ HttpSender::HttpSender()
...
+ * file accordingly. The priority is given to FLEDGE_DATA. If neither
+ * environment variable is set, an error message is printed, and an empty
+ * string is returned.
+ *
+ * @return A string representing the path to the OMF log file, or an empty
+ *         string if neither environment variable is set.
+ */
+std::string HttpSender::getOMFTracePath() 
+{
+    const char* fledgeData = getenv("FLEDGE_DATA");
```
RVerhoeff, line 48: Are these environment variables defined if Fledge is installed in its default location, that is, _/usr/local/fledge_? There is a subroutine in _utils.h_ called _getDataDir()_ that does all the logic for you.

C/plugins/north/OMF/plugin.cpp
```diff
@@ -348,7 +348,14 @@ const char *PLUGIN_DEFAULT_CONFIG_INFO = QUOTE(
...
 			"group": "Formats & Types",
 			"displayName": "Complex Types"
-		}
+		},
+        "EnableOMFLogging" : {
+            "description" : "If true, write to the omf.log file. If false, do not create log entries.",
+            "type" : "boolean",
+            "default" : "false",
+            "order" : "31",
+            "displayName" : "Enable OMF Logging"
```
RVerhoeff, line 357: Consistent with @MarkRiddoch 's comment, could this be "_Enable OMF Tracing_?"

praveen-garg: May be better to say,

Enable Tracing 
Or
Enable Debug Tracing

Repeating the OMF word is for some specific purpose?

ashish-jabble:
@ashwini-k-pandey Please rebase the branch correctly with develop, as there are changes in unrelated files that should not be included in this.

C/plugins/north/OMF/plugin.cpp
```diff
@@ -390,6 +397,7 @@ PLUGIN_INFORMATION *plugin_info()
  */
 PLUGIN_HANDLE plugin_init(ConfigCategory* configData)
 {
+    Logger::getLogger()->error("plugin_init called");
```
ashish-jabble, line 400: Remove this logger

C/plugins/north/OMF/plugin.cpp
```diff
@@ -468,6 +476,7 @@ string plugin_shutdown(PLUGIN_HANDLE handle)
 
 	string rval = info->saveData();
 	delete info;
+    Logger::getLogger()->error("plugin_shutdown called");
```
ashish-jabble, line 479: Remove this logger

C/plugins/north/OMF/plugin.cpp
```diff
@@ -341,7 +341,14 @@ const char *PLUGIN_DEFAULT_CONFIG_INFO = QUOTE(
 			"order": "29",
 			"group": "Formats & Types",
 			"displayName": "Complex Types"
-		}
+		},
+        "EnableTracing" : {
+            "description" : "If enabled, a detailed tracing of OMF Messages will be written to debug-trace/omf.log file in Fledge data directory.",
+            "type" : "boolean",
+            "default" : "false",
+            "order" : "31",
```
ashish-jabble, line 349: 30

C/plugins/common/http_sender.cpp
```diff
@@ -25,3 +33,66 @@ HttpSender::HttpSender()
...
+/**
+ * @brief Creates the "debug-trace" directory under the base directory returned by getDataDir().
+ * 
+ * This function ensures that the "debug-trace" directory is created only if the base directory exists
+ */
+bool HttpSender::createDebugTraceDirectory() 
+{
+    // Step 1: Retrieve the base data directory
+    std::string baseDir = getDataDir(); 
+    std::string debugTraceDir = baseDir + "/debug-trace";
```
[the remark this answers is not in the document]
ashwini-k-pandey: @ashish-jabble please confirm if there is any change from what we discussed this morning ..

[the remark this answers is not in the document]
ashish-jabble: `$FLEDGE_ROOT/data/logs/debug-trace` OR `$FLEDGE_DATA/logs/debug-trace`

See the related PR-https://github.com/fledge-iot/fledge/pull/1501 - A sample output tree structure is provided in the description.

C/plugins/common/include/http_sender.h
```diff
@@ -56,6 +56,25 @@ class HttpSender
 		virtual void setOCSClientSecret      (std::string& OCSClientSecret) = 0;
 		virtual void setOCSToken             (std::string& OCSToken) = 0;
 
+        /**
+         * @brief Constructs the file path for the OMF log based on environment variables.
+         *
+         * This function checks for the existence of two environment variables:
```
RVerhoeff, line 62: The implementation of _getOMFTracePath()_ now calls _getDataDir()_ which is great. Because of that change, this comment is no longer accurate. The function no longer requires the existence of _FLEDGE_ROOT_ and _FLEDGE_DATA_ and will no longer return an empty string if neither is set.

RVerhoeff (changes requested):
Update the comment in _http_server.cpp_.

docs/OMF.rst
```diff
@@ -165,6 +165,9 @@ The *Basic* tab contains the most commonly modified items
     When using the PI Buffer Subsystem, set the action code to *create* which will allow new data to be compressed normally.
     One disadvantage of the *create* action code is that multiple values with the same timestamp will all be stored. 
 
+  - **Enable Tracing**: The Enable Tracing flag allows users to toggle the Tracing functionality on or off. If enabled, a detailed tracing of OMF Messages will be written to `debug-trace/omf.log` file in Fledge data directory.
```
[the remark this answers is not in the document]
ashwini-k-pandey: OK

C/plugins/north/OMF/plugin.cpp
```diff
@@ -341,7 +341,14 @@ const char *PLUGIN_DEFAULT_CONFIG_INFO = QUOTE(
 			"order": "29",
 			"group": "Formats & Types",
 			"displayName": "Complex Types"
-		}
+		},
+        "EnableTracing" : {
+            "description" : "If enabled, a detailed tracing of OMF Messages will be written to debug-trace/omf.log file in Fledge data directory.",
```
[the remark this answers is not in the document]
ashwini-k-pandey: ok

C/plugins/common/include/http_sender.h
ashwini-k-pandey: ok

C/plugins/common/http_sender.cpp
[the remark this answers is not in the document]
ashwini-k-pandey: ok

C/common/include/utils.h
```diff
@@ -45,6 +45,16 @@ static const string getDataDir()
 	return (dataDir ? string(dataDir) : string(getRootDir() + "/data"));
 }
 
+/**
+ * @brief Constructs the path for the debug-trace dir based on the directory returned by getDataDir().
```
RVerhoeff, line 49: I wouldn't document a subroutine based on the behavior of another. How about "Constructs the path for the debug-trace subdirectory in the Fledge data directory" ?

C/common/include/utils.h
```diff
@@ -45,6 +45,16 @@ static const string getDataDir()
...
+ *
+ * @return A string representing the path to the debug-trace dir.
```
RVerhoeff, line 51: dir -> directory

C/plugins/common/http_sender.cpp
```diff
@@ -35,64 +35,71 @@ HttpSender::~HttpSender()
 }
 
 /**
- * @brief Creates the "debug-trace" directory under the base directory returned by getDataDir().
+ * @brief Creates the '/logs/debug-trace' directory under the directory returned by getDataDir().
```
RVerhoeff, line 38: Creates the '/logs/debug-trace' director in the Fledge data directory

C/plugins/common/http_sender.cpp
```diff
@@ -35,64 +35,71 @@ HttpSender::~HttpSender()
...
-    if (stat(baseDir.c_str(), &baseInfo) == 0 && (baseInfo.st_mode & S_IFDIR))
-    {
-        // Base directory exists, proceed to creating the debug-trace directory
+    // Ensure path consistency with getDebugTracePath(). Assert is commented out to prevent unexpected runtime interruptions in execution
+    // std::assert(debugTraceDir == getDebugTracePath()); 
 
-        // Step 3: Check if the debug-trace directory exists
-        struct stat debugTraceInfo;
-        if (stat(debugTraceDir.c_str(), &debugTraceInfo) != 0) 
+    auto createDir = [](const std::string& dirPath) -> bool
```
RVerhoeff, line 51: This code is fine but I wanted to point out that there is a method in our common library called _createDirectory_ (see [file_utils.cpp](https://github.com/fledge-iot/fledge/blob/develop/C/common/file_utils.cpp)). I did not comment before because you were doing additional work like setting access permissions. If any of these ideas could be put into _createDirectory_, that would be good to know.

C/plugins/common/include/http_sender.h
```diff
@@ -57,21 +57,14 @@ class HttpSender
...
-         *
-         * @return A string representing the path to the OMF log file, or an empty
-         *         string if neither environment variable is set.
+         * @return A string representing the path to the OMF log file.
          */
         static std::string getOMFTracePath();
 
         /**
-         * @brief Creates the "debug-trace" directory under the base directory returned by getDataDir().
+         * @brief Creates the "debug-trace" directory under the directory returned by getDataDir().
```
RVerhoeff, line 67: Creates the "debug-trace" subdirectory in the Fledge data directory

C/plugins/north/OMF/plugin.cpp
```diff
@@ -343,7 +343,7 @@ const char *PLUGIN_DEFAULT_CONFIG_INFO = QUOTE(
 			"displayName": "Complex Types"
 		},
         "EnableTracing" : {
-            "description" : "If enabled, a detailed tracing of OMF Messages will be written to debug-trace/omf.log file in Fledge data directory.",
+            "description" : "If enabled, a detailed tracing of OMF Messages will be written to logs/debug-trace/omf.log file in Fledge data directory.",
```
RVerhoeff, line 346: Messages -> messages. No need for upper case.

RVerhoeff (changes requested):
More updates to comments

C/plugins/common/http_sender.cpp
ashwini-k-pandey: There are two issue why I can not use createDirectory (see [file_utils.cpp](https://github.com/fledge-iot/fledge/blob/develop/C/common/file_utils.cpp)).
1. The function do not return status, it throws exception.
2. It do not check if the path is a file or a directory. It do not check this 
            if (dirInfo.st_mode & S_IFDIR)
            {
                return true; // Directory exists
            }

RVerhoeff: As I wrote, you don't need to use it. A couple of points though:

- I don't see why a method throwing an exception disqualifies it. The design was influenced by a scenario in S2OPCUA South in which I needed to create a complex directory subtree, one subdirectory at a time. All paths were created in code so the path strings were always correct and the directory creation was unlikely to fail. This way I could put all _createDirectory()_ calls into a single try block. If anything failed, the failed path would be in the exception string. This way, I don't have check a status code, log an error message and return false for every _createDirectory()_ call. If you only need to create one directory, put _createDirectory()_ in the `try` block and the error log in the `catch`.
- _createDirectory()_ does indeed not check that the path exists but is not a directory. Good idea; this can be added.

I see you went with a _mode_t_ value of 0755 which gives the Other group read and search access but not write. Method _createDirectory()_ uses _S_IRWXU | S_IRWXG_ which translates to 0770 meaning Other gets nothing. Your _mode_t_ setting is probably better. By the way, I found the _mode_t_ symbolic names on the [Gnu website](https://www.gnu.org/software/libc/manual/html_node/Permission-Bits.html). I also found a cool utility while researching this: the [Chmod Calculator](https://chmodcommand.com/)!

RVerhoeff (approved):

C/plugins/common/http_sender.cpp
ashwini-k-pandey: Thanks, @RVerhoeff , for sharing! It is very insightful.

In the meantime, I was wondering if we could use C++17 in our project. If we did, we would have access to the standard [create_directory](https://en.cppreference.com/w/cpp/filesystem/create_directory).

I am not aware of the background for sticking to C++11.

RVerhoeff: I don't know why either. Upgrading to C++ 17 might be a big project and would have to include the entire Fledge and FogLamp code bases. We will have to ask @MarkRiddoch.

PR #1509, opened 2024-12-02
FOGL-6371 pyjq pip dependency has been removed because it is no longer in use

ashish-jabble:


praveen-garg (commented):
lgtm!

Check, 
a) Should we remove fledge/python/fledge/common/jqfilter.py and tasks/north/sending_process.py; or add add a try-except around import pyjq
b) Update text in docs/building_fledge/building_fledge.rst

ashish-jabble:
> Should we remove fledge/python/fledge/common/jqfilter.py and tasks/north/sending_process.py; or add add a try-except around import pyjq

This will be handled separately in ticket FOGL-3258, as the code is unused and is never called directly by any process.

> Update text in docs/building_fledge/building_fledge.rst

It's just plain console output, and there are many instances of unused logs, which are essentially an outdated entries per version

MarkRiddoch (approved):

PR #1521, opened 2024-12-17
FOGL-9311 - Add  properties to key/value lists in C++

ashish-jabble:
Configuration support for optional attributes, such as keyName and keyDescription, in the kvlist for the C++


MarkRiddoch (approved):
Looks good to me, thank you.

PR #1537, opened 2025-01-09
FOGL-9031: Recover buffer memory

gnandan:


praveen-garg:
@gnandan Can you please add unit test for the same?

gnandan:
> @gnandan Can you please add unit test for the same?

Since changes are in south service which is an executable having main function. While writing Google Unit test it will also require a main function.  Complier will complain about multiple main function. So adding unit test cases is not feasible here. A executed following job [fledge_on_demand_code_coverage_ub2004] to ensure that no regression is there. Also looking some other ways  which can further help in testing the changes.

C/services/south/ingest.cpp
```diff
@@ -1082,11 +1083,20 @@ void Ingest::useFilteredData(OUTPUT_HANDLE *outHandle,
...
 		ingest->m_data = new vector<Reading *>;
+		isBufferMemory = true;
 	}
 	ingest->m_data->insert(ingest->m_data->end(), newData->cbegin(), newData->cend());
 	
 	readingSet->clear();
 	delete readingSet;
+
+	// Recover buffer memory generated to ingest data
+	if (isBufferMemory)
```
MarkRiddoch, line 1094: I do not believe this is safe as you have no idea what the ingest thread is doing with the memory, in particular Ingest::processQueue(). Once you have inserted the data into the ingest m_data queue that memory is owned by the ingest thread. It may work in some tests because of timing, but I don't think this is not a satisfactory solution.

C/services/south/ingest.cpp
```diff
@@ -720,12 +721,30 @@ void Ingest::processQueue()
 					// Block of code to execute holding the mutex
 					lock_guard<mutex> guard(m_qMutex);
 					std::vector<Reading *> *newQ = new vector<Reading *>;
+					// Recover buffer memory generated to ingest data
```
MarkRiddoch, line 724: This breaks the entire point of the change that added the issue, sending the data in the shutdown.

MarkRiddoch (commented):
Let's abandon this change and put the Jira back in the backlog

gnandan:
Closing this PR as these changes are not being taken forward.

PR #1547, opened 2025-02-04
FOGL-6845 fledge script inconsistency  

ashwini-k-pandey:


ashish-jabble:
@Mohit04tomar Run tests since it involves different areas.

scripts/fledge
```diff
@@ -334,14 +333,14 @@ fledge_stop() {
...
-  while [ $attempts -gt 0 ]; do
+  while [[ $attempts -gt 0 ]]; do
     sleep 1
     new_attempt=`fledge_status "silent"`
     case "$new_attempt" in
 
       0|1 )  # Still running
 
-        let attempts=attempts-1
+        attempts=$((attempts - 1))
```
ashish-jabble, line 343: Is this the main reason behind the issue?

ashwini-k-pandey: `let attempts=0` this caused failure is our case

RVerhoeff (dismissed):

scripts/fledge
```diff
@@ -363,7 +362,7 @@ fledge_stop() {
 
         echo
         fledge_log "info" "Fledge stopped." "all" "pretty"
-        let attempts=0
+        attempts=0
```
ashish-jabble, line 365: We should implement a graceful exit on success, indicating that the process has been stopped. In this case, the loop should `break` and not continue further. Additionally, for any other unexpected `new_attempt` status, the function should return an exit status of 1.

scripts/fledge
```diff
@@ -251,20 +251,19 @@ fledge_start() {
...
+            while [[ $attempts -gt 0 ]]; do
                 sleep 1
                 new_attempt=`fledge_status "silent"`
                 case "$new_attempt" in
 
                   "0")  # Started
                     echo
                     fledge_log "info" "Fledge started." "all" "pretty"
-                    let attempts=0
+                    attempts=0
```
ashish-jabble, line 262: same [comment](https://github.com/fledge-iot/fledge/pull/1547/files#r1942330912)

scripts/fledge
ashwini-k-pandey: The current code already manages the comment effectively. Line 364 outputs the log. Setting attempts = 0 ensures that the loop does not execute further.

ashish-jabble: I mean here to exit early with `break` without resetting  the counter variable, and add a case to handle any  unexpected status other than `0,1, or 2` exiting with status 1

MarkRiddoch (dismissed):
Doesn't appear to be related to the issue in the Jira, but the changes look ok.

scripts/fledge
```diff
@@ -251,20 +251,20 @@ fledge_start() {
...
```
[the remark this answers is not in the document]
ashwini-k-pandey: I would keep this line for safety

Mohit04tomar (approved):

ashish-jabble (approved):
> Doesn't appear to be related to the issue in the Jira, but the changes look ok.

I agree with Mark; we should stick to JIRA, and the PR summary/description should be more meaningful.


Additionally, using attempts=0 introduces an unnecessary change, as break is a cleaner, more direct, and readable approach. It clearly conveys the intent to exit the loop immediately.

PR #1549, opened 2025-02-05
FOGL-9481 listName optional attribute for the value when the type is a list

ashish-jabble:


MarkRiddoch (dismissed):

hemantgarg923 (approved):

python/fledge/common/configuration_manager.py
```diff
@@ -220,6 +220,15 @@ async def _run_callbacks_child(self, parent_category_name, child_category, opera
 
     async def _merge_category_vals(self, category_val_new, category_val_storage, keep_original_items,
                                    category_name=None):
+
+        def _modify_value_by_list_name(key_name):
+            old_value = json.loads(item_val_new[key_name])
+            # Create a new dictionary where listName is the key and old_value is the value,
+            # then convert the resulting dictionary into a JSON string
+            if item_val_new["listName"] not in old_value:
```
ashwini-k-pandey, line 228: should we check that 'old_value' is a list ?

python/fledge/common/configuration_manager.py
```diff
@@ -656,10 +668,16 @@ async def _create_new_category(self, category_name, category_val, category_descr
         try:
             if isinstance(category_val, dict):
                 new_category_val = copy.deepcopy(category_val)
-                # Remove "deprecated" items from a new category configuration
                 for i, v in category_val.items():
+                    # Remove "deprecated" items from a new category configuration
                     if 'deprecated' in v and v['deprecated'] == 'true':
                         new_category_val.pop(i)
+                    # Add "listName" to the value when the type is a list
+                    if 'listName' in v and v['type'] == 'list':
```
ashwini-k-pandey, line 676: i think check should be
```
if  v['type'] == 'list' and 'listName' in v:
```

python/fledge/common/configuration_manager.py
```diff
@@ -1210,6 +1234,12 @@ async def set_category_item_value_entry(self, category_name, item_name, new_valu
                 new_val = new_value_entry
                 await self._handle_update_config_for_acl(category_name, old_value, new_val)
 
+            # Special case: If type is list and listName is given then modify the value internally
```
ashwini-k-pandey, line 1237: The code section is duplicate of line 946. Please explore if can define them in a function

ashwini-k-pandey (changes requested):

python/fledge/common/configuration_manager.py
```diff
@@ -220,6 +220,15 @@ async def _run_callbacks_child(self, parent_category_name, child_category, opera
 
     async def _merge_category_vals(self, category_val_new, category_val_storage, keep_original_items,
                                    category_name=None):
+
+        def _modify_value_by_list_name(key_name):
+            old_value = json.loads(item_val_new[key_name])
```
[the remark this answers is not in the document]
ashish-jabble: This won't be possible, as it is called with a default and value key-value pair, which is part of the contract for a config item. If anything is missing, it will be caught by the initial functions.

python/fledge/common/configuration_manager.py
ashish-jabble: There is no functional difference, both will evaluate the same and produce the same result. However, checking the existence of 'listName' first allows for an early exit. If it doesn't exist, the second condition won't be evaluated.

python/fledge/common/configuration_manager.py
ashish-jabble: Both cases are somewhat different, and creating a generic function to handle them adds complexity.

python/fledge/common/configuration_manager.py
ashwini-k-pandey: No if 'old_value' is not a list, that is why the comment .

ashish-jabble: The code is no longer needed and has been removed, as we don't need to modify values in the non-upgrade scenario when the merge rule is applied.

MarkRiddoch (approved):

hemantgarg923 (approved):

ashwini-k-pandey (approved):

PR #1557, opened 2025-02-12
FOGL-9473 Validation of duplicate filters is enforced when updating the pipeline via API

ashish-jabble:


praveen-garg (approved):

firozalikhan (commented):
A filter used in control pipeline can still be used in south/north with the same name (using API, GUI does not allow it). Going to log a Jira to block this.

PR #1559, opened 2025-02-17
FOGL-9556 Add grouping to plugin documentation

MarkRiddoch:
Create additional sections in the plugin list section to give tables of plugins grouped by keywords.

Notification rule and delivery plugins are currently excluded as we have not added keywords for those.


docs/scripts/fledge_plugin_list
```diff
@@ -96,6 +96,103 @@ function table {
...
+			    fi
+                            echo "      - ${description}" >> "${keyfile}"
+                elif [[ $repo == "fledge-rule-DataAvailability" ]]; then
+			echo TODO deal with inbuilt data availability plugin
+                elif [[ $repo == "fledge-rule-Threshold" ]]; then
+			echo TODO deal with inbuilt threshold plugin
+                fi
+                else
+                    rm -rf ${repo}
+                    git clone https://${USERNAME}:${GITHUB_ACCESS_TOKEN}@github.com/fledge-iot/${repo}.git --branch ${DOCBRANCH} >/dev/null 2>&1
```
ashish-jabble, line 143: Pulling the repository twice will add unnecessary time. We should utilize the existing `function table` to handle this more efficiently.

docs/scripts/fledge_plugin_list
```diff
@@ -96,6 +96,103 @@ function table {
...
+	    echo $heading >> $output
+	    echo $heading | sed -e 's/[A-Za-z0-9 ]/~/g' >> $output
+	fi
+	echo >> $output
+	cat "${file}" >> $output
+	echo >> $output
+    done
+    if [[ -d /tmp/keywords ]]; then
+	rm -rf /tmp/keywords
+    fi
```
ashish-jabble, line 193: Duplicate by L#114 - L#116

docs/scripts/fledge_plugin_list
```diff
@@ -172,3 +269,36 @@ to extend the set of notification deliveries.
 
 EOF6
 table "$REPOSITORIES" "Fledge Notification Delivery Plugins" notify
+
+if [[ -d /tmp/keywords ]]; then
+	rm -rf /tmp/keywords
+fi
```
ashish-jabble, line 275: For the last attempt?

docs/scripts/fledge_plugin_list
```diff
@@ -172,3 +269,36 @@ to extend the set of notification deliveries.
...
+cat >>$output << EOF7
+
+South Plugins Grouped by Category
```
ashish-jabble, line 278: Category or Keyword?

docs/keywords/README.md
```diff
@@ -0,0 +1,4 @@
+Keyword section headers
+=======================
+
+This directory contains a set of markdown files whose names match keywords used in plugins. The file will be used to create an intruction in the generated sections for each group of plugins that mention the keyword named in the filename.
```
ashish-jabble, line 4: instruction

ashish-jabble (changes requested):
With this approach, we'd have to update the script every time we want to introduce a new keyword. Instead, we could build the keyword list by scanning all the keyword files and then reference them once the list is generated; same as had in detailed plugin documentation. This would make maintenance much easier. 

However, there's another challenge: how can we retrieve the descriptions of these keywords? We might need to pull them from a centralized source.

MarkRiddoch:
> With this approach, we'd have to update the script every time we want to introduce a new keyword. Instead, we could build the keyword list by scanning all the keyword files and then reference them once the list is generated; same as had in detailed plugin documentation. This would make maintenance much easier.
> 
> However, there's another challenge: how can we retrieve the descriptions of these keywords? We might need to pull them from a centralized source.

No we don't. We only have to add a header if we want to. There is a default mechanism if no keyword header file is found.

docs/scripts/fledge_plugin_list
MarkRiddoch: Keyword means less to users, hence category is a better title to use.

docs/keywords/README.md
MarkRiddoch: Yes, the readme is here to tell people what to do.

docs/scripts/fledge_plugin_list
MarkRiddoch: Yes, final cleanup to prevent leaving a lot of files around

ashish-jabble (approved):

PR #1564, opened 2025-02-18
FOGL-9540 Correct key name used when persisting data in pipeline element

MarkRiddoch:
shutdown code.


pintomax (approved):

firozalikhan:
@MarkRiddoch now entries are updating fine in the plugin_data table.
One behavior observed though, after deleting a filter from a pipeline "plugin_data" table still holds the entries for the deleted filter. Should we handle this in the same PR or log a Jira for this?

firozalikhan (approved):

PR #1568, opened 2025-02-20
FOGL-9458 Plugin data fixes are needed for filters, especially with the new schema

ashish-jabble:
- [x] New schema
- [x] Persisted plugins are now available under Developer tab for a service along with their data


firozalikhan:
@ashish-jabble  when a data persistent filter is being used in Control Pipeline, where would the persistent data appear?
If it is not implemented yet, do we need a fresh Jira for this?

MarkRiddoch (approved):
Looks fine to me

firozalikhan (approved):

PR #1584, opened 2025-03-05
FOGL-8895: update OMF North troubleshooting page

RVerhoeff:
Update the *Troubleshooting the PI Server integration* page to include:

- Differences between Complex and Linked Types
- New Information messages that log Types, Containers and Links created
- Sample error messages and what to do about them 

A couple of notes:

- I wanted to include a link to the main *OMF* documentation page, *Basic* tab section. I used a relative page reference *../plugins/fledge-north-OMF/index.html#basic*. This is hard to test because the pages don't have this relationship while in the repository.
- I used *.. code-block:: bash* instead of *.. code-block:: console* when documenting log messages. Some keywords were highlighted improperly when using *console*.


docs/troubleshooting_pi-server_integration.rst
```diff
@@ -56,20 +82,187 @@ screenshot from the Fledge GUI
...
+    If an OMF POST call returns an HTTP return code of 200 (OK), it means an item already exists and is correctly defined.
+    If an OMF POST call returns an HTTP return code of 201 (Created), it means a new item has been created.
+    
+Tracing File
+------------
+
+It is possible to generate a detailed trace of all OMF messages POSTed to the AVEVA web server for troubleshooting purposes.
+This applies to all AVEVA OMF web server types: PI Web API, AVEVA CONNECT and Edge Data Store.
+To enable this feature, click the *Enable Tracing* checkbox on the `OMF Basic tab <../plugins/fledge-north-OMF/index.html#basic>`_.
+
```
MarkRiddoch, line 221: We should add a note about having this features turned off in production use.

MarkRiddoch (changes requested):
This looks really good. The only thing I would say is add a warning about not leaving the off log file turned on. There is also a little inconsistency in header capitalisation, but I don't think it is worth solving this here as we probably have this all over the documentation.

MarkRiddoch (approved):

PR #1586, opened 2025-03-07
FOGL-9566 Monitor core with debugger status for each Southbound and Northbound service, along with updates in the service, South/North API

ashish-jabble:


python/fledge/services/core/service_registry/monitor.py
```diff
@@ -104,6 +104,11 @@ async def _monitor_loop(self):
                             res = json.loads(text)
                             if res["uptime"] is None:
                                 raise ValueError('res.uptime is None')
+                            # Set the 'debug' status for non-empty values, applicable only to
+                            # Southbound and Northbound services
+                            if (service_record._type in ('Southbound', 'Northbound') and
+                                    res.get("debug") not in [None, {}, '']):
+                                service_record._debug = res["debug"]
```
praveen-garg, line 111: We should think more around this. 

```
if service_record._type in ('Southbound', 'Northbound'):
    debug_value = res.get("debug", {})

    # Check if debug_value is not a dictionary
    if not isinstance(debug_value, dict):
        raise TypeError(f"Invalid 'debug' value: {debug_value} - Expected dict, but got {type(debug_value).__name__}.") # may be just log a warning?
    
    if debug_value:  # Only assign if debug_value is not empty
        service_record._debug = debug_value

```

python/fledge/common/service_record.py
```diff
@@ -61,6 +61,7 @@ def __init__(self, s_id, s_name, s_type, s_protocol, s_address, s_port, m_port):
             self._port = int(s_port)
         self._management_port = int(m_port)
         self._status = ServiceRecord.Status.Running
+        self._debug = {}
```
ashish-jabble, line 64: @MarkRiddoch As debug output for both Southbound and Northbound microservices returns `{"debug": {"debugger": "Detached"}}`. Should we keep this as the default, or is it fine to leave it empty? 
Please confirm.

MarkRiddoch: I suspect the GIU will find it easier if the they only have to look for one value rather than empty or a specific value.

ashish-jabble: Currently, the API returns a debug key-value pair when the object is non-empty, so it's not an issue for them to find it on their own.

MarkRiddoch (approved):

praveen-garg (approved):

PR #1589, opened 2025-03-13
FOGL-9639 Add mechanism to fledge/foglamp script to pass a cert for

MarkRiddoch:
authentication


pintomax (approved):

docs/quick_start/starting.rst
```diff
@@ -37,4 +37,16 @@ In both cases the user will be prompted to enter a password. It is possible, but
...
+
+.. code-block:: console
+
+   $ /usr/local/fledge/bin/fledge -c ~/.fledge/admin.cert stop
+   Stopping Fledge..........
+   Fledge Stopped
+
+.. note::
+
+   Extreme caution should be taken when storing certificate files that they not be readable by any other user within the system.
```
ashish-jabble, line 50: This note is intended for Admins who are responsible for setting up the FL instance with user-specific configurations. Only they have the ability to stop the FL instance using a certificate.
Given these conditions, we can proceed forward.Although the `systemctl stop` issue persists, a separate ticket to address it.

docs/quick_start/starting.rst
```diff
@@ -37,4 +37,16 @@ In both cases the user will be prompted to enter a password. It is possible, but
 
    The *start*, *status* and *help* commands do not require authentication.
 
+It is also possible to use certificate based authentication to login to the system. In this case the "fledge" command line utility should be passed the *-c* flag with the name of the certificate file to use to authenticate.
+
+.. code-block:: console
+
+   $ /usr/local/fledge/bin/fledge -c ~/.fledge/admin.cert stop
```
ashish-jabble, line 44: Instead of `~/.fledge/admin.cert`, we can simply refer to the certificate path.

scripts/fledge
```diff
@@ -805,14 +811,17 @@ get_fledge_version
 get_rest_api_url
 
 # Call getopt to get any command line options for username and password
-while getopts "u:p:" option; do
+while getopts "u:p:c:" option; do
```
ashish-jabble, line 814: We can add the `-h` option to display the help usage, showing all possible options for stopping the FL instance.

docs/quick_start/starting.rst
MarkRiddoch: > Instead of `~/.fledge/admin.cert`, we can simply refer to the certificate path.

We could, this is only an example and I wanted to avoid making specific reference to where Fledge stores certificates as it may well move in future.

scripts/fledge
MarkRiddoch: We already have fledge help, so the information is available. True that should be enhanced to cover the -u, -p and -c arguments.

docs/quick_start/starting.rst
ashish-jabble: Here, I mean a placeholder for the certificate path.

scripts/fledge
```diff
@@ -22,7 +22,7 @@ set -e
 #
 # This is the startup script for fledge
 #
-USAGE="Usage: `basename ${0}` {start|start --safe-mode|stop|status|reset|purge|kill|healthcheck|help|version}"
+USAGE="Usage: `basename ${0}` <-h> <-u username> <-p password> <-c certificate> {start|start --safe-mode|stop|status|reset|purge|kill|healthcheck|help|version}"
```
ashish-jabble, line 25: These flags are primarily used for cases like stop and health checks, and they might cause confusion in other scenarios.

ashish-jabble (approved):
The current solution still encounters an issue with systemctl stop, which is being addressed through other JIRA tickets.

PR #1591, opened 2025-03-14
FOGL-9653 Fix memory leaks in C/common unit tests and library

MarkRiddoch:
One definite leak left, in the unit test code where I catches an exception which is a pointer
==2576541== LEAK SUMMARY:
==2576541==    definitely lost: 1,600 bytes in 200 blocks
==2576541==    indirectly lost: 0 bytes in 0 blocks
==2576541==      possibly lost: 190,584 bytes in 149 blocks
==2576541==    still reachable: 14,528,263 bytes in 4,576 blocks
==2576541==         suppressed: 0 bytes in 0 blocks

`TEST(Categorytest, parseError)
{
        EXPECT_THROW(ConfigCategory("parseTest", json_parse_error), ConfigMalformed*);
}`

This is not usually an issue, but due to the way GTEST is picking up the exception there is no way to free the exception class.

Fixed leaks in ReadingSetCircularBuffer, 2D datapoint arrays, caused by non-virtual Reading destructor, PythonNumpy  issue and Image Encoding.

There are some dubious references in the Python code that should be the subject of another investigation.

### DIFF
--- a/C/common/JSONPath.cpp
+++ b/C/common/JSONPath.cpp
@@ -27,6 +27,10 @@ JSONPath::JSONPath(const string& path) : m_path(path)
  */
 JSONPath::~JSONPath()
 {
+	for (int i = 0; i < m_parsed.size(); i++)
+	{
+		delete m_parsed[i];
+	}
 }
 
 /**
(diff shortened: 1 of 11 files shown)


pintomax (approved):

ashish-jabble (approved):
No issues were found with the changes in the validation grind and on-demand tests run CI job.

PR #1599, opened 2025-03-28
FOGL-9650 - A new endpoint for user authentication certificate creation has been added to the admin API

ashish-jabble:


docs/securing_fledge.rst
```diff
@@ -349,23 +349,20 @@ Default ca certificate is available inside $FLEDGE_DATA/etc/certs and named as c
...
-b) Now, it's time to create user with name **test** (case sensitive). Also only admin can create user. Below are the cURL Commands
+b) It is now time to generate a new certificate for the **test** username that was created earlier.
 
 .. code-block:: console
 
-    $ AUTH_TOKEN=$(curl -d '{"username": "admin", "password": "fledge"}' -sX POST <PROTOCOL>://<FLEDGE_IP>:<FLEDGE_REST_API_PORT>/fledge/login | jq '.token' | tr -d '""')
-    $ curl -H "authorization: $AUTH_TOKEN" -skX POST <PROTOCOL>://<FLEDGE_IP>:<FLEDGE_REST_API_PORT>/fledge/admin/user -d '{"username":"test","real_name":"Test","access_method":"cert","description":"Non-admin based role","role_id":2}'
+    $ curl -H "authorization: $AUTH_TOKEN" -sX POST <PROTOCOL>://<FLEDGE_IP>:<FLEDGE_REST_API_PORT>/fledge/admin/$USER_ID/authcertificate
+
+You can now locate the **test** certificate within the $FLEDGE_DATA/etc/certs/, which will be used for login. It is advisable to relocate this certificate to a different location to prevent it from being shared with others.
```
praveen-garg, line 365: We better create a cert store entry and allow admin/ that specific user to download from certificate store via GUI.

We should add warn block for 


> For security concerns, it is highly recommended to download or copy the certificate and delete it from the certificate store. Relocate it to a secure, private location to prevent unauthorized access and sharing, using encrypted storage or password protection.

docs/securing_fledge.rst
```diff
@@ -349,23 +349,20 @@ Default ca certificate is available inside $FLEDGE_DATA/etc/certs and named as c
 
 Below are the steps to create custom certificate along with existing fledge based ca signed for auth certificates.
 
-a) Create a new certificate for username. Let say **test**
+a) To create a user with the name **test** (case sensitive), please note that only an admin has the permission to do so. The cURL commands are provided below.
```
praveen-garg, line 352: To create a user with a unique and case-insensitive username (e.g., uname1), you can follow these steps:

a) Using cURL

Create User
Create Authentication Certificate 

b) later via GUI

....

Note: _You need administrator privileges to perform this action._

docs/securing_fledge.rst
praveen-garg: > , which will be used for login. 

and can be used to log in."

docs/securing_fledge.rst
praveen-garg: We can also extend the create user API to allow generate auth cert with some flag in payload? To avoid 2 steps process?

ashish-jabble: Extension is okay but it is good to have separate endpoint and do not want to add all stuff in one go. It would also admin will see what action required for the user.

praveen-garg: Yes

Hence I meant a flag to allow as an option while creating user(name).

docs/securing_fledge.rst
```diff
@@ -349,23 +349,32 @@ Default ca certificate is available inside $FLEDGE_DATA/etc/certs and named as c
 
 Below are the steps to create custom certificate along with existing fledge based ca signed for auth certificates.
 
-a) Create a new certificate for username. Let say **test**
+**Using cURL**
 
-.. code-block:: console
+- To create a user with a unique name and case-insensitive. The cURL commands are provided below.
```
MarkRiddoch, line 354: The "and case-insensitive" does not read well. It maybe better to say that user names must be unique and are case insensitive in a note to draw attention this requirement.

docs/securing_fledge.rst
```diff
@@ -349,23 +349,32 @@ Default ca certificate is available inside $FLEDGE_DATA/etc/certs and named as c
...
 
-    $ AUTH_TOKEN=$(curl -d '{"username": "admin", "password": "fledge"}' -sX POST <PROTOCOL>://<FLEDGE_IP>:<FLEDGE_REST_API_PORT>/fledge/login | jq '.token' | tr -d '""')
-    $ curl -H "authorization: $AUTH_TOKEN" -skX POST <PROTOCOL>://<FLEDGE_IP>:<FLEDGE_REST_API_PORT>/fledge/admin/user -d '{"username":"test","real_name":"Test","access_method":"cert","description":"Non-admin based role","role_id":2}'
+    $ curl -H "authorization: $AUTH_TOKEN" -sX POST <PROTOCOL>://<FLEDGE_IP>:<FLEDGE_REST_API_PORT>/fledge/admin/$USER_ID/authcertificate
+
+You can now locate the **test** certificate within the $FLEDGE_DATA/etc/certs/ and can be used for login.
+
+.. warning::
+
+    For security concerns, it is highly recommended to download or copy the certificate and delete it from the certificate store. Relocate it to a secure, private location to prevent unauthorized access and sharing, using encrypted storage or password protection.
```
MarkRiddoch, line 377: We should probably be more explicit about which field to download. It might also be worth considering an API to do this, otherwise we rely upon users having access to the Linux machine to download the certificate.

python/fledge/services/core/api/auth.py
```diff
@@ -872,6 +875,51 @@ async def delete_user(request):
     return web.json_response({'message': "User has been deleted successfully."})
 
 
+@has_permission("admin")
```
MarkRiddoch, line 878: Maybe this should return the certificate (cert file) such that the user can save it rather than having a separate download step.

ashish-jabble: API has now been updated to retrieve the file content.

python/fledge/services/core/api/auth.py
```diff
@@ -872,6 +879,66 @@ async def delete_user(request):
...
+async def certificate(request):
```
praveen-garg, line 883: Please name it create_user_auth_certificate or generate_user_auth_certificate

python/fledge/services/core/api/auth.py
```diff
@@ -872,6 +879,66 @@ async def delete_user(request):
...
+        raise web.HTTPForbidden
+
+    user_id = int(request.match_info.get('user_id'))
+    try:
+        data = await request.json()
+        expiration_days = data.get("expiration_days", 365)
+        if not isinstance(expiration_days, int):
+            msg = "expiration_days must be an integer."
+            raise web.HTTPBadRequest(reason=msg, body=json.dumps(msg))
+        if expiration_days < 1 or expiration_days > 365:
```
praveen-garg, line 908: why max 365 is a hard limit?

praveen-garg (approved):

MarkRiddoch (approved):

PR #1600, opened 2025-03-28
FOGL-8155 Add hard limit for number of buffered readings in a south service

ashwini-k-pandey:


ashish-jabble:
First, please address the conflicts and then proceed with the compilation fix.

MarkRiddoch (approved):

python/fledge/services/core/server.py
```diff
@@ -318,6 +318,39 @@ class Server:
...
+            'order': '2',
+            "validity" : "serviceBuffering == \"Limited\"",
+            'permissions': ['admin']
+        },
+        'discardPolicy': {
+            'description': 'The different discard policies for the South Service',
+            'type': 'enumeration',
+            'displayName': 'Discard Policy',
+            'options': ['Discard Oldest', 'Reduce Fidelity', 'Discard Newest'],
+            'minimum': '1',
```
ashish-jabble, line 346: What is minimum here?

C/services/south/south.cpp
```diff
@@ -774,6 +779,83 @@ void SouthService::start(string& coreAddress, unsigned short corePort)
...
+
+		// Set default values to ensure the system can continue running
+		m_serviceBufferingType = SERVICE_BUFFER_BUFFER_TYPE_DEFAULT;
+		m_serviceBufferSize = SERVICE_BUFFER_SIZE_DEFAULT;
+		m_discardPolicy = SERVICE_BUFFER_DISCARD_POLICY_DEFAULT;
+
+		Logger::getLogger()->info("Using default configuration: "
+								"serviceBufferingType=UNLIMITED, "
+								"serviceBufferSize=1000, "
+								"discardPolicy=REDUCE_FIDELITY.");
```
ashish-jabble, line 855: Discard Oldest

ashish-jabble, line 855: Better to use variables m_serviceBufferingType, m_serviceBufferSize, m_discardPolicy in the logger values.

C/services/south/south.cpp
```diff
@@ -774,6 +779,83 @@ void SouthService::start(string& coreAddress, unsigned short corePort)
...
+
+		// Parse and validate the service buffer size
+		try 
+		{
+			std::string bufferSizeStr = m_configResourceLimit.getValue("serviceBufferSize");
+			m_serviceBufferSize = (unsigned int)std::stoi(bufferSizeStr); // Convert to integer
+
+			if (m_serviceBufferSize <= 0)
+			{
+				Logger::getLogger()->error("Invalid serviceBufferSize value: %s. Must be greater than zero. Using default: 1000.", bufferSizeStr.c_str());
```
ashish-jabble, line 811: SERVICE_BUFFER_SIZE_DEFAULT instead of using a hard-coded value of 1000.

C/services/south/south.cpp
```diff
@@ -774,6 +779,83 @@ void SouthService::start(string& coreAddress, unsigned short corePort)
...
+				m_serviceBufferSize = SERVICE_BUFFER_SIZE_DEFAULT; // Default value
+			}
+		} 
+		catch (const std::exception& e)
+		{
+			// Handle conversion errors and out-of-range values
+			Logger::getLogger()->error("Error parsing serviceBufferSize: %s. Using default value: 1000.", e.what());
```
ashish-jabble, line 818: same as above

C/services/south/south.cpp
```diff
@@ -774,6 +779,83 @@ void SouthService::start(string& coreAddress, unsigned short corePort)
...
+		{
+			m_serviceBufferingType = ServiceBufferingType::UNLIMITED;
+		} 
+		else if (serviceBuffering == "Limited") 
+		{
+			m_serviceBufferingType = ServiceBufferingType::LIMITED;
+		} 
+		else
+		{
+			Logger::getLogger()->error("Invalid serviceBuffering value: %s. Using default: UNLIMITED.", serviceBuffering.c_str());
```
ashish-jabble, line 799: SERVICE_BUFFER_BUFFER_TYPE_DEFAULT instead of using a hard-coded value of UNLIMITED

C/services/south/south.cpp
```diff
@@ -774,6 +779,83 @@ void SouthService::start(string& coreAddress, unsigned short corePort)
...
+		{
+			m_discardPolicy = DiscardPolicy::DISCARD_NEWEST;
+		} 
+		else if (discardPolicy == "ReduceFidelity") 
+		{
+			m_discardPolicy = DiscardPolicy::REDUCE_FIDELITY;
+		} 
+		else 
+		{
+			Logger::getLogger()->error("Invalid m_discardPolicy value: %s. Using default: REDUCE_FIDELITY.", discardPolicy.c_str());
```
ashish-jabble, line 838: DISCARD_OLDEST

SERVICE_BUFFER_DISCARD_POLICY_DEFAULT

C/services/south/include/ingest.h
```diff
@@ -47,6 +47,22 @@
...
+
+// Enum for discard policy
+enum class DiscardPolicy {
+	DISCARD_OLDEST,
+	REDUCE_FIDELITY,
+	DISCARD_NEWEST
+};
+
+#define SERVICE_BUFFER_BUFFER_TYPE_DEFAULT 	ServiceBufferingType::UNLIMITED
+#define SERVICE_BUFFER_DISCARD_POLICY_DEFAULT 	DiscardPolicy::REDUCE_FIDELITY
```
ashish-jabble, line 64: https://github.com/fledge-iot/fledge/blob/481bda78939e3c291581bfc9c2c6964898cb3c27/python/fledge/services/core/server.py#L347

The default Discard Policy must be uniform across all references.

python/fledge/services/core/server.py
```diff
@@ -643,6 +676,22 @@ async def core_logger_setup(cls):
...
+        """ Get the south service resource limit configuration """
+        try:
+            config = cls._RESOURCE_LIMIT_DEFAULT_CONFIG
+            category = 'RESOURCE_LIMIT'
+            description = "Resource Limit of South Service"
+            if cls._configuration_manager is None:
+                cls._configuration_manager = ConfigurationManager(cls._storage_client_async)
+            await cls._configuration_manager.create_category(category, config, description, True,
+                                                             display_name='Resource Limit')
+            config = await cls._configuration_manager.get_category_all_items(category)
```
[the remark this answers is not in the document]
ashwini-k-pandey: config is created in line 682. Any way removing the assignment

C/services/south/south.cpp
ashwini-k-pandey: m_serviceBufferingType and m_discardPolicy are enums , hence making less sense, other way was proving a enum to string. I avoided that since the conversion was not needed in main logic.

m_serviceBufferSize can be used in the log

ashish-jabble (commented):
Also it would be great if we add documentation regarding it under probably in tuning guide section.

python/fledge/services/core/server.py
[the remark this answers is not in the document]
ashwini-k-pandey: Yes, I have considered that. The reason I added it here is that it serves as a validation check to ensure successful creation. All other functions in this file follow the same pattern. Since these functions also utilize the newly created configuration, they do not explicitly highlight the validation step.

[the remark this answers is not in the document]
ashish-jabble: > The reason I added it here is that it serves as a validation check to ensure successful creation.

The create_category function is responsible for validating the creation of a category; if the validation fails, it will raise an Exception.

> All other functions in this file follow the same pattern

The existence of a category call indicates that the subsequent steps require the utilization of configuration values.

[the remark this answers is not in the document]
ashwini-k-pandey: ok, removed

docs/tuning_fledge.rst
```diff
@@ -933,3 +934,47 @@ Care should be taken when using performance counters, as with almost any system
...
+
++---------------------------------+
+| |resource_limit_south_advanced| |
++---------------------------------+
+
+The following parameters are available for configuration:
+
+  - **South Service Buffering** : Defines whether the buffering for South Services is unlimited or capped. If set to `"Limited"`, additional configuration options become applicable.  
+
+  - **South Service Limit** : Specifies the maximum number of readings that can be buffered in the South Service. This setting is only valid when the *South Service Buffering* option is set to `"Limited"`.
```
ashish-jabble, line 951: **South Service Buffer Size**

ashish-jabble:
Currently, there are conflicts that need to be addressed.

docs/tuning_fledge.rst
```diff
@@ -933,3 +934,47 @@ Care should be taken when using performance counters, as with almost any system
 
   Performance counters can be a very useful tool when tuning or debugging Fledge systems, but should **never** be left on during production use.
 
+
+Resource Limit Configuration of South Services
+==============================================
+
+Fledge includes a *Resource Limit* configuration for South Services, offering controls to manage resource usage and buffering effectively, thereby preventing excessive memory consumption or system overload due to unbounded buffering.
```
MarkRiddoch, line 941: This is rather long and rambling. Better to split it up. Also introduce the concept as this is intended for an end user who may not fully appreciate the internal mechanisms. Something like...

> South service within Fledge will buffer reading before forwarding them onto the internal storage service. The default behaviour is to buffer all readings until they reach a certain age or a certain number. When this occurs they will be sent to the storage layer. The threshold that cause data to be sent to the storage layer can be configured and are discussed elsewhere in this section of the documentation.
> 
> If there is a problem sending readings to the storage layer the south service will continue to buffer the data until that fault is cleared. This behaviour will cause the south service to consume increasing amounts of system memory until the fault clears. This can be undesirable if that consumption is not checked. The resource limit configuration for the south service allows the administrator to can the buffering that will occur in this case.

docs/tuning_fledge.rst
```diff
@@ -933,3 +934,47 @@ Care should be taken when using performance counters, as with almost any system
 
   Performance counters can be a very useful tool when tuning or debugging Fledge systems, but should **never** be left on during production use.
 
+
```
MarkRiddoch, line 937: Move this section up to just after the "South Service Advanced Configuration" section as it interacts with settings discussed in that section.

docs/tuning_fledge.rst
```diff
@@ -933,3 +934,47 @@ Care should be taken when using performance counters, as with almost any system
...
+Resource Limit Configuration of South Services
```
MarkRiddoch, line 938: Don't really need the work Configuration in the heading, tuning is all about configuration.

docs/tuning_fledge.rst
```diff
@@ -933,3 +934,47 @@ Care should be taken when using performance counters, as with almost any system
...
+
+     - **Reduce Fidelity**: Reduces the fidelity of buffered readings by discarding every second reading, starting from the oldest. This policy tracks the next reading to discard to avoid repeated reduction of fidelity for the same data.  
+     
+     - **Discard Newest**: Discards the newest readings to maintain the buffer size.  
+
+Access Control
+--------------
+Only users with administrative privileges can modify the **Resource Limit** configuration items.
+
+Dynamic Configuration Updates
```
MarkRiddoch, line 965: No need to say this. All Fledge configuration is Dynamic, with a few exceptions that should already be documented. Saying this will mislead people into thinking they need to do restarts for other things that do not require it.

MarkRiddoch (changes requested):

docs/tuning_fledge.rst
ashwini-k-pandey: Ok

docs/tuning_fledge.rst
ashwini-k-pandey: Ok

docs/tuning_fledge.rst
ashwini-k-pandey: OK

C/services/south/south.cpp
```diff
@@ -780,6 +785,116 @@ void SouthService::start(string& coreAddress, unsigned short corePort)
 	logger->info("South service shutdown %s completed", m_dryRun ? "from dry run " : "");
 }
 
+void SouthService::getResourceLimit()
```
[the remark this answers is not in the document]
ashwini-k-pandey: Ok

C/services/south/ingest.cpp
```diff
@@ -611,11 +791,11 @@ void Ingest::processQueue()
 				std::map<std::string, int>		statsEntriesCurrQueue;
 				AssetTracker *tracker = AssetTracker::getAssetTracker();
 				if (tracker == nullptr)
-                                {
-                                        Logger::getLogger()->error("%s could not initialize asset tracker",
-								__FUNCTION__);
+				{
+					Logger::getLogger()->error("%s could not initialize asset tracker",
```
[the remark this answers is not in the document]
ashwini-k-pandey: Done.
FYI.. This PR only modified the indentation

MarkRiddoch (approved):

firozalikhan (approved):

PR #1602, opened 2025-03-31
FOGL-9694 Add API blocks to stop south services as a source and north

MarkRiddoch:
services as a destination for control pipelines


python/fledge/services/core/api/control_service/pipeline.py
```diff
@@ -490,9 +490,25 @@ async def _check_parameters(payload, request):
...
+        schedules = await server.Server.scheduler.get_schedules()
+        south_schedules = [sch.name for sch in schedules if sch.schedule_type == 1 and sch.process_name == "south_c"]
+        if source_name in south_schedules:
+            raise ValueError("South services can not be the source for control pipelines")
+    # Destination can not be a north service
+    if destination is not None and des_type == 2:
+        schedules = await server.Server.scheduler.get_schedules()
+        north_schedules = [sch.name for sch in schedules if sch.schedule_type == 1 and sch.process_name == "north_C"]
+        if des_name in north_schedules:
+            raise ValueError("North services can not be the destination for control pipelines")
```
ashish-jabble, line 511: This block can be move up where we actually verifying the source and destination individually.

python/fledge/services/core/api/control_service/pipeline.py
```diff
@@ -490,9 +490,25 @@ async def _check_parameters(payload, request):
             if (source_name in south_schedules and des_name in south_schedules) or (
                     source_name in north_schedules and des_name in north_schedules):
                 raise ValueError(error_msg)
+            if source_name in south_schedules:
+                raise ValueError("South services can not be the source for control pipelines")
+            if des_name in north_schedules:
+                raise ValueError("North services can not be the destination for control pipelines")
```
ashish-jabble, line 496: From the outset of the initial implementation, this information was not known to us.
As per doc - 

```
The combination of source and destination must be unique for a given pipeline. No two pipelines can have the same source and destination defined, however multiple pipelines can exist with the same source if each has a different destination and likewise multiple pipelines can exist with the same destination if they each have a different source. There can only be one pipeline with a wildcard source and a wildcard destination
```

We should update the DOC as well

ashish-jabble (changes requested):
a) Error messages do not conclude with a full stop, unlike others.
b) unit tests are missing.

python/fledge/services/core/api/control_service/pipeline.py
MarkRiddoch: Doc has been updated

MarkRiddoch:
> b) unit tests are missing.

Can you point me at unit tests for the other conditions that I cn base any new unit tests on. I'm unfamiliar with pytest so this would be helpful?

python/fledge/services/core/api/control_service/pipeline.py
ashish-jabble: Thanks

ashish-jabble:
> Can you point me at unit tests for the other conditions that I cn base any new unit tests on. I'm unfamiliar with pytest so this would be helpful?

See the [PR changes](https://github.com/fledge-iot/fledge/pull/1604) on top of this branch

ashish-jabble (approved):

MarkRiddoch:
> > Can you point me at unit tests for the other conditions that I cn base any new unit tests on. I'm unfamiliar with pytest so this would be helpful?
> 
> See the [PR changes](https://github.com/fledge-iot/fledge/pull/1604) on top of this branch

Thanks @ashish-jabble

PR #1610, opened 2025-04-10
Compatible with all fledge based supported platforms for the prerequisite setup of C-based unit tests

ashish-jabble:


tests/unit/C/requirements.sh
```diff
@@ -17,34 +17,38 @@
...
-    sudo apt-get update -y
     sudo apt-get install -y libgtest-dev
 
-   if [[ ${OS_VERSION} == *"18."* ]]; then
-        echo "Building Google Test libraries manually..."
-        cd /usr/src/gtest
-        sudo cmake -E make_directory build
-        sudo cmake -E chdir build cmake ..
-        sudo cmake --build build
+    echo "Building Google Test libraries manually..."
```
ashwini-k-pandey, line 35: "Building Google Test libraries manually..." is needed only for "Ubuntu 18"

ashish-jabble: This indicates that the manual steps are unnecessary for other Ubuntu versions, and only the libgtest-dev package is needed. 

@Mohit04tomar We must verify this on Raspberry Pi OS; it is confirmed for Buster, but we need to investigate its status on Bullseye and Bookworm.

PR #1629, opened 2025-05-05
FOGL-9746 implementation of Remote UDP Logging for syslog

ashwini-k-pandey:
### **Remote UDP Logging:**

1. **Added support for logging messages to a remote syslog via UDP protocol.**
The logging system initializes a UDP socket (m_UdpSockFD) and sends log messages to a remote server defined by environment variables.

1. **Environment Variables:**
Introduced the following environment variables for dynamic configuration:
**SYSLOG_UDP_ENABLED:** Enables or disables UDP logging. If set to true, logs are sent to the configured UDP sink.
**LOG_IP:** Specifies the IP address of the remote logging server. Defaults to 127.0.0.1 if not set.
**LOG_PORT:** Specifies the port number for the remote logging server. Defaults to 5140 if not set.
The environment variables allow customization without requiring code changes, making the system adaptable to different deployment environments

### DIFF
--- a/C/common/logger.cpp
+++ b/C/common/logger.cpp
@@ -15,12 +15,16 @@
 #include <memory>
 #include <string.h>
 #include <sys/time.h>
+#include <sys/socket.h>
 #include <exception>
+#include <arpa/inet.h>
 
 using namespace std;
 
 // uncomment line below to get uSec level timestamps
 // #define ADD_USEC_TS
+const char * DEFALUT_LOG_IP = "127.0.0.1";
+const int DEFAULT_LOG_PORT = 5140;
 
 inline long getCurrTimeUsec()
 {
@@ -28,6 +32,20 @@ inline long getCurrTimeUsec()
 	gettimeofday(&m_timestamp, NULL);
 	return m_timestamp.tv_usec;
 }
+
+/**
+ * Compare two strings in a case insensitive manner
+ *
+ * @param str1		The first string
+ * @param str2		The second string
+ * @return bool		True if the strings are equal, false otherwise
+ */
+bool caseInsensitiveCompare(const std::string& str1, const std::string& str2) {
+    if (str1.size() != str2.size()) return false;
+    return std::equal(str1.begin(), str1.end(), str2.begin(),
+        [](char c1, char c2) { return std::tolower(c1) == std::tolower(c2); });
+}
+
 
 /**
  * The singleton pointer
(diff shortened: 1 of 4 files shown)


C/common/logger.cpp
```diff
@@ -29,6 +33,20 @@ inline long getCurrTimeUsec()
...
 }
 
+/**
+ * Compare two strings in a case insensitive manner
+ *
+ * @param str1		The first string
+ * @param str2		The second string
+ * @return bool		True if the strings are equal, false otherwise
+ */
+bool caseInsensitiveCompare(const std::string& str1, const std::string& str2) {
```
MarkRiddoch, line 43: You don't use this, why is it here? It is not within the Logger class or belong with the Logger functionality.

C/common/logger.cpp
```diff
@@ -58,9 +76,53 @@ static char ident[80];
 	{
 		strncpy(ident, application.c_str(), sizeof(ident));
 	}
-	openlog(ident, LOG_PID|LOG_CONS, LOG_USER);
+
+	//openlog(ident, LOG_PID|LOG_CONS, LOG_USER);
```
MarkRiddoch, line 80: Don't leave commented out code in the final checkin.

C/common/logger.cpp
```diff
@@ -402,30 +452,55 @@ void Logger::error(const string& msg, ...)
...
-/**
- * Apply the formatting to the error message
- * 
- * @param format	The printf format string
- * @param args		The printf argument list
- * @return string	The formatted string
- */
-string *Logger::format( const std::string& format, va_list args)
+
+void Logger::log(int lvl, const char * lvlName, const string& msg, ...)
```
MarkRiddoch, line 465: Please put back the doxygen header for the renamed method.

MarkRiddoch (changes requested):
It is probably the fault of your editing tool, but the code formatting of the C++ code is messed up. Please make sure you use something that shows tabs correctly and correct the formatting.

C/common/logger.cpp
ashwini-k-pandey: Done

MarkRiddoch (dismissed):
This gives us the short term workaround for logging with containers. The problem is it also means that the ability to view logs and create support bundles has been compromised. Longer term we will need to think about a better way to tackle this. However we should in terms of short term merge this for customer use.

ashwini-k-pandey:
@MarkRiddoch The current solution addressing the sending logs over UDP in a Docker container. I agrees that reading log from syslog has few limitations.  We should enhance the log reading, log writing, and log searching capabilities in FogLAMP holistically, several improvements can be proposed to create a more robust and efficient logging system.  If you agree I would raise a JIRA with my proposal on this

ashwini-k-pandey:
@MarkRiddoch From FogLAMP running in the docker container, we can see the Logs from GUI and create the support bundle, same as usual FogLAMP instance. There is no change in user experience with these changes.

MarkRiddoch (dismissed):

MarkRiddoch (approved):

PR #1630, opened 2025-05-05
FOGL-9792: FledgeAsset Type mismatch error when Static Data properties change

RVerhoeff:
Support for Static Data values for Linked Types was added in [FOGL-9056](https://scaledb.atlassian.net/browse/[FOGL-9056](https://scaledb.atlassian.net/browse/FOGL-9056)). Static Data values become AF Attribute Templates in the AF Template _FledgeAsset_. Upgrading existing Fledge systems failed because OMF North would attempt to define _FledgeAsset_ with Static Data values that did not exist before. This resulted in OMF HTTP 409 (Conflict) which would stop all processing until "unstable" PI was corrected. This has been fixed. If creation of _FledgeAsset_ results in HTTP 409, OMF North will write the message "_FledgeAsset Type exists with a different definition_" and continue processing normally.

Sending Static Data in later OMF Data messages is benign. Static Data properties that do not exist in the AF Elements derived from _FledgeAsset_ are ignored.

### DIFF
--- a/C/plugins/north/OMF/omf.cpp
+++ b/C/plugins/north/OMF/omf.cpp
@@ -5313,13 +5313,11 @@ bool OMF::sendFledgeAssetType()
 	catch (const Unauthorized &e)
 	{
 		Logger::getLogger()->error(MESSAGE_UNAUTHORIZED);
-		
 	}
 	catch (const Conflict& e)
 	{
-		handleRESTException(e, "The OMF endpoint reported a Conflict when sending FledgeAsset Type");
-		Logger::getLogger()->warn(MESSAGE_PI_UNSTABLE, 409);
-		m_PIstable = false;
+		Logger::getLogger()->warn("FledgeAsset Type exists with a different definition");
+		retCode = true;
 	}
 	catch (const std::exception &e)
 	{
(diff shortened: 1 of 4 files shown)


C/plugins/north/OMF/omf.cpp
```diff
@@ -5313,13 +5313,11 @@ bool OMF::sendFledgeAssetType()
...
 	{
 		Logger::getLogger()->error(MESSAGE_UNAUTHORIZED);
-		
 	}
 	catch (const Conflict& e)
 	{
-		handleRESTException(e, "The OMF endpoint reported a Conflict when sending FledgeAsset Type");
-		Logger::getLogger()->warn(MESSAGE_PI_UNSTABLE, 409);
-		m_PIstable = false;
+		Logger::getLogger()->warn("FledgeAsset Type exists with a different definition");
```
MarkRiddoch, line 5319: Can we give the user any help here? Seeing this in the log it is unclear what, if anything the user can do about it.

RVerhoeff: I am working on this. It's more complicated than I thought. It turns out that if you edit Static Data values, Linked Type instances will update the Static Data values in all AF Element instances. In Complex Types, this doesn't happen.

ashish-jabble:
@Mohit04tomar Could you please execute the CI tests? Several CI jobs are failing due to the issue outlined.

Mohit04tomar (approved):

PR #1639, opened 2025-05-19
FOGL-9793: Updates in e2e/test_e2e_pi_scaleset.py as per changes in scale-set filter plugn

Mohit04tomar:


tests/system/python/e2e/test_e2e_pi_scaleset.py
```diff
@@ -39,7 +39,7 @@
 SENSOR_VALUE = 21
 
 # scale(set) factor
-SCALE = "9/5"
+SCALE = "1.8"
```
ashish-jabble, line 42: So, the problem occurs due to the scale-set filter now using a list of objects with a float type, whereas it was previously in JSON format.

tests/system/python/e2e/test_e2e_pi_scaleset.py
```diff
@@ -39,7 +39,7 @@
...
 OFFSET = "32"
 OUTPUT = (SENSOR_VALUE * eval(SCALE)) + int(OFFSET)
```
ashish-jabble, line 44: > eval(SCALE)

Do we really need to evaluate SCALE again?

PR #1648, opened 2025-06-03
FOGL-9648 Improve the time precisions

ashwini-k-pandey:
1) Improve the time precisions earlier we log time in second only. Changes in this PR would report time in microsecond.

2) Added the time zone information

Here is the new time format
image

### DIFF
--- a/C/common/logger.cpp
+++ b/C/common/logger.cpp
@@ -142,10 +142,10 @@ Logger::~Logger()
 		instance = NULL;
 	else if (!instance)
 		return;	// Already destroyed
+	m_runWorker = false;
 	m_condition.notify_one();
 	if (m_workerThread && m_workerThread->joinable())
 	{
-		m_runWorker = false;
 		m_workerThread->join();
 		delete m_workerThread;
 		m_workerThread = NULL;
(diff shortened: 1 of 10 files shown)


MarkRiddoch (dismissed):

python/fledge/tasks/purge/purge.py
```diff
@@ -133,7 +134,8 @@ async def purge_data(self, config):
         unsent_retained = 0
         duration = 0
         method = None
-        start_time = time.strftime('%Y-%m-%d %H:%M:%S.%s', time.localtime(time.time()))
+        timezone = pytz.timezone('UTC')  # You can replace 'UTC' with your desired timezone
+        start_time = datetime.now(timezone).strftime('%Y-%m-%d %H:%M:%S.%f %Z%z')
```
MR901, line 138: Why have both `%Z` and `%z` been used together?

Also, I feel that the following way is cleaner, it doesn't even require `pytz` module

```python
from datetime import datetime, timezone
datetime.now(timezone.utc).strftime("%Y-%m-%d %H:%M:%S.%f%z")
```

MR901 (approved):

PR #1659, opened 2025-06-17
Added merge method in ReadingSet class

gnandan:


C/common/reading_set.cpp
```diff
@@ -213,6 +214,46 @@ ReadingSet::append(vector<Reading *>& readings)
...
+	auto orig_size = m_readings.size();
+
+	// Extend m_readings with the new readings
+	for (auto it = readings->begin(); it != readings->end(); ++it)
+	{
+		if ((*it)->hasId() && (*it)->getId() > m_last_id)
+		{
+			m_last_id = (*it)->getId();
+		}
+		m_readings.push_back(*it);
```
ashwini-k-pandey, line 238: assuming that m_readings are ordered do not appears to be valid assumption. Are we not making weak contract here. Repercussions of the broken contacts is undefined.

C/common/reading_set.cpp
```diff
@@ -213,6 +214,46 @@ ReadingSet::append(vector<Reading *>& readings)
 	readings.clear();
 }
 
+/**
+ * merge the readings in a vector with the set of readings in the reading set.
+ * @param readings	A vector of Reading pointers to merge with the ReadingSet
```
ashwini-k-pandey, line 219: mention that readings must be ordered

C/common/reading_set.cpp
```diff
@@ -213,6 +214,46 @@ ReadingSet::append(vector<Reading *>& readings)
...
+		if ((*it)->hasId() && (*it)->getId() > m_last_id)
+		{
+			m_last_id = (*it)->getId();
+		}
+		m_readings.push_back(*it);
+		m_count++;
+	}
+
+	// In-place merge, assuming both halves are sorted
+	std::inplace_merge(
```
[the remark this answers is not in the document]
ashwini-k-pandey: @gnandan / @MarkRiddoch 
Here is the libc++ implementation of `inplace_merge`
``` 
The std::inplace_merge algorithm in C++, as implemented in libc++, aims to merge two consecutive sorted ranges within a single container into a single sorted range, in place. Its performance characteristics are primarily determined by the availability of temporary memory.
**Complexity:**
Optimal Case (with auxiliary buffer):
If libc++ can successfully allocate a temporary buffer of sufficient size (linear in the number of elements N in the combined range), inplace_merge achieves linear time complexity, O(N), in terms of comparisons. This is because the merge can effectively be performed by copying elements to and from the buffer.
Worst Case (without auxiliary buffer):
If libc++ cannot allocate a temporary buffer (e.g., due to memory constraints or if the implementation chooses not to), inplace_merge falls back to an in-place merging strategy that typically involves rotations and other operations. In this scenario, its time complexity degrades to O(N log N), similar to a general-purpose sort.
```

Issues in current implementation 
We copy the pointer and then merge.  since merge its-self does hidden memory allocation and copies we have multiple copies in this logic.

To avoid multiple copies following algo could be used
```
vector<Readings*> newSet;
newSet.resevre(M+N);
i = 0;
while(p1 || p2){
  if(p1 && p2 && cmp(p2>p1)) 
    newSet[i] = *p2++
  else if (p1 && p2)
   newSet[i] = *p1++;
  else if(p1)
   newSet[i] = *p1++;
   else if(p2)
    newSet[i] = *p1++;
   i++;
}
```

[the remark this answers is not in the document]
gnandan: removed inplace merge

ashwini-k-pandey (approved):

C/common/reading_set.cpp
MarkRiddoch: No, Ashwini is right, the doxygen comment is the long lived documentation of the method. The Jira will be lost in a sea of other Jiras. This comment should document the method without reference to the Jira. It should also mention what happens to the readings parameter in this comment.

gnandan: @MarkRiddoch Added doxygen comment

C/common/reading_set.cpp
```diff
@@ -213,6 +213,57 @@ ReadingSet::append(vector<Reading *>& readings)
...
+
+	size_t i = 0;
+	auto p1 = m_readings.begin();
+	auto p2 = readings->begin();
+
+	while (p1 != m_readings.end() || p2 != readings->end()) {
+		if (p1 != m_readings.end() && p2 != readings->end()) {
+			struct timeval ta, tb;
+			(*p1)->getUserTimestamp(&ta);
+			(*p2)->getUserTimestamp(&tb);
```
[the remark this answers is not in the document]
gnandan: Yes, it could be, but maintaining the previous timestamp would require additional logic without offering much benefit, as getUserTimestamp simply returns the saved timestamp without any further calculation.

PR #1660, opened 2025-06-17
Added configuration category convenience methods

gnandan:


C/common/config_category.cpp
```diff
@@ -458,6 +458,164 @@ string ConfigCategory::getValue(const string& name) const
...
+	try
+	{
+		return getValue(name);
+	}
+	catch (ConfigItemNotFound& e)
+	{
+		Logger::getLogger()->info("'%s' %s , returning default value '%s'", name.c_str(), e.what(), defaultValue.c_str());
+		return defaultValue;
+	}
+	catch(...)
```
ashwini-k-pandey, line 479: Catch all behaviour is not documented in JIRA.
I think remove this line since getValue() throws only ConfigItemNotFound() exception, because absorbing unwanted errors could potentially hide bugs, that could have been caught during development / testing.

C/common/config_category.cpp
```diff
@@ -458,6 +458,164 @@ string ConfigCategory::getValue(const string& name) const
...
+		if (lower == "false" || lower == "0") return false;
+		Logger::getLogger()->info("Config item '%s' expected to be boolean but got '%s'", name.c_str(), val.c_str());
+		return defaultValue;
+	}
+	catch (ConfigItemNotFound& e)
+	{
+		Logger::getLogger()->info("'%s' %s , returning default value '%d'", name.c_str(), e.what(), defaultValue);
+		return defaultValue;
+	}
+	catch(...)
```
ashwini-k-pandey, line 510: same as above

C/common/config_category.cpp
```diff
@@ -458,6 +458,164 @@ string ConfigCategory::getValue(const string& name) const
...
+ * @param name         The name of the item
+ * @param defaultValue The value to return if item is not found or invalid
+ * @return bool        The boolean value
+ */
+bool ConfigCategory::getBoolValue(const std::string& name, bool defaultValue) const
+{
+	try
+	{
+		string val = getValue(name);
+		std::string lower = val;
```
ashwini-k-pandey, line 498: avoid copy here use 'val'

C/common/config_category.cpp
```diff
@@ -458,6 +458,164 @@ string ConfigCategory::getValue(const string& name) const
...
+		std::transform(lower.begin(), lower.end(), lower.begin(), ::tolower);
+		if (lower == "true" || lower == "1") return true;
+		if (lower == "false" || lower == "0") return false;
+		Logger::getLogger()->info("Config item '%s' expected to be boolean but got '%s'", name.c_str(), val.c_str());
```
ashwini-k-pandey, line 502: I think when configItem exist but excepted datatype mismatch then returning default value could be misleading.
@MarkRiddoch  we should raise exception here, caller should know about this condition.

C/common/config_category.cpp
```diff
@@ -458,6 +458,164 @@ string ConfigCategory::getValue(const string& name) const
...
+			name.c_str(), e.what(), defaultValue);
+		return defaultValue;
+	}
+	catch (std::out_of_range& e)
+	{
+		Logger::getLogger()->info("Config item '%s' out of range: %s, returning default value '%d'", 
+			name.c_str(), e.what(), defaultValue);
+		return defaultValue;
+	}
+	catch(...)
```
ashwini-k-pandey, line 544: same as above, applies to other places as well

C/common/config_category.cpp
```diff
@@ -458,6 +458,164 @@ string ConfigCategory::getValue(const string& name) const
...
+int ConfigCategory::getIntegerValue(const std::string& name, int defaultValue) const
+{
+	try
+	{
+		string val = getValue(name);
+		return stoi(val);
+	}
+	catch (ConfigItemNotFound& e)
+	{
+		Logger::getLogger()->info("'%s' %s , returning default value '%d'", name.c_str(), e.what(), defaultValue);
```
ashwini-k-pandey, line 529: JIRA do not say that we should write message when item do not exit. I think we can be these as debug log. Applies to all places

tests/unit/C/common/test_config_category.cpp
```diff
@@ -362,6 +362,57 @@ const char *kvlistObjectConfig = "{ \"name\": {"
...
 
+const std::string DefaultConfigJson = R"({
+	"bool_true":     { "type": "boolean", "default": "true", "value": "true" },
+	"bool_false":    { "type": "boolean", "default": "false", "value": "false" },
+	"int_val":       { "type": "integer", "default": "10", "value": "42" },
+	"long_val":      { "type": "integer", "default": "1000", "value": "10000000000" },
+	"double_val":    { "type": "float",   "default": "3.14", "value": "2.718" },
+	"invalid_bool":  { "type": "boolean", "default": "true", "value": "maybe" },
+	"invalid_int":   { "type": "integer", "default": "5", "value": "not_a_number" },
+	"invalid_double":{ "type": "float", "default": "1.0", "value": "oops" }
```
ashwini-k-pandey, line 373: add out of range tests for int/long/float type

tests/unit/C/common/test_config_category.cpp
```diff
@@ -362,6 +362,57 @@ const char *kvlistObjectConfig = "{ \"name\": {"
...
+	EXPECT_TRUE(cat.getBoolValue("non_existing_bool", true)); // fallback
+	EXPECT_FALSE(cat.getBoolValue("invalid_bool", false)); // wrong type fallback
+}
+
+TEST(CategoriesTest, GetIntegerValue)
+{
+	ConfigCategory cat("test", DefaultConfigJson);
+	EXPECT_EQ(cat.getIntegerValue("int_val"), 42);
+	EXPECT_EQ(cat.getIntegerValue("non_existing_int", 123), 123); // fallback
+	EXPECT_EQ(cat.getIntegerValue("invalid_int", -1), -1); // wrong type fallback
```
ashwini-k-pandey, line 397: out of range test .? applies to all tests

ashwini-k-pandey (changes requested):

C/common/config_category.cpp
```diff
@@ -458,6 +458,164 @@ string ConfigCategory::getValue(const string& name) const
...
+long ConfigCategory::getLongValue(const std::string& name, long defaultValue) const
+{
+	try
+	{
+		string val = getValue(name);
+		return stol(val);
+	}
+	catch (ConfigItemNotFound& e)
+	{
+		Logger::getLogger()->info("'%s' %s , returning default value '%ld'", name.c_str(), e.what(), defaultValue);
```
MarkRiddoch, line 563: Debug rather than info

MarkRiddoch (commented):
All the info messages you have added should be debug, if they are needed at all

C/common/config_category.cpp
MarkRiddoch: This is extremely unlikely to happen as the API for configuration categories will guard against it. We would only end up getting the error if somebody went in and edited the database directly.

C/common/config_category.cpp
MarkRiddoch: Probably debug rather than info if anything at all

C/common/config_category.cpp
MarkRiddoch: Again range checking is done with the API

C/common/config_category.cpp
MarkRiddoch: OK, I had missed that. Leave it as info.

MarkRiddoch (dismissed):

C/common/config_category.cpp
gnandan: Catchall is removed

C/common/config_category.cpp
gnandan: Catchall is removed

tests/unit/C/common/test_config_category.cpp
gnandan: range checking is done be API

tests/unit/C/common/test_config_category.cpp
gnandan: range checking is done be API

ashwini-k-pandey (approved):

MarkRiddoch (approved):

PR #1661, opened 2025-06-20
FOGL-9728: Add support for float-to-integer and integer-to-float data type coercion in OMF North

RVerhoeff:
The description of this Jira describes the impact on OMF of sending a stream of integers to PI and then introducing a filter that changes the integers to floating point numbers. The Scale Filter is one example. This problem can now be corrected by introducing a *number* OMFHint that sets the data type to *int64*. Data type coercion is now supported.

The section *Detecting the Data Type Mismatch Problem* in the OMF North documentation was updated to describe this feature and to fix details that have changed because of recent changes to the plugin.


RVerhoeff:
@MarkRiddoch , this is not quite done. I forgot to handle the coercion of floating point to _unsigned_ integer.

MarkRiddoch (approved):

PR #1668, opened 2025-07-11
FOGL-9981 Fledge restart and shutdown fixes

ashish-jabble:
Problem with Python 3.12 and aiohttp 3.10.11 specific resolution: wait_closed() hang indefinitely or block for minutes when there are active connections.

Therefore, we addressed the asyncio timeout exception to ensure a forced restart in such scenarios. As we approach the release, I have refrained from modifying the aiohttp specific release version, as this stable version encompasses our functionality and includes the necessary test dependencies.

### DIFF
--- a/python/fledge/services/core/server.py
+++ b/python/fledge/services/core/server.py
@@ -1707,20 +1707,22 @@ class Server:
         :Example:
             curl -X POST http://localhost:<core mgt port>/fledge/service/shutdown
         """
-        try:
-            await cls._stop()
-            loop = request.loop
-            # allow some time
+        async def _shutdown_event_loop(loop):
             await async_sleep(2.0)
             _logger.info("Stopping the Fledge Core event loop. Good Bye!")
             loop.stop()
 
-            return web.json_response({'message': 'Fledge stopped successfully. '
-                                                 'Wait for few seconds for process cleanup.'})
+        try:
+            await cls._stop()
+            await _shutdown_event_loop(request.loop)
+        except asyncio.TimeoutError as err:
+            await _shutdown_event_loop(request.loop)
         except TimeoutError as err:
             raise web.HTTPInternalServerError(reason=str(err))
         except Exception as ex:
             raise web.HTTPInternalServerError(reason=str(ex))
+        else:
+            return web.json_response({'message': 'Fledge stopped successfully. Wait for few seconds for process cleanup.'})
 
     @classmethod
     async def restart(cls, request):
--- a/python/fledge/services/core/server.py
+++ b/python/fledge/services/core/server.py
@@ -1252,7 +1252,7 @@ class Server:
             try:
                 await asyncio.wait_for(cls.service_server.wait_closed(), timeout=5.0)
             except asyncio.TimeoutError:
-                _logger.warning("REST server wait_closed() timeout - continuing with shutdown")
+                _logger.debug("REST server wait_closed() timeout - continuing with shutdown")
         else:
             await cls.service_server.wait_closed()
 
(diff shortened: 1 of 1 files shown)


python/fledge/services/core/server.py
```diff
@@ -1697,29 +1707,33 @@ async def shutdown(cls, request):
         :Example:
             curl -X POST http://localhost:<core mgt port>/fledge/service/shutdown
         """
-        try:
-            await cls._stop()
-            loop = request.loop
-            # allow some time
+        async def _shutdown_event_loop(loop):
```
praveen-garg, line 1710: _stop_event_loop

python/fledge/services/core/server.py
```diff
@@ -1697,29 +1707,33 @@ async def shutdown(cls, request):
...
+            return web.json_response({'message': 'Fledge stopped successfully. Wait for few seconds for process cleanup.'})
 
     @classmethod
     async def restart(cls, request):
         """ Restart the core microservice and its components """
-        try:
-            await cls._stop()
+
+        async def _restart_process():
+            """ Handle the logic for restarting the current process """
```
praveen-garg, line 1732: logic? current?

python/fledge/services/core/server.py
```diff
@@ -1728,10 +1742,13 @@ async def restart(cls, request):
                 sys.argv.append('')
 
             python3 = sys.executable
-            os.execl(python3, python3, *sys.argv)
+            os.execl(python3, python3, *sys.argv)  # Replaces current process, no return
 
-            return web.json_response({'message': 'Fledge stopped successfully. '
```
praveen-garg, line 1733: What was it doing earlier?

ashish-jabble: NOTHING! As it replaces current process, no return

python/fledge/services/core/server.py
```diff
@@ -1245,7 +1245,17 @@ async def stop_rest_server(cls):
...
         cls.service_server.close()
-        await cls.service_server.wait_closed()
+
+        # Python 3.12 + aiohttp 3.10.11 specific fix: wait_closed() can hang indefinitely
+        # if there are active connections, so we add a timeout
+        if sys.version_info >= (3, 12):
+            try:
+                await asyncio.wait_for(cls.service_server.wait_closed(), timeout=5.0)
+            except asyncio.TimeoutError:
+                _logger.warning("REST server wait_closed() timeout - continuing with shutdown")
```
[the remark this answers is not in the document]
praveen-garg: I mean what if added timeout exceeds, then what will happen and what end user action is possible?

firozalikhan:
PR resolves the issue, have checked the restart and kill/stop/reset/start behaviour on Ubuntu24, Ubuntu22 & Ubuntu20, with authentication and HTTPS on and off.
Deployment with FogLAMP Manage is also good.

python/fledge/services/core/server.py
praveen-garg: `restart` def is bound to endpoint, We must return some response to caller.

ashish-jabble:
> Seems this is [fixed in aiohttp==3.12.14 ](https://github.com/aio-libs/aiohttp/blame/master/CHANGES.rst#L27)and was introduced in  ??

in between `3.9.0 - 3.10.11`

> @Mohit04tomar Can we run some CI jobs on this?

Already ran nightly tests Job (on all Ub platforms) and no regression found as such

> check Restart endpoint JSON response on GUI. The restart is scheduled ...

```
curl -sX PUT localhost:8081/fledge/restart | jq
{
  "message": "Fledge restart has been scheduled."
}

$ curl -sX PUT localhost:8081/fledge/shutdown | jq
{
  "message": "Fledge shutdown has been scheduled. Wait for few seconds for process cleanup."
}

```

python/fledge/services/core/server.py
ashish-jabble: It is there
```
$ curl -sX PUT localhost:8081/fledge/restart | jq
{
  "message": "Fledge restart has been scheduled."
}
```

firozalikhan:
Restart API Verified

Request URL: PUT http://192.168.1.130:8081/foglamp/restart

Status Code: 200 OK

Response:
```
{
  "message": "FogLAMP restart has been scheduled."
}

```

PR #1672, opened 2025-07-15
FOGL-9962 : Added support for auto generating support bundle

gnandan:


docs/tuning_fledge.rst
```diff
@@ -977,3 +978,19 @@ Care should be taken when using performance counters, as with almost any system
...
+The support bundle is a collection of diagnostic data about the Fledge instance, used to identify and troubleshoot system issues more effectively. 
+The following configuration parameters control the automatic generation and retention of support bundles:
+
++--------------------------------+
+| |support_bundle_configuration| |
++--------------------------------+
+
+ - **Auto Generate On Failure**: This option controls whether a support bundle is automatically created when a service failure occurs. By default, this is set to **true**. When enabled, a support bundle is generated and saved in the **support** directory within the Fledge data directory. An alert is also triggered to notify the user that the bundle has been created.
+
+ - **Bundles To Retain**: This setting defines how many support bundles should be retained. The default and minimum value is **1**, and the maximum number of bundles that can be retained is **3**. If the number of stored bundles exceeds this limit, the oldest one is automatically deleted to make room for the new bundle.
```
ashish-jabble: The specified maximum threshold for the number of bundles is too low.

python/fledge/services/core/service_registry/monitor.py
```diff
@@ -54,6 +54,8 @@ def __init__(self):
 
         self.restarted_services = []
         self._acl_handler = None
+        self._auto_support_bundle = True # type: bool
+        self._support_bundle_retain_count = 1 # type: int
```
ashish-jabble: Please include comments explaining the purpose of these variables.

python/fledge/services/core/service_registry/monitor.py
```diff
@@ -145,13 +147,48 @@ async def _monitor_loop(self):
...
+            
+            self._logger.info("Support bundle created: %s for failed service: %s",
+                            bundle_name, service_name)
+        except Exception as ex:
+            self._logger.error("Failed to create support bundle for %s: %s",
+                            service_name, str(ex))
+    
+    async def raise_support_bundle_alert(self, service_name, bundle_name):
+        """Raise alert for automated support bundle creation"""
+        from fledge.services.core import server
```
ashish-jabble: We should avoid cyclic import as server also imports monitor.

scripts/plugins/storage/postgres/upgrade/76.sql
```diff
@@ -0,0 +1,9 @@
+INSERT INTO fledge.configuration ( key, display_name, description, value )
+     VALUES ( 'SUPPORT_BUNDLE',
+              'Support Bundle',
+              'Support Bundle Configuration',
+              '{"auto_support_bundle":{"description": "Automatically create support bundle when service fails","type": "boolean","default": "true","displayName": "Auto Generate On Failure","value":"true"},"support_bundle_retain_count":{"description": "Number of support bundles to retain (minimum 1)","type": "integer","default": "3","minimum": "1","displayName": "Bundles To Retain","value":"3"}}'
+            );
+
+INSERT INTO fledge.category_children (parent, child) VALUES ('Advanced', 'SUPPORT_BUNDLE');
+
```
ashish-jabble: We should not require downgrade or upgrade scripts for new category or merge. It is the responsibility of the core service to be established and to create a child relationship with Advanced.

python/fledge/services/core/support.py
```diff
@@ -145,10 +146,10 @@ async def build(self):
     def check_and_delete_bundles(self, support_dir):
         files = glob.glob(support_dir + "/" + "support*.tar.gz")
         files.sort(key=os.path.getmtime)
-        if len(files) >= _NO_OF_FILES_TO_RETAIN:
-            for f in files[:-2]:
-                if os.path.isfile(f):
-                    os.remove(os.path.join(support_dir, f))
+        if len(files) >= self._num_of_files_to_retain:
+            for i in range(len(files) - (self._num_of_files_to_retain-1)):
```
ashish-jabble: Use slicing instead of range(len(...))

`num_to_remove = len(files) - self._num_of_files_to_retain + 1`

`files[:num_to_remove]` is much clearer and more Pythonic than looping over index values.

python/fledge/services/core/support.py
```diff
@@ -145,10 +146,10 @@ async def build(self):
...
+                if os.path.isfile(files[i]):
+                    os.remove(os.path.join(support_dir, files[i]))
```
ashish-jabble: Avoid redundant path joins

You were checking `os.path.isfile(files[I])`, which would look in the current working directory, not support_dir.

It may be like a fix is to join support_dir and the filename first.

```
file_path = os.path.join(support_dir, file)
if os.path.isfile(file_path):
    os.remove(file_path)
```

ashish-jabble: Also, add error handling. Use try-except to catch OSError.

python/fledge/services/core/service_registry/monitor.py
```diff
@@ -145,13 +147,48 @@ async def _monitor_loop(self):
...
                         self._logger.info("Failed to audit service failure %s", str(ex))
             await self._sleep(self._sleep_interval)
 
+    async def create_automated_support_bundle(self, service_name):
+        """Create support bundle asynchronously when service fails"""
+        try:
+            from fledge.services.core.support import SupportBuilder
+            from fledge.common.common import _FLEDGE_DATA
+            support_dir = _FLEDGE_DATA + "/support" if _FLEDGE_DATA else _FLEDGE_ROOT + "/data/support"
+            builder = SupportBuilder(support_dir, self._support_bundle_retain_count)
```
ashish-jabble: Does the service-related support bundle contain information regarding the service name? The on-demand support bundle is only labeled with a support prefix and a timestamp, such as (`support-250716-06-14-24.tar.gz`). It would be beneficial to include the service name in the naming convention of the support bundle to facilitate easier identification of the support bundle.

python/fledge/services/core/service_registry/monitor.py
```diff
@@ -145,13 +147,48 @@ async def _monitor_loop(self):
...
+                            service_name, str(ex))
+    
+    async def raise_support_bundle_alert(self, service_name, bundle_name):
+        """Raise alert for automated support bundle creation"""
+        from fledge.services.core import server
+        try:
+            param = {
+                "key": f"{bundle_name}",
+                "message": f"Support bundle {bundle_name} created for failed service '{service_name}'",
+                "urgency": "3"  # Normal urgency
```
ashish-jabble: Normal?

python/fledge/services/core/service_registry/monitor.py
```diff
@@ -145,13 +147,48 @@ async def _monitor_loop(self):
...
+        except Exception as ex:
+            self._logger.error("Failed to create support bundle for %s: %s",
+                            service_name, str(ex))
+    
+    async def raise_support_bundle_alert(self, service_name, bundle_name):
+        """Raise alert for automated support bundle creation"""
+        from fledge.services.core import server
+        try:
+            param = {
+                "key": f"{bundle_name}",
```
ashish-jabble: if bundle_name is already a string, using f"{bundle_name}" is redundant.

python/fledge/services/core/support.py
```diff
@@ -45,9 +44,11 @@ class SupportBuilder:
     _out_file_path = None
     _interim_file_path = None
     _storage = None
+    _num_of_files_to_retain = 1
```
ashish-jabble, line 47: `num_files_to_keep` or `num_files_to_retain`

python/fledge/services/core/service_registry/monitor.py
```diff
@@ -145,13 +147,48 @@ async def _monitor_loop(self):
...
+            support_dir = _FLEDGE_DATA + "/support" if _FLEDGE_DATA else _FLEDGE_ROOT + "/data/support"
+            builder = SupportBuilder(support_dir, self._support_bundle_retain_count)
+            bundle_name = await builder.build()
+            # Raise alert about support bundle creation
+            await self.raise_support_bundle_alert(service_name, bundle_name)
+            
+            self._logger.info("Support bundle created: %s for failed service: %s",
+                            bundle_name, service_name)
+        except Exception as ex:
+            self._logger.error("Failed to create support bundle for %s: %s",
```
ashish-jabble: use multistacktrace syntax for ex

python/fledge/services/core/service_registry/monitor.py
```diff
@@ -145,13 +147,48 @@ async def _monitor_loop(self):
                 if check_count[service_record._id] > self._max_attempts:
                     ServiceRegistry.mark_as_failed(service_record._id)
                     check_count[service_record._id] = 0
+                    if self._auto_support_bundle:
+                        self._logger.info("Service %s failed, creating automated support bundle",
+                                          service_record._name)
+                        asyncio.ensure_future(self.create_automated_support_bundle(service_record._name))
```
ashish-jabble: Is it works well with Python3.12?

ashish-jabble (changes requested):
I have provided some comments on the preliminary dry review and also include unit tests.

docs/tuning_fledge.rst
MarkRiddoch: Yes, 3 is the current value, and should probably remain the default, but the upper limit should be settable much higher than this. I would suggest something like 20 or even more. If a system has a probably I can image setting this high to get a series of support bundles and then bringing it down when the issue is resolved. The documentation should;l warn of the dangers of setting this too high and the impact on disk usage.

python/fledge/services/core/service_registry/monitor.py
MarkRiddoch: It could, although we should make sure that we create a log entry that says we are creating the support bundle before we create it. That way we have the second inside the support bundle itself and don't rely upon the name to know the reason.

MarkRiddoch (commented):
The number of bundles to retain should be applied to the manually created support bundles as well.  If we set it higher, have a number of automatically created ones and then do a manual one, we don't want to loose a large number of support bundles.

python/fledge/services/core/service_registry/monitor.py
gnandan: @MarkRiddoch Should alert- generated for automated support bundle- have Normal urgency or some other urgency level?

python/fledge/services/core/support.py
gnandan: Underscore is prefixed with name according to the existing naming convention

docs/tuning_fledge.rst
gnandan: Fixed

python/fledge/services/core/service_registry/monitor.py
gnandan: done

python/fledge/services/core/service_registry/monitor.py
gnandan: Fixed

scripts/plugins/storage/postgres/upgrade/76.sql
gnandan: Removed upgrade / downgrade scripts

python/fledge/services/core/support.py
gnandan: Fixed

python/fledge/services/core/support.py
gnandan: Fixed

python/fledge/services/core/service_registry/monitor.py
gnandan: Added service name in the support bundle, Logging is already there.

python/fledge/services/core/service_registry/monitor.py
gnandan: Fixed

python/fledge/services/core/service_registry/monitor.py
gnandan: Fixed

python/fledge/services/core/service_registry/monitor.py
gnandan: Yes, it works with Python 3.12, However, I modified it to use create_task() method instead of ensure_future() method because it ensure_furture() is deprecated.

MarkRiddoch (dismissed):

python/fledge/services/core/api/support.py
```diff
@@ -18,7 +18,7 @@
 from fledge.common.logger import FLCoreLogger
 from fledge.common.web.middleware import has_permission
 from fledge.services.core.support import SupportBuilder
-
+from fledge.common.configuration_manager import ConfigurationManager
```
ashish-jabble: This import is only utilized for obtaining configuration; therefore, we can inline the import directly within the method itself, as other related import statements are present.

python/fledge/services/core/server.py
```diff
@@ -615,6 +615,32 @@ async def password_config(cls):
...
+                    "description": "Automatically create support bundle when service fails",
+                    "type": "boolean",
+                    "default": "true",
+                    "displayName": "Auto Generate On Failure"
+                },
+                "support_bundle_retain_count": { "description": "Number of support bundles to retain (minimum 1)",
+                "type": "integer",
+                "default": "3",
+                "minimum": "1",
+                "displayName": "Bundles To Retain"
```
ashish-jabble: fix indent

python/fledge/services/core/support.py
```diff
@@ -60,11 +61,12 @@ def __init__(self, support_dir):
             _LOGGER.error(ex, "Error in initializing SupportBuilder class.")
             raise RuntimeError(str(ex))
 
-    async def build(self):
+    async def build(self, service_name=None):
```
ashish-jabble: bundle_name

python/fledge/services/core/support.py
```diff
@@ -60,11 +61,12 @@ def __init__(self, support_dir):
...
         try:
             today = datetime.datetime.utcnow()
             file_spec = today.strftime('%y%m%d-%H-%M-%S')
-            tar_file_name = self._out_file_path+"/"+"support-{}.tar.gz".format(file_spec)
+            support_file_name = "support-{}-{}".format(service_name, file_spec) if service_name else "support-{}".format(file_spec)
```
ashish-jabble: Use f-strings

support_file_name = f"support-{service_name}-{file_spec}" if service_name else f"support-{file_spec}"

python/fledge/services/core/api/support.py
```diff
@@ -110,7 +110,9 @@ async def create_support_bundle(request):
     """
     support_dir = _get_support_dir()
     try:
-        bundle_name = await SupportBuilder(support_dir).build()
+        support_bundle_config = await get_support_bundle_config()
+        num_of_files_to_retain = max(1, int(support_bundle_config['support_bundle_retain_count']['value']))
+        bundle_name = await SupportBuilder(support_dir, num_of_files_to_retain).build()
```
ashish-jabble: The support bundle configuration object should be integrated into the Core Support builder, as we may need to include additional configuration items in the future. 

And extract retain value from config

python/fledge/services/core/support.py
```diff
@@ -45,9 +44,11 @@ class SupportBuilder:
     _out_file_path = None
     _interim_file_path = None
     _storage = None
+    _num_of_files_to_retain = 1
 
-    def __init__(self, support_dir):
+    def __init__(self, support_dir, num_of_files_to_retain=1):
```
ashish-jabble: As per above, the config can either be passed in explicitly or the configuration logic can be incorporated directly within the constructor itself.
This may also eliminates the duplicity of imports for storage connection and configuration manager

tests/unit/python/fledge/services/core/api/test_support.py
```diff
@@ -362,3 +373,338 @@ def mock_syslog(_level):
                 res = await resp.text()
                 jdict = json.loads(res)
                 assert actual_count == jdict['count']
+
+    async def test_automated_support_bundle_on_service_failure(self, client):
+        """Test that support bundle is automatically created when service fails"""
+        from fledge.common.service_record import ServiceRecord
```
ashish-jabble: Where this will be used?

tests/unit/python/fledge/services/core/api/test_support.py
```diff
@@ -362,3 +373,338 @@ def mock_syslog(_level):
...
+                    # Test manual support bundle creation (which simulates automated creation)
+                    resp = await client.post('/fledge/support')
+                    res = await resp.text()
+                    jdict = json.loads(res)
+                    assert 200 == resp.status
+                    assert "bundle created" in jdict
+                    
+                    # Verify the bundle name format for manual creation (no service name)
+                    bundle_name = jdict["bundle created"]
+                    assert bundle_name == "support-180301-13-35-23.tar.gz"
```
ashish-jabble: This test solely confirms the creation of the support bundle with the specified name in case of service failure, and it does not outline the actual test case scenario for the service failure.

tests/unit/python/fledge/services/core/api/test_support.py
```diff
@@ -362,3 +373,338 @@ def mock_syslog(_level):
...
+
+    async def test_automated_support_bundle_disabled(self, client):
```
ashish-jabble: This test case is same as above except the auto_support_bundle: true/false

This test case should be parameterised a tests which accepts config

tests/unit/python/fledge/services/core/api/test_support.py
```diff
@@ -362,3 +373,338 @@ def mock_syslog(_level):
...
+                    res = await resp.text()
+                    jdict = json.loads(res)
+                    assert 200 == resp.status
+                    assert "bundle created" in jdict
+
+                    # Verify the bundle name format for manual creation
+                    bundle_name = jdict["bundle created"]
+                    assert bundle_name == "support-180301-13-35-23.tar.gz"
+
+    async def test_support_builder_with_service_name(self, client):
```
ashish-jabble: Is this test case already included in test_automated_support_bundle_on_service_failure? Here, we are solely verifying the support bundle by its name.

We should see parameterized test, review the current tests, and utilize them.

tests/unit/python/fledge/services/core/api/test_support.py
```diff
@@ -362,3 +373,338 @@ def mock_syslog(_level):
...
+                "description": "Number of support bundles to retain (minimum 1)",
+                "type": "integer",
+                "default": "3",
+                "minimum": "1",
+                "displayName": "Bundles To Retain"
+            }
+        }
+        
+        # Mock the build method to raise an exception
+        async def mock_build_failure(service_name=None):
```
ashish-jabble: service_name unused?

tests/unit/python/fledge/services/core/api/test_support.py
```diff
@@ -362,3 +373,338 @@ def mock_syslog(_level):
...
+            raise RuntimeError("Support bundle creation failed")
+        
+        # Test that no alert is generated when support bundle creation fails
+        with patch.object(support, 'get_support_bundle_config', return_value=mock_config):
+            with patch.object(SupportBuilder, "__init__", return_value=None):
+                with patch.object(SupportBuilder, "build", side_effect=mock_build_failure):
+                    # Mock the storage connection to avoid ServiceRegistry issues
+                    with patch.object(connect, 'get_storage_async') as mock_storage:
```
ashish-jabble: mock_storage unused?

tests/unit/python/fledge/services/core/api/test_support.py
```diff
@@ -362,3 +373,338 @@ def mock_syslog(_level):
...
+                            alert_manager = AlertManager(mock_storage)
+                            result = await alert_manager.add(alert_params)
+                            
+                            # Verify that the alert was created with correct urgency level
+                            assert result == {"alert": alert_params}
+                            mock_alert_add.assert_called_once_with(alert_params)
+                            
+                            # Verify the urgency level is correct (Normal = 3)
+                            created_alert = result["alert"]
+                            assert created_alert["urgency"] == "3"
```
ashish-jabble: Add empty line at the end

ashish-jabble (changes requested):
We have combined the unit test cases in the single location (api which is NOT correct)

- Core/support serves as the primary builder module
- api/support utilizes the core builder functions and returns http status code along with response appropriately
- Alert generation and service failure support for bundle generation in the monitoring module is not included here.

Due to the extensive use of mocking, which does not encompass all potential scenarios required, it is advisable to separate the cases into distinct modules.

Additionally, review the lines of Python code coverage that pertain to the new changes you have included in your tests. This will also allow you to assess the coverage and scenarios you have addressed.

python/fledge/services/core/api/support.py
gnandan: Fixed

python/fledge/services/core/server.py
gnandan: Fixed

python/fledge/services/core/support.py
gnandan: Fixed

python/fledge/services/core/support.py
gnandan: Fixed

python/fledge/services/core/api/support.py
gnandan: Fixed

python/fledge/services/core/support.py
gnandan: Fixed

tests/unit/python/fledge/services/core/api/test_support.py
gnandan: moved test case to correct location

tests/unit/python/fledge/services/core/api/test_support.py
gnandan: Fixed

MarkRiddoch (dismissed):

python/fledge/services/core/service_registry/monitor.py
```diff
@@ -145,13 +150,48 @@ async def _monitor_loop(self):
...
                     except Exception as ex:
                         self._logger.info("Failed to audit service failure %s", str(ex))
             await self._sleep(self._sleep_interval)
 
+    async def create_automated_support_bundle(self, service_name):
+        """Create support bundle asynchronously when service fails"""
+        try:
+            from fledge.services.core.support import SupportBuilder
+            from fledge.common.common import _FLEDGE_DATA
+            support_dir = _FLEDGE_DATA + "/support" if _FLEDGE_DATA else _FLEDGE_ROOT + "/data/support"
```
ashish-jabble, line 169: `_FLEDGE_ROOT` undefined? import is missing?

python/fledge/services/core/service_registry/monitor.py
```diff
@@ -145,13 +150,48 @@ async def _monitor_loop(self):
...
+            builder = SupportBuilder(support_dir, self._support_bundle_config)
+            bundle_name = await builder.build(service_name)
+            # Raise alert about support bundle creation
+            await self.raise_support_bundle_alert(service_name, bundle_name)
+            
+            self._logger.info("Support bundle created: %s for failed service: %s",
+                            bundle_name, service_name)
+        except Exception as ex:
+            self._logger.error(ex, "Failed to create support bundle for {}".format(service_name))
```
ashish-jabble: Avoid mixing string formatting styles; choose one between % and .format() and apply it consistently.
The suggested method for string formatting is to utilize `f-strings`.

python/fledge/services/core/service_registry/monitor.py
```diff
@@ -191,11 +231,13 @@ async def _read_config(self):
         await cfg_manager.create_category('SMNTR', default_config, 'Service Monitor', display_name='Service Monitor')
 
         config = await cfg_manager.get_category_all_items('SMNTR')
+        support_bundle_config = await cfg_manager.get_category_all_items('SUPPORT_BUNDLE')
```
ashish-jabble: Cannot we directly assign with `self._support_bundle_config =` ?

tests/unit/python/fledge/services/core/service_registry/test_monitor.py
```diff
@@ -141,3 +141,135 @@ class TestMonitorException(Exception):
...
+                mock_interest_instance.unregister.return_value = None
+                
+                with patch('asyncio.create_task', side_effect=mock_create_task):
+                    # Simulate the logic from monitor loop when service fails
+                    service_record = ServiceRegistry.get(idx=s_id_1)[0]
+                    check_count = {service_record._id: monitor._max_attempts + 1}  # Exceed max attempts
+                    
+                    # This is the logic from the monitor loop when max attempts are exceeded
+                    if check_count[service_record._id] > monitor._max_attempts:
+                        ServiceRegistry.mark_as_failed(service_record._id)
```
ashish-jabble: I believe we need to implement a patch for the logger as well when the service is marked as failed; otherwise, it may log an entry in the syslogs. To prevent that entry, we need to apply the patch, as it originates from the service registry.

tests/unit/python/fledge/services/core/service_registry/test_monitor.py
```diff
@@ -141,3 +141,135 @@ class TestMonitorException(Exception):
...
```
ashish-jabble: same as above

ashish-jabble (commented):
API tests will also be impacted as we have introduced a new category.

python/fledge/services/core/service_registry/monitor.py
gnandan: Fixed

python/fledge/services/core/service_registry/monitor.py
gnandan: Fixed

python/fledge/services/core/service_registry/monitor.py
gnandan: Fixed

tests/unit/python/fledge/services/core/service_registry/test_monitor.py
gnandan: Fixed

tests/unit/python/fledge/services/core/service_registry/test_monitor.py
gnandan: Fixed

gnandan:
> API tests will also be impacted as we have introduced a new category.

modified API test cases to accommodate new category SUPPORT_BUNDLE

ashish-jabble (dismissed):
Dry review appears satisfactory now. Let us proceed to have it tested by QA.

python/fledge/services/core/service_registry/monitor.py
ashish-jabble:
```
      {
            "key": "/usr/local/foglamp/data/support/support-sin-250812-11-11-16.tar.gz",
            "message": "Support bundle /usr/local/foglamp/data/support/support-sin-250812-11-11-16.tar.gz created for failed service 'sin'",
            "urgency": 3,
            "ts": "2025-08-12 11:11:16.954"
        }
```
Furthermore, the key should include the service name followed by the suffix 'support bundle' - `sin-support-bundle` instead of path. We cannot accept it with just the service name, as upon restarting services, we have already utilized the key in the scenario, and only one alert with a specific key can be active at any given moment.

gnandan: @MarkRiddoch is **servicename-support-bundle** as a key ok ? What convention should be used while naming the key for alerts? 
Also what should be the urgency level for the alert here.

MarkRiddoch: Yes, that key works fine. I would suggest the alert urgency can be High

gnandan: done

ashish-jabble (approved):

PR #1674, opened 2025-07-23
Cursor Rules 

ashish-jabble:
- [x] Only for Python specific and documentation side
- [ ] Other areas can be fit into same way in the existing directory structure I believe

Also, check the README


praveen-garg (commented):
I suggest to create python/tests/ directory and put unit.mdc, api.mdc and system.mdc.

We can use and update the system.mdc from @Mohit04tomar

.cursor/rules/README.md
```diff
@@ -0,0 +1,378 @@
...
+```
+Using @tests/python/unit rules, create unit tests that:
+- Use pytest framework
+- Include proper mocking with pytest-mock
+- Test both success and failure cases
+- Follow the test file naming conventions
+- Include code coverage setup
+```
+
+### System API Testing
```
praveen-garg: API integration testing

Or

REST API Testing


To have cursor avoid confusion with System tests.

praveen-garg (approved):

PR #1675, opened 2025-07-23
Minimum required version of Python has been updated in the setup

ashish-jabble:


praveen-garg (approved):
We can remove this in general, We are not publishing a pip package

MarkRiddoch (approved):

ashish-jabble:
> We can remove this in general, We are not publishing a pip package

Indeed. It is not possible to delete it in a straightforward manner, as the Makefile uses setup.py to build Python packages.

https://github.com/fledge-iot/fledge/blob/21c7c74f04804e96ab5209022a29a9e4d399a1e4/Makefile#L27

PR #1680, opened 2025-07-31
FOGL-9930 Added endpoint for plugin configuration validation (Phase 1 as outlined in the documentation)

ashish-jabble:
Below are the sample of JSON response:

a) Both HostReachable and Listening are good
```
$ curl -H "authorization: $TOKEN" -sX PUT localhost:8081/fledge/plugin/validate -d '{"plugin":{"description":"HTTP Listener South Plugin","type":"string","default":"http_south","readonly":"true"},"host":{"description":"Address to accept data on","type":"string","default":"127.0.0.1","order":"1","displayName":"Host"},"port":{"description":"Port to listen on","type":"integer","default":"6683","order":"2","displayName":"Port"}}' | jq
{
  "HostReachable": {
    "description": "Host Reachable",
    "result": "pass",
    "values": [
      {
        "host": "127.0.0.1"
      }
    ]
  },
  "Listening": {
    "description": "Listening",
    "result": "pass",
    "values": [
      {
        "host": "127.0.0.1",
        "port": "6683"
      }
    ]
  }
}
```

b) HostReachable is good but Listening is Not

```
{
  "HostReachable": {
    "description": "Host Reachable",
    "result": "pass",
    "values": [
      {
        "host": "127.0.0.1"
      }
    ]
  },
  "Listening": {
    "description": "Listening",
    "result": "fail",
    "values": [
      {
        "host": "127.0.0.1",
        "port": "6683"
      }
    ],
    "detail": {
      "reason": "No service is listening on 127.0.0.1:6683"
    }
  }
}

```

### DIFF
--- a/docs/plugin_developers_guide/10_testing.rst
+++ b/docs/plugin_developers_guide/10_testing.rst
@@ -466,3 +466,429 @@ You can also check your default configuration. Although in Python this is usuall
    $ python3 -c 'from fledge.plugins.south.sinusoid.sinusoid import plugin_info; print(plugin_info()["config"])'
    {'plugin': {'description': 'Sinusoid Poll Plugin which implements sine wave with data points', 'type': 'string', 'default': 'sinusoid', 'readonly': 'true'}, 'assetName': {'description': 'Name of Asset', 'type': 'string', 'default': 'sinusoid', 'displayName': 'Asset name', 'mandatory': 'true'}}
 
+Plugin Configuration Validation
+-------------------------------
+
+The plugin configuration validation feature provides automated connectivity testing for plugin configurations. This feature validates that network-related configuration parameters are correct and reachable before deploying or activating a plugin, helping to identify configuration issues early in the plugin testing process.
+
+This is the first iteration of the configuration validation feature, focusing on network connectivity validation for common plugin patterns.
+
+Overview
+~~~~~~~~
+
+The validation system performs two primary types of tests:
+
+**Host Reachability Test**
+   Tests whether the configured host or server is reachable over the network using socket-based connectivity testing.
+
+**Listening Test**
+   Tests whether a service is actively listening on the specified host and port combination.
+
+The validation automatically detects relevant configuration fields and applies appropriate tests based on the field types and naming conventions.
+
+Supported Configuration Keys
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The validation system recognizes the following configuration field patterns (case insensitive):
+
+**Address Fields**
+   - ``address``
+   - ``ip``
+   - ``server``
+   - ``host``
+   - ``hostname``
+
+**Port Fields**
+   - ``port``
+   - ``brokerport``
+
+**URL Fields**
+   - ``url``
+
+**Broker Fields**
+   - ``broker``
+   - ``brokerhost``
+
+Configuration Value Priority
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When both ``default`` and ``value`` keys are present in a configuration field, the validation system uses the following priority:
+
+1. **``value``** - Takes precedence when present
+2. **``default``** - Used when ``value`` is not present
+3. **Error** - Raised if neither ``value`` nor ``default`` is present
+
+Common Configuration Patterns
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Host and Port Configuration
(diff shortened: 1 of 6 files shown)


docs/plugin_developers_guide/10_testing.rst
```diff
@@ -466,3 +466,429 @@ You can also check your default configuration. Although in Python this is usuall
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The validation API returns a JSON response with test results:
+
+**Success Response (HTTP 200):**
+
+.. code-block:: json
+
+   {
+       "HostReachable": {
```
praveen-garg, line 823: Please avoid mixing cases. What is the convention we are following throughout - camelCase, snake_case, what? 

We are doing a poor job in writing semantically incorrect endpoints URLs and request / response structure with giving proper thoughts. And this is happeing every now and then for many services.

python/fledge/services/core/api/plugins/config_validator.py
```diff
@@ -0,0 +1,726 @@
...
+                                        
+                                        await asyncio.get_event_loop().run_in_executor(
+                                            None, sock.connect, (ip_addr, alt_port)
+                                        )
+                                        sock.close()
+                                        _logger.debug(f"Successfully connected to {ip_addr}:{alt_port}")
+                                        return True, f"Host '{hostname}' is reachable"
+                                        
+                                    except (socket.timeout, OSError):
+                                        sock.close()
```
praveen-garg: cannot we  use with context to avoid any issues?

python/fledge/services/core/api/plugins/config_validator.py
```diff
@@ -0,0 +1,726 @@
...
+    """
+    
+    # Configuration item names to check (case insensitive)
+    ADDRESS_FIELDS = ['address', 'ip', 'server', 'host', 'hostname']
+    URL_FIELDS = ['url']
+    BROKER_FIELDS = ['broker', 'brokerhost']
+    PORT_FIELDS = ['port', 'brokerport']
+    
+    # Standard protocol ports
+    STANDARD_PORTS = {
```
praveen-garg, line 44: Should we make them configurable? or atleast allow user to set in a file (could be json, yml or better ini file) using Env variable to override anything without a need to new release or inline edit?!

praveen-garg (commented):
We bettter return per-port/host result details; You can imagine fallback or secondary URLs.

python/fledge/services/core/api/plugins/config_validator.py
ashish-jabble: This is only for the first iteration. If no port is specified, it will examine the protocol and utilize the standard port associated with the indicated protocol. Default ports are outlined in `/etc/services`, along with the corresponding protocols they for. 
The second iteration can provide enhancements, as the first iteration is more closely aligned with the specific configurations of the plugin.

docs/plugin_developers_guide/10_testing.rst
ashish-jabble: We need to establish a standard approach to prevent mixing these conventions across.

Also, For configuration items in plugins, we mostly use snake_case in Python plugins, whereas in C, we often use camelCase or a combination of both.

python/fledge/services/core/api/plugins/config_validator.py
ashish-jabble: using context managers is indeed the more elegant and reliable approach for resource management compared to explicit `.close()` calls. Therefore code has been updated.

docs/plugin_developers_guide/10_testing.rst
MarkRiddoch: These names date from when we adopted no description in the spec and can be camelCase now that we don;t want to use them in messages.

python/fledge/services/core/api/plugins/config_validator.py
MarkRiddoch: This is fine like this. It is certainly not something users will want to, or should be configuring. If anybody does it will be the plugin writers.

MarkRiddoch (approved):

python/fledge/services/core/api/plugins/config_validator.py
praveen-garg: Then we must remove plugin specific items from here, to avoid any confusion. 

And we should test plugins having primary and secondary url/hosts

PR #1687, opened 2025-08-20
FOGL-10025 : Added support for handling dynamic configuration changes

gnandan:
Improvement for raising alert while creating automated support bundle is also done. Alert will be generated only if not exists already


python/fledge/services/core/service_registry/monitor.py
```diff
@@ -179,19 +226,111 @@ async def raise_support_bundle_alert(self, service_name, bundle_name):
         """Raise alert for automated support bundle creation"""
         if not self._alert_manager:
             self._alert_manager = AlertManager(connect.get_storage_async())
+        # Don't create alert if already exists
```
MarkRiddoch, line 229: This seems unrelated to the Jira. Is it intentionally included?

gnandan: Yes, It's intentional, I realized that this improvement is necessary it was left in last JIRA. I mentioned it in PR  description and created a separate JIRA  [FOGL-10208](https://scaledb.atlassian.net/browse/FOGL-10208) for this as well. Since it is a small change that's why I clubbed it here.

MarkRiddoch (approved):

firozalikhan (approved):

PR #1688, opened 2025-08-20
FOGL-10187 GET endpoint for retrieving service information

ashish-jabble:
1. Get all services info
```
$ curl -sX GET localhost:8081/fledge/service/info | jq
{
  "services": [
    {
      "name": "storage",
      "description": "The storage service buffers data within a single Fledge instance",
      "type": "storage",
      "process": "storage",
      "process_script": "storage"
    },
    {
      "name": "Notification Service",
      "description": "Fledge Notification Service to send notifications",
      "package": "fledge-service-notification",
      "type": "notification",
      "process_name": "notification_c",
      "process_script": "services/notification_c"
    },
    {
      "name": "dispatcher",
      "description": "",
      "type": "",
      "process": "",
      "process_script": ""
    },
    {
      "name": "south",
      "description": "Service used to interact with device, API and generic sources of data",
      "type": "south",
      "process": "south",
      "process_script": "south_c"
    },
    {
      "name": "north",
      "description": "Service used to interact with device, API and generic sources of data",
      "type": "north",
      "process": "north",
      "process_script": "north_C"
    }
  ]
}
```

**Note**: 
a) notification service with `--info parameter` to fetch service information.
b) the dispatcher service is installed and record contains only empty values, as there is currently no `--info` entry point available for the service. Therefore, this demonstration is to show empty values for a service which have info is missing.
c) The services for south, north, and storage are prebuilt, thus providing a hardwired API response. It is unclear whether this information is also derived from the --info command, which posed a similar query in the documentation itself.
d) process_script is also unclear whether it with process name or with path like
`["services/south_c"], ["services/notification_c"], ["services/north_C"]`

2. Get service info by name
```
$ curl -sX GET localhost:8081/fledge/service/info/notification | jq
{
  "name": "Notification Service",
  "description": "Fledge Notification Service to send notifications",
  "package": "fledge-service-notification",
  "type": "notification",
  "process_name": "notification_c",
  "process_script": "services/notification_c"
}


```

### DIFF
--- a/python/fledge/services/core/api/service.py
+++ b/python/fledge/services/core/api/service.py
@@ -11,6 +11,7 @@ import datetime
 import uuid
 import json
 import multiprocessing
+import subprocess
 from aiohttp import web
 
 from typing import Dict, List
@@ -34,7 +35,7 @@ from fledge.common.audit_logger import AuditLogger
 from fledge.common.web.middleware import has_permission
 
 
-__author__ = "Mark Riddoch, Ashwin Gopalakrishnan, Amarendra K Sinha"
+__author__ = "Mark Riddoch, Ashwin Gopalakrishnan, Amarendra K Sinha, Ashish Jabble"
 __copyright__ = "Copyright (c) 2018 OSIsoft, LLC"
 __license__ = "Apache 2.0"
 __version__ = "${VERSION}"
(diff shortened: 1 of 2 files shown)


MarkRiddoch:
> 
> **Note**: a) notification service with `--info parameter` to fetch service information. 
> b) the dispatcher service is installed and record contains only empty values, as there is currently no `--info` entry point available for the service. Therefore, this demonstration is to show empty values for a service which have info is missing.

This all be resolved by FOGL-10186

> c) The services for south, north, and storage are prebuilt, thus providing a hardwired API response. It is unclear whether this information is also derived from the --info command, which posed a similar query in the documentation itself. 

It's probably best to implement the --info for each of these and call them for consistency.

> d) process_script is also unclear whether it with process name or with path like `["services/south_c"], ["services/notification_c"], ["services/north_C"]`

I think this should be without the path component since they should always be in scripts/services

MarkRiddoch (commented):
Looks good so far, we need to add the cache to stop it running the commands too frequently

gnandan:
> > **Note**: a) notification service with `--info parameter` to fetch service information.
> > b) the dispatcher service is installed and record contains only empty values, as there is currently no `--info` entry point available for the service. Therefore, this demonstration is to show empty values for a service which have info is missing.
> 
> This all be resolved by FOGL-10186
> 
> > c) The services for south, north, and storage are prebuilt, thus providing a hardwired API response. It is unclear whether this information is also derived from the --info command, which posed a similar query in the documentation itself.
> 
> It's probably best to implement the --info for each of these and call them for consistency.
> 
> > d) process_script is also unclear whether it with process name or with path like `["services/south_c"], ["services/notification_c"], ["services/north_C"]`
> 
> I think this should be without the path component since they should always be in scripts/services



> > d) process_script is also unclear whether it with process name or with path like `["services/south_c"], ["services/notification_c"], ["services/north_C"]`
> 
> I think this should be without the path component since they should always be in scripts/services

@MarkRiddoch process_name is used as ```notification_c``` . if process_script is used as ```notification_c``` , then what should be the process_name?  

It would be great if you can provide actual values of each key for below requests, so that I will implement accordingly for other services.

```
$ curl -sX GET localhost:8081/fledge/service/info/notification | jq
{
  "name": "Notification Service",
  "description": "Fledge Notification Service to send notifications",
  "package": "fledge-service-notification",
  "type": "notification",
  "process_name": "notification_c",
  "process_script": "services/notification_c"
}
```

MarkRiddoch (approved):

PR #1691, opened 2025-08-26
FOGL-10108 Service login API core management route

ashish-jabble:
Changes Made:
- [x] Removed the existing GET endpoint for authtoken, which previously used hardcoded admin credentials.
- [x] Introduced a new service-based login API route accessible via the core management port, with added flexibility to pass the username in the request payload.

🔍 Notes:

- Currently, only the Management Service consumes the removed GET endpoint. Therefore, its removal is safe.
- Updating the Management Service to use the new login API will be handled separately under JIRA-10237.


praveen-garg (commented):
Looks fine to me.
Cannot we reuse the user log in code partially or completely.

We always have admin user in the system so what is the guarantee that without the password or certificate it cannot be mis-used.

MarkRiddoch (approved):

PR #1695, opened 2025-09-02
FOGL-10166 : Used integer value within 32-bit range

gnandan:


ashish-jabble:
@Mohit04tomar Please execute the custom runs on Rpi-based environments, as we currently do not have the on-demand environment available for this purpose. Therefore, to prevent any disruption to the existing nightly tests, please conduct them accordingly.

Mohit04tomar (approved):

MarkRiddoch (approved):

PR #1698, opened 2025-09-12
FOGL-10226 - service info updates

ashish-jabble:


MarkRiddoch (commented):
Looks like the right way yo go to me. Removing the hardwired knowledge from the Fledge repository, especially for FogLAMP specific services, is a good step to making it more extensible. In the future we want to be able to add new services without having to touch the core code.

PR #1703, opened 2025-09-23
FOGL-10313: check correctness for default values of list and kvlist

gnandan:

PR #1704, opened 2025-09-23
FOGL-10313 based unit tests for default for list/kvlist  

praveen-garg:


ashish-jabble:
Out of the 6 new tests, 4 have failed. The report indicates that the issue is related specifically to the list type only, and not to the kvlist. However, the error message we're seeing is: `"For test category, KV pair invalid in default value for item name test_item_name"`. This suggests that the validation is correctly identifying non-dictionary items being passed as defaults.

praveen-garg:
Fixed by #1703

ashish-jabble:
> Fixed by https://github.com/fledge-iot/fledge/pull/1703

A. Single Parsing (Parses input only once)

B. Accurate Exception Handling Raises TypeError for type mismatches — semantically correct

C. Early Validation - Performs immediate checks and halts execution on invalid types.

D. Clearer Error Messages

Please check if that makes sense!

praveen-garg:
The good part in #1703 :
1) A clear separation of tests per type and good / bad
2) [API test fixes  ](https://github.com/fledge-iot/fledge/pull/1703/files#diff-c2fb5374273395db9de3cb665cfb4e0f1ce07d99816b4a47a8e28dd7168b0296)

Ultimately we should target refactoring as cursor [review (comment)](https://github.com/fledge-iot/fledge/pull/1703#issuecomment-3327820315) suggests.

tests/unit/python/fledge/common/test_configuration_manager.py
```diff
@@ -793,10 +816,10 @@ async def test__validate_category_val_bucket_type_bad(self, config, exc_name, re
                       "listSize": "blah"}}, ValueError, "For {} category, listSize value must be an integer value for "
                                                         "item name {}".format(CAT_NAME, ITEM_NAME)),
         ({ITEM_NAME: {"description": "expression", "type": "kvlist", "default": "[\"1\"]", "items": "float",
-                      "listSize": "1"}}, TypeError, "For {} category, KV pair invalid in default value for item name "
+                      "listSize": "1"}}, TypeError, "For {} category, default value should be a list of KV pairs in string format for item name "
```
praveen-garg: for end user we better use term key-value instead of KV

python/fledge/common/configuration_manager.py
```diff
@@ -487,6 +487,15 @@ def get_entry_val(k):
                                                  'for item name {}'.format(category_name, item_name))
                             list_size = int(item_val['listSize'])
                         msg = "array" if item_val['type'] == 'list' else "KV pair"
+                        expected_type = list if item_val['type'] == 'list' else dict
+                        expected_format = "a list of values" if expected_type is list else "a list of KV pairs"
```
praveen-garg: else "an object with key-value pair"

praveen-garg: a list of items OR  a list of values for items type

praveen-garg: We should not confuse **type** _list_ of **items** _object_ with **type** _kvlist_

tests/unit/python/fledge/common/test_configuration_manager.py
praveen-garg: and not a list

PR #1705, opened 2025-10-03
FOGL-10281 Don't allow an empty pipeline to be debugged

MarkRiddoch:


pintomax (dismissed):

firozalikhan:
@MarkRiddoch The response is not informative when user attaches a service with no filter, it does not tell why-
`status: "Failed to attach the debugger to the pipeline"`

firozalikhan:
The new message correctly informs the user that a service with no filters cannot be attached. However, even when the service is not attached, running the suspendIngest command still stops ingestion — which seems inconsistent.
Also, what should be the expected behavior when a service has filters that are disabled? Currently, the debugger attaches successfully in that case as well.

pintomax (approved):

PR #1706, opened 2025-10-07
FOGL-10278: Added check for allowed characters

gnandan:
Added two function for checking valid identifiers. 

for C++ **isValidIdentifier** method is implemented in fledge/C/common/include/utils.h
for python **is_valid_identifier** method is implemented in fledge/python/fledge/common/utils.py

### DIFF
--- a/C/common/include/management_client.h
+++ b/C/common/include/management_client.h
@@ -146,9 +146,10 @@ class ManagementClient {
 		template<class T> bool	addCategory(const T& t, bool keepOriginalItems = false)
 		{
 			try {
-				if (!isValidIdentifier(t.getName()))
+				std::string blockedCharacter = {};
+				if (!isValidIdentifier(t.getName(), blockedCharacter))
 				{
-					m_logger->error("The category name '%s' contains invalid characters.", t.getName().c_str());
+					m_logger->error("The category name %s contains %s invalid character(s).", blockedCharacter.c_str(), t.getName().c_str());
 					return false;
 				}
 				std::string url = "/fledge/service/category";
--- a/C/common/include/management_client.h
+++ b/C/common/include/management_client.h
@@ -24,6 +24,7 @@
 #include <thread>
 #include <bearer_token.h>
 #include <acl.h>
+#include "utils.h"
 
 using HttpClient = SimpleWeb::Client<SimpleWeb::HTTP>;
 using HttpServer = SimpleWeb::Server<SimpleWeb::HTTP>;
@@ -145,6 +146,11 @@ class ManagementClient {
 		template<class T> bool	addCategory(const T& t, bool keepOriginalItems = false)
 		{
 			try {
+				if (!isValidIdentifier(t.getName()))
+				{
+					m_logger->error("The category name '%s' contains invalid characters.", t.getName().c_str());
+					return false;
+				}
 				std::string url = "/fledge/service/category";
 
                                 // Build the JSON payload
--- a/python/fledge/common/configuration_manager.py
+++ b/python/fledge/common/configuration_manager.py
@@ -1447,8 +1447,9 @@ class ConfigurationManager(ConfigurationManagerSingleton):
         category_val_prepared = ''
         try:
             # Don't allow invalid character in category name
-            if not common_utils.is_valid_identifier(category_name):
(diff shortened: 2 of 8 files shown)


python/fledge/common/utils.py
```diff
@@ -40,6 +40,22 @@ def check_reserved(string):
...
 
+def is_valid_identifier(string):
+    """
+    Check if the given string is a valid identifier which doesn't have any disallowed characters.
+   
+    :param string: The string to check.
+    :return: True if the string is a valid identifier, False otherwise.
+    """
+    disallowed_characters = ["\\"]
+    if string is None or not isinstance(string, str) or string == "":
```
ashwini-k-pandey, line 51: what is str ..?

praveen-garg: str is type for string datatype. We must name the arg better say simply "s" instead of "string" to avoid confusion.

gnandan: @ashwini-k-pandey @praveen-garg I used variable name string to make the changes look consistent with other functions in the utils.py.

C/common/include/management_client.h
```diff
@@ -145,6 +146,11 @@ class ManagementClient {
 		template<class T> bool	addCategory(const T& t, bool keepOriginalItems = false)
 		{
 			try {
+				if (!isValidIdentifier(t.getName()))
+				{
+					m_logger->error("The category name '%s' contains invalid characters.", t.getName().c_str());
```
MarkRiddoch: It would be good to report which characters are invalid - it would certainly help the user to fix things. Consider adding an optional string argument that is used to get a error report.

python/fledge/common/configuration_manager.py
```diff
@@ -1446,6 +1446,10 @@ async def create_category(self, category_name, category_value, category_descript
 
         category_val_prepared = ''
         try:
+            # Don't allow invalid character in category name
+            if not common_utils.is_valid_identifier(category_name):
+                raise ValueError("Invalid character found in category name {}".format(category_name))
```
MarkRiddoch: As above, a more helpful error report would make this easier for a user to resolve.

python/fledge/services/core/api/service.py
```diff
@@ -847,6 +853,11 @@ async def add_service(request):
             _logger.exception("Failed to create schedule. %s", ex.error)
             raise web.HTTPInternalServerError(reason='Failed to create service.')
         except Exception as ex:
+            if "Invalid character found" in ex.args[0]:
+                    msg = "Failed to create service. {}".format(ex.args[0])
```
MarkRiddoch: Again, make tis more informative please

python/fledge/services/core/scheduler/scheduler.py
```diff
@@ -1068,6 +1068,9 @@ async def save_schedule(self, schedule: Schedule, is_enabled_modified=None, dryr
         # TODO should these checks be moved to the storage layer?
         if schedule.name is None or len(schedule.name) == 0:
             raise ValueError("name can not be empty")
+        # Don't allow invalid character in the schedule name
+        if not common_utils.is_valid_identifier(schedule.name):
+            raise ValueError("Invalid character found in schedule name {}".format(schedule.name))
```
MarkRiddoch: Can we also make this more informative?

MarkRiddoch (commented):
I'd like to see the error messages be more helpful to the end user, but otherwise this is ok.

C/common/include/management_client.h
gnandan: Improved error message

python/fledge/common/configuration_manager.py
gnandan: Improved error message

python/fledge/services/core/api/service.py
gnandan: ex.args[0] contains details why it failed.

python/fledge/services/core/scheduler/scheduler.py
gnandan: Improved error message

MarkRiddoch (approved):

firozalikhan (approved):

PR #1709, opened 2025-10-30
FOGL-10204 : Increased wait time for co-routine by 1 second to finish

gnandan:
Increased wait time for co-routine so that disable schedule can finish before deleting category

### DIFF
--- a/python/fledge/services/core/api/service.py
+++ b/python/fledge/services/core/api/service.py
@@ -476,7 +476,7 @@ async def delete_service(request):
         if svc_schedule['enabled'].lower() == 't':
             await server.Server.scheduler.disable_schedule(sch_id)
             # return control to event loop
-            await asyncio.sleep(1)
+            await asyncio.sleep(2)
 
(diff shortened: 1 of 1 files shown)


python/fledge/services/core/api/service.py
```diff
@@ -476,7 +476,7 @@ async def delete_service(request):
         if svc_schedule['enabled'].lower() == 't':
             await server.Server.scheduler.disable_schedule(sch_id)
             # return control to event loop
-            await asyncio.sleep(1)
+            await asyncio.sleep(2)
```
praveen-garg, line 479: How does this gurantee? and why does it works for all other services?

praveen-garg (approved):

python/fledge/services/core/api/service.py
gnandan: In case of management service two schedules are created one for Management service and one for FogMan poll Agent.  However, for other services only one schedule is there. That's why cleaning up might take some extra time.

